PR #3, opened 2021-06-17
Added peer.exchanging module for processing new `exn` messages for peer to peer message exchange.

pfeairheller:
This PR includes:

* Added peer.exchanging module for processing new `exn` messages for peer to peer message exchange.

* Added Signer Seal (-S) a couple of (pre+ControllerIndexSigs) for verification of `req` and `exn` messages against the existing signing keys of a controller.

### DIFF
--- a/docs/Peer2PeerCredentials.md
+++ b/docs/Peer2PeerCredentials.md
@@ -1,0 +1,434 @@
+# KERI Peer to Peer Communication / Credential Exchange Protocol
+
+## Introduction
+
+This document proposes additions to the KERI protocol to facilitate peer to peer secure communications. It also
+introduces two enveloped protocols embedded in these new KERI messages for Credential Issuance and Presentation Exchange
+leveraging the [https://identity.foundation/credential-manifest/](Credential Manifest)
+and [https://identity. foundation/presentation-exchange/](Presentation Exchange) data formats.  Finally, CESR Proof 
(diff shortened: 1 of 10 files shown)


m00sey:
A test comment

SmithSamuelM:
Couple of things. In the counter code table we have been allocating codes in order not using mnemonics So the -S should be instead the next one in the list. I worked down from -A for attachments that are primitive groups and up from -Z for pipeline counters.  

Also saw that you added pygtrie as a dependency for route strings but no sure why a dict is not sufficient. Unless you need lexicographic ordering a dict is on average performance faster than  a trie. I don't believe that you need lexicographic ordering? But maybe I misunderstand.

pfeairheller:
I will change to -H.

The trie allows some operations on routes by prefix which I thought might come in handy later on. But certainly not needed.  I change to a dict and remove the dependency.

SmithSamuelM:
What happens when signature for req, or exn is from non-trans derivation code?   In this case there is no requirement that there be a KEL for that identifier as the public key is in the identifier. So a receipt couple is sufficient. So maybe should support cigar as another argument to function call.  NonTransReceiptCouples   It seems like the Sig Source couple does not provide enough information. Don't you really need a TransIndexedSigGroups?  do you want to jump on a zoom call to discuss?

PR #6, opened 2021-06-24
Schemer and Saider for working with self-addressing schema.

pfeairheller:
This PR contains:

* Schemar class for serializing and verifying schema of type JSON Schema (for now) with self-addressing identifiers.

* Generic Sadder class for verifying self-addressing identifiers of JSON documents.

### DIFF
--- a/docs/Peer2PeerCredentials.md
+++ b/docs/Peer2PeerCredentials.md
@@ -52,7 +52,6 @@ KERI Message: `exn` for peer to peer exchanges.
    "dt": "2020-08-22T17:50:12.988921+00:00"
    "r": "/credential/offer"                                // resource with nested namespace
    "q": {
-      "x": "EMbzB4tnrjBUdhcMadmXq2F2adYdW42iw2NPJz8kUo_c",
       "issuer": "did:keri:EEBp64Aw2rsjdJpAR0e2qCq3jX7q7gLld3LjAwZgaLXU"
       "output_descriptors: [
          "EckOnHB11J4H9q16I3tN8DdpNXnCiP5QJQ7yvkWqTDdA"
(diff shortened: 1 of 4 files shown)


SmithSamuelM:
The Saider class looks good.  The Schemer class seems OK although there is only one serialization type currently supported but I suppose more might be in the offing and using the same protocol as Serder makes it easier on the message stream parser.

SmithSamuelM:
The Schemage named tuple seemed odd. It looks likes the values are function references. Usually this would be a class with methods. Its not very pythonic to use a named tuple that way.  Might want to revisit Schemage.  The idea of a named tuple is that is multi-valued data i.e. a data structure and the named tuple allows one to access the data fields by name instead of offset.  As soon as you start using functions as values though it crosses the line into just use a class with methods and attributes.  Likewise a dataclass is just a more powerful named tuple, its a way to have data structures in python that are simpler than a class to define.

pfeairheller:
> The Schemage named tuple seemed odd. It looks likes the values are function references. Usually this would be a class with methods. Its not very pythonic to use a named tuple that way. Might want to revisit Schemage. The idea of a named tuple is that is multi-valued data i.e. a data structure and the named tuple allows one to access the data fields by name instead of offset. As soon as you start using functions as values though it crosses the line into just use a class with methods and attributes. Likewise a dataclass is just a more powerful named tuple, its a way to have data structures in python that are simpler than a class to define.

I was most concerned with the Schemage namedtuple.  I will revisit.

SmithSamuelM:
Under the hood everything in python is an object (i.e. has a class and is an instance of a class)  Special classes have syntactic sugarr to make them easier to use in the special cases they are meant for. So it can be tempting to use them in cases that they are not meant for.  Tuples are immutable lists so they are lighter weight than a list and good for little snippets of multivalued data that can be tuple packed and upacked (syntactic suger).  A named tuple is just more syntactic sugarr to allow named reference when the tuple is not unpacked.  But as soon as you are using the values of a tuple to call functions and not for packing unpacking then its looks abusive.  Under the hood a named tuple is just an immutable instance of a class with named attributes.  In many cases a dataclass would  be a better fit than a named tuple.  I used both because I am more familiar with named tuples and dataclasses are relatively new to python. Both work well as enums  in fact for non numeric values better than python's enum class. What I ended up deciding is that for simple data that was very homogeneous that a namedtuple is easier to code up vs a dataclass   but for bigger tables that the dataclass definition allows for  more structure with comments etc that make is more self-documenting in spite of being harder to code.  So Tierage is a namedtuple and MatterCodex is a frozen dataclass that iterates as a tuple. The latter is because the of complexity of the codes of different lengths etc and its just reads better to see them structured that way.  But in neither case are there methods of functionality that are not native to dataclasses and named tuples. At that point I would always create a class.  I hope that helps.  Dataclasses can do much more than named tuples so as in the Komer we use them not as an alternative to a nemedtuple but as a schema for a database structure.  Python is getting more complex which can be a bad thing because they are now multiple ways to do the same thing which is unpythonic.

pfeairheller:
Schemage has been rewritten as a class.  I'm going to merge this PR now and add Serder like protocol support to Schemer in my next PR since I want to do the exact same thing to a new Credentialer class for the CESR Proof Format.

PR #45, opened 2021-08-03
Command line command for incepting a multi-sig identifier

pfeairheller:
This PR includes:

* Command line command for incepting a multi-sig identifier with keys that originate from distributed key stores.
* Command line for key rotation.  Supports new key and witness configurations.
* HTTP Mailbox for peer to peer communication with a Respondant that routes replies to either local mailbox or witness mailbox depending on the recipient (witnesses or not).

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -4,9 +4,12 @@ KERI
 keri.app.agenting module
 
 """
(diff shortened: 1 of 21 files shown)


m00sey (commented):
lgtm

PR #47, opened 2021-08-04
Replace @doize decorator on methods with doify() wrapper.  Update hio dependency. Update version keripy.

SmithSamuelM:


SmithSamuelM:
didn't merge yet. In case it messes with the demo.

PR #188, opened 2022-05-18
Make witnesses reentrant and identifiers use randy algo when no salt is provided.

pfeairheller:
…


m00sey:
lgtm

PR #193, opened 2022-05-22
adds /boot/{alias} endpoint for determining if a keystone has been created.

m00sey:
Supports Keep initialization.


m00sey:
will squash after review

PR #230, opened 2022-07-13
Added multisig-holder-agent.sh in scripts/demo/credentials

rc-provenant:
Created new script multisig-holder-agent.sh
The script includes the schema OOBis


pfeairheller (approved):
Looks good to me

PR #292, opened 2022-09-03
Minor updates for multisig create (metadata fix and supporting establishment only)

pfeairheller:


m00sey (approved):
Lgtm

PR #316, opened 2022-09-26
Change JavaScript MIME type

euj1n0ng:
A minor fix for successful testing.

`text/javascript` is a recommended standard (both by IETF and by MDN). `application/javascript` is obsolete.

### DIFF
--- a/tests/end/test_ending.py
+++ b/tests/end/test_ending.py
@@ -285,7 +285,7 @@ def test_get_static_sink():
     # get trial.js
     rep = client.simulate_get('/static/index.js')
     assert rep.status == falcon.HTTP_OK
-    assert rep.headers['content-type'] == 'application/javascript; charset=UTF-8'
+    assert rep.headers['content-type'] == 'text/javascript; charset=UTF-8'
     assert len(rep.text) > 0
     assert rep.text == '// vanilla index.js\n\nm.render(document.body, "Hello world")\n'
 


pfeairheller (changes requested):
This change will break this test as it's testing the return from the HTTP request.  

If you want to change the MIME type of the actual request you'll have to update Hio, I believe

euj1n0ng:
> This change will break this test as it's testing the return from the HTTP request.
> 
> If you want to change the MIME type of the actual request you'll have to update Hio, I believe


Yeah, you are right. At a glance, `falcon` still uses `application/javascript` for the JS MIME type in the simulation of HTTP request. However, surprisingly, on Ubuntu machine the test was failed on the stale MIME type while there was no problem on a Mac machine.
image

If you want, I'll close this PR. Of course, you can directly close it.

euj1n0ng:
https://github.com/falconry/falcon/issues/2110

pfeairheller:
So we can close this issue here I assume?

euj1n0ng:
Yes.

vytas7:
Hello @euj1n0ng & @pfeairheller!
While we are looking to enforce Falcon's standard media types for some common file types, you can absolutely override this in your application if you wish.

Just add the desired types to your application's [`app.resp_options`](https://falcon.readthedocs.io/en/stable/api/app.html#falcon.App.resp_options), for instance,
```python
app.resp_options.static_media_types['.js'] = 'text/javascript'
```

vytas7:
**Update**: you should now get @euj1n0ng's improvement automatically via Falcon 4.0.0+.

Issue #313, opened 2022-09-23
Keri agent crashes when one or more witness receipts are missing


psteniusubi:
Steps to reproduce

1. Reset
2. Do not start witness
3. Start Keri agent
4. Run boot, unlock, incept 
5. Running GET /oobi/alias causes Keri agent to crash

This also happens if I start only two witness nodes (eg. only was and wel witnesses). Or if witness nodes are "slow" and GET /oobi/alias happens before receipts are received.

keripy is at revision ed016a06c56a1faaff26f2cca24b8b9e385ffc54 

```
$ rm -rf ~/.keri

$ kli agent vlei & 

$ curl -s -X POST "http://localhost:5620/boot" -H "accept: */*" -H "Content-Type: application/json" -d "{\"name\":\"alias1\",\"passcode\":\"DoB2-6Fj4x-9Lbo-AFWJr-a17O\",\"salt\":\"0ACDEyMzQ1Njc4OWxtbm9aBc\"}" | jq

$ curl -s -X PUT "http://localhost:5620/boot" -H "accept: */*" -H "Content-Type: application/json" -d "{\"name\":\"alias1\",\"passcode\":\"DoB2-6Fj4x-9Lbo-AFWJr-a17O\"}" | jq

$ sleep 3

$ curl -s -X POST "http://localhost:5620/ids/alias1" -H "accept: */*" -H "Content-Type: application/json" -d "{\"transferable\":true,\"wits\":[\"BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha\",\"BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM\",\"BIKKuvBwpmDVA4Ds-EpL5bt9OqPzWPja2LigFYZN2YfX\"],\"toad\":2,\"icount\":1,\"ncount\":1,\"isith\":1,\"nsith\":1}" | jq

$ curl -i -s "http://localhost:5620/oobi/alias1?role=witness" -H "accept: */*"
```

Keri agent quits, output and stack trace is

```
$ kli agent vlei

******* Starting Agents for vLEI scenairo testing on ports:

    RootGARs:  5620, 5621
    ExtGARs:   5622, 5623
    IntGARs:   5624, 5625
    QARs:      5626, 5627
    LARs:      5628, 5629
    Person:    5630

*******

ERR: 'http'
Traceback (most recent call last):
[...]

PR #325, opened 2022-10-05
Issue 313 crash when witness missing

psteniusubi:
Fix for issue #313 

The code was not handling "not urls" case which caused KeyError from following statement. Added raise MissingEntryError. Handling this error in indirecting.py by letting the while loop retry the operation.

agenting.httpClient is invoked from some other places in code. I have not improved error handling in those places.


pfeairheller (approved):
Looks good to me

Issue #314, opened 2022-09-23
Invalid content-type in Keri agent http error messages


psteniusubi:
Steps to reproduce

1. Reset
2. Start Keri agent
3. Run GET /ids/notfound or GET /contacts/unknown

keripy is at revision ed016a06c56a1faaff26f2cca24b8b9e385ffc54

```
$ kli agent vlei

$ curl -s -X POST "http://localhost:5620/boot" -H "accept: */*" -H "Content-Type: application/json" -d "{\"name\":\"alias1\",\"passcode\":\"DoB2-6Fj4x-9Lbo-AFWJr-a17O\",\"salt\":\"0ACDEyMzQ1Njc4OWxtbm9aBc\"}" | jq

$ curl -s -X PUT "http://localhost:5620/boot" -H "accept: */*" -H "Content-Type: application/json" -d "{\"name\":\"alias1\",\"passcode\":\"DoB2-6Fj4x-9Lbo-AFWJr-a17O\"}" | jq

$ sleep 3
```

GET /ids/notfound

Error message is misplaced into content-type header value

```
$ curl -i -s http://localhost:5620/ids/notfound
HTTP/1.1 404 Not Found
Content-Type: no identifier for alias notfound
Content-Length: 0
Server: Ioflo WSGI Server
Date: Fri, 23 Sep 2022 06:57:51 GMT

```

GET /contacts/unknown

Error message content type is json but body is plain text

```
$ curl -i -s http://localhost:5620/contacts/BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha
HTTP/1.1 404 Not Found
Content-Length: 71
Content-Type: application/json
Server: Ioflo WSGI Server
Date: Fri, 23 Sep 2022 06:58:15 GMT

BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha is not a known identifier.
```

pfeairheller:
Agent removed from KERIpy for WebOfTrust/KERIA

PR #327, opened 2022-10-05
Issue 314 invalid content type

psteniusubi:
Fix for issue #314

I'd like to suggest fixing this by raising HTTPNotFound exception

With this change the error looks like this

```
$  curl -i -s http://localhost:5620/ids/notfound
HTTP/1.1 404 Not Found
Content-Type: application/json
Vary: Accept
Content-Length: 77
Server: Ioflo WSGI Server
Date: Wed, 05 Oct 2022 15:49:19 GMT

{"title": "404 Not Found", "description": "no identifier for alias notfound"}
```

I can also apply this style of fix to other places where http errors are generated.

### DIFF
--- a/src/keri/app/kiwiing.py
+++ b/src/keri/app/kiwiing.py
@@ -220,9 +220,7 @@ class IdentifierEnd(doing.DoDoer):
         """
         hab = self.hby.habByName(alias)
         if hab is None:
-            rep.status = falcon.HTTP_404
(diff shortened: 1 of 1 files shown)


pfeairheller (approved):
Good catch, that "falcon.content_type" was supposed to be "falcon.text".

Overall we want to add structure to all error return types from the Agent APIs but haven't gotten there yet.

PR #334, opened 2022-10-17
Fix docker build and add installation docs

kentbull:
The docker build was failing on the `blake3` library for me due to not seeing the Cargo environment in the path. I also had issues downloading the `ordered-set` dependency since it requires Python higher than 3.9.7 so I upgraded the Docker image to 3.10.8.

Prefixing the `pip install` with the `source $HOME/.cargo/env` ensures that the `pip install` command sees the Rust environment that was installed in an earlier new step.

I collected the apt-get instructions into one `RUN` instruction for less layers in the final Docker image.


Dockerfile
```diff
@@ -1,11 +1,17 @@
-FROM python:3.9.7-buster
+FROM python:3.10.8-buster
```
pfeairheller, line 1: Can this be changed to 3.10.4 that matches setup.py?

pfeairheller (changes requested):
Please change the python docker image reference to 3.10.4 and we can merge this in.

Dockerfile
kentbull: done

kentbull:
I added one extra commit to order the newest changes last in the Dockerfile to take maximum advantage of Docker layer caching. This allows you to only have to download and build the Rust binaries once and then on each incremental Docker image build only the new Python source has to be mounted.

SmithSamuelM:
So should we update keripy pip install to be python 3.10.8?

pfeairheller:
No, the opposite.  I made Kent change the Dockerfile to 3.10.4

SmithSamuelM:
Since we are releasing a new version of KERIpy on PyPI wouldn’t that be a good time to update the version. In general.

pfeairheller:
Yes, that's a good point.  I will give it a try after my current meeting.

PR #349, opened 2022-10-26
Remove old use of Respondant from REST endpoint for signing and sending a challenge response.

pfeairheller:


m00sey (approved):
lgtm

PR #400, opened 2022-12-07
Fix witnessing election logic of Counselor

pfeairheller:
Fix witnessing election logic of Counselor to correctly select the participant (signer) with the lowest index.


m00sey (approved):
lgtm

dhh1128 (commented):
LGTM

PR #410, opened 2022-12-23
Fix error: COPY failed: forbidden path outside the build context docker compose

s-a-tanjim:


.github/workflows/python-app-ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         os: [ macos-latest, ubuntu-latest ]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
```
pfeairheller, line 23: Why are we changing this version?

images/keripy.dockerfile
```diff
@@ -1,15 +1,19 @@
...
+RUN apk add alpine-sdk
+RUN apk add libffi-dev
+RUN apk add libsodium
+RUN apk add libsodium-dev
 
 # Setup Rust for blake3 dependency build
 RUN curl https://sh.rustup.rs -sSf | bash -s -- -y
 
-COPY .. /keripy
+COPY . /keripy
```
pfeairheller, line 16: This doesn't work from the Makefile.  Either revert this or fix the Makefile too.

pfeairheller (changes requested):

.github/workflows/python-app-ci.yml
s-a-tanjim: Upgrading runtime `node12 --> node16`. It won't break anything, just making sure everything is up-to-date while developing.

images/keripy.dockerfile
```diff
@@ -1,15 +1,19 @@
 
-FROM python:3.10.4-buster
+FROM python:3.10.4-alpine3.16
```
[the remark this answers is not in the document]
s-a-tanjim: Yes we can. I also ran the test, and it doesn't have any issue with the image.

pfeairheller (approved):

PR #413, opened 2022-12-27
fix: respect headDirPath from --config-dir

kentbull:
The headDirPath for each the `Regery`, `Baser`, and the `Keeper` is not set from the value specified by the `--config-dir` argument. 

This value, when present, should cause a `configing.Configer` instance to be created with the specified `headDirPath` value and this `Configer` instance should be used to specify both the `base` and `headDirPath` for the `Keeper`, the `Regery`, and the `Baser` instances created by `kli init [name]`.

The --config argument to init.py was never used so I added the "-dir" suffix to clarify the intent of the argument.

The Keeper did not recognize the headDirPath passed in through the Configer instance in **kwargs['cf']**. This was causing the Keeper to create files in either the default HeadDirPath or the AltHeadDirPath when everything else was being created in the configured `headDirPath`.


kentbull:
Looking back it is clear I misunderstood the purpose of the config-dir property by conflating it with headDirPath. I will close this PR.

PR #416, opened 2022-12-29
Unify incept and rotate KLI options from file config or CLI arguments

kentbull:
This adds in the support for using a configuration file (JSON) to configure a rotation event as well as support for using command line arguments to configure an inception event for all relevant configuration properties of each type of event.

The --file argument is no longer required for "incept" nor is it required for "rotate" though there are a minimum set of required attributes checked for "incept" that maps to the existing sample configuration "transferable-sample.json" file.

There are a few code cleanups like the `config.parse_data` for the "data" attribute parsing and the moving of parsing the file options to `config.load_file_options`.

### DIFF
--- a/scripts/demo/data/rotate-sample.json
+++ b/scripts/demo/data/rotate-sample.json
@@ -1,0 +1,10 @@
+{
+  "isith": "1",
+  "nsith": "1",
+  "ncount": 1,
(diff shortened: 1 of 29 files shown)


kentbull:
closing since I need to rebase on the latest development.

Issue #274, opened 2022-08-18
Unify command line argument for incept and rotate commands


pfeairheller:
Current inception command accepts only a file of configuration information for the new identifier and rotate command accepts only arguments specified individually on the command line.

We need to have both commands support both individual arguments on the command line and a configuration file, with individual argument overriding anything in the configuration file.

kentbull:
Came across a related issue when working on this: #412 and related PR #413.

kentbull:
This looks like it can be closed now.

pfeairheller:
Done by @kentbull

PR #417, opened 2022-12-29
Unify CLI args for incept and rotate

kentbull:
This adds in the support for using a configuration file (JSON) to configure a rotation event as well as support for using command line arguments to configure an inception event for all relevant configuration properties of each type of event.

The --file argument is no longer required for "incept" nor is it required for "rotate" though there are a minimum set of required attributes checked for "incept" that maps to the existing sample configuration "transferable-sample.json" file.

There are a few code cleanups like the `config.parse_data` for the "data" attribute parsing and the moving of parsing the file options to `config.load_file_options`.

This addresses #274

### DIFF
--- a/src/keri/app/cli/commands/rotate.py
+++ b/src/keri/app/cli/commands/rotate.py
@@ -4,12 +4,13 @@ keri.kli.commands module
 
 """
 import argparse
+from dataclasses import dataclass
 
 from hio.base import doing
 
 from keri import kering
-from keri.app.cli.common import rotating, existing
-from ... import habbing, agenting, indirecting, directing, delegating, forwarding
+from keri.app.cli.common import rotating, existing, config
+from ... import habbing, agenting, indirecting, delegating, forwarding
 
 parser = argparse.ArgumentParser(description='Rotate keys')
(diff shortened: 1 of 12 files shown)


src/keri/app/cli/commands/rotate.py
```diff
@@ -19,31 +20,89 @@
...
+def empty_options():
+    """
+    Empty rotation options used for merging file and command line options
+    """
+    return RotateOptions(
+        isith=None, ncount=None
+    )
+
+
+def merge_args_with_file(args):
```
pfeairheller, line 74: Convention in KERIPY is to use camelcase for method names, please fix

(I won't comment on all methods, but they all need to be fixed)

pfeairheller (changes requested):

src/keri/app/cli/commands/rotate.py
kentbull: I updated the method names to all be camelCase.

kentbull:
accidentally closed this, reopening.

pfeairheller (approved):

PR #423, opened 2023-01-02
Fix: FAILED test_configing.py::test_configer - AssertionError: assert False

s-a-tanjim:
If the user has root permission `configing.Configer()` creates the file inside `/usr/local/var/keri/cf/main/conf.json`. And `configing.Configer(headDirPath="/root/keri")` creates the file inside `/root/keri/keri/cf/main/conf.json`. But without the root permission, both `configing.Configer()` & `configing.Configer(headDirPath="/root/keri")` creates file inside `${HOME}/.keri/cf/main/conf.json`

### DIFF
--- a/tests/app/test_configing.py
+++ b/tests/app/test_configing.py
@@ -3,6 +3,7 @@
 tests.app.configin module
 """
 import os
+from pathlib import Path
 import shutil
 
 import pytest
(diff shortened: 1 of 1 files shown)


dhh1128:
@m00sey and @pfeairheller , can we get this merged?

rojones94:
Potentually a silly question but, why have all of the "." been removed from the "assert cfr.path.endswith("keri/cf/main/conf.json")" as in the pullrequest description the keri directory is intened to be a hidden directory.

s-a-tanjim:
I had the same thoughts. But it doesn't shows same behavior in all cases(Described above). I don't know why this is the case, but to make tests this behavior agnostic "." has been removed. Though fixing this behavior inside the code can be a better solution.

pfeairheller:
Sorry, I mistook this PR for one involving a Dockerfile so I sat on it until after the merge to main.

I think a better fix for this would be to raise an exception of headDirPath is explicitly set and the user does not have permission to write to it.  The current behavior was written to allow for multiple default values based on permissions, but I don't think its appropriate to assume a different default value if one if explicitly passed in.

pfeairheller:
Closing this PR as suggested.

PR #437, opened 2023-01-10
KEL Walking for Rotation

pfeairheller:
Implementation of KEL walking to resolve rotation participation of members with signing authority and rotation authority in group multisig AIDs.


SmithSamuelM:
Comment wording is wrong. The logic is walking the group Hab’s KEL not the member’s KEL

Issue #456, opened 2023-02-28
Cannot start witness(es) for local AID inception


tosirisuk:
This issue was taken from the KERIpy slack channel. The thread is [here](https://keriworld.slack.com/archives/C025C56K2SC/p1677507611116489).

I have set up 5 witnesses already. Then, I tried to incept a local AID (says Alice's AID)
I did two steps.

The first step is to initialize the Key Store `ks` using the `kli init` command, and I passed the `--config-file` during the `init `event as shown below (to initialize the key store to know about my 5 witnesses)
```
{
  "dt": "2023-02-26T12:57:59.823350+00:00",
  "iurls": [
    "http://13.213.0.148:5631/oobi/BC7qjybOGR7-UPVp-R-gqPmeeFY2eI6FYr-OE2OI8iMX",
    "http://13.213.0.148:5633/oobi/BLQOLhl3SAhuBEWbNT3CzyU2u4gesbpKx70Tzvkrk-1Y",
    "http://13.213.0.148:5635/oobi/BAHVSvATS6N8fQfV-UfJB_xZXzRuEL5mX7cWKdxNKVzK",
    "http://13.213.0.148:5637/oobi/BBmSeiRdPwwH3Ph5VSPJXFa1ub2zfYiR1XS7LUqGyJ_g",
    "http://13.213.0.148:5639/oobi/BMF2DG6OJ-En0Rb2UqSfZ0g5VtruxhIIrOE-XNcBprpR"
  ]
}
```

The output of the `kli init` command is
```
KERI Keystore created at: /home/ubuntu/.keri/ks/QAR-ALICE
KERI Database created at: /home/ubuntu/.keri/db/QAR-ALICE
KERI Credential Store created at: /home/ubuntu/.keri/reg/QAR-ALICE
        aeid: BN7nMtdIVbhhY8c6Vslof3AhvaVT1KQY2scX_XLZsW9t

Loading 5 OOBIs...
http://13.213.0.148:5631/oobi/BC7qjybOGR7-UPVp-R-gqPmeeFY2eI6FYr-OE2OI8iMX succeeded
http://13.213.0.148:5633/oobi/BLQOLhl3SAhuBEWbNT3CzyU2u4gesbpKx70Tzvkrk-1Y succeeded
http://13.213.0.148:5635/oobi/BAHVSvATS6N8fQfV-UfJB_xZXzRuEL5mX7cWKdxNKVzK succeeded
[...]

PR #457, opened 2023-03-01
feat: config witnesses by file and dir args

kentbull:
This aligns `kli witness start` with `kli agent start` for the `--config-dir` and `--config-file` args.

While this is functional there is one problem. There are empty config files written to `$CONFIG_DIR/keri/cf` while the files in `$CONFIG_DIR/keri/cf/main` are the ones that must be populated. This is annoying and I would rather eliminate this behavior before this is merged.

### DIFF
--- a/src/keri/app/cli/commands/witness/start.py
+++ b/src/keri/app/cli/commands/witness/start.py
@@ -10,7 +10,7 @@ import logging
 
 from keri import __version__
 from keri import help
-from keri.app import directing, indirecting, habbing, keeping
+from keri.app import directing, indirecting, habbing, keeping, configing
 from keri.app.cli.common import existing
 
 d = "Runs KERI witness controller.\n"
@@ -38,6 +38,12 @@ parser.add_argument('--base', '-b', help='additional optional prefix to file loc
 parser.add_argument('--alias', '-a', help='human readable alias for the new identifier prefix', required=True)
 parser.add_argument('--passcode', '-p', help='22 character encryption passcode for keystore (is not saved)',
                     dest="bran", default=None)  # passcode => bran
+parser.add_argument("--config-dir", "-c", dest="configDir", help="directory override for configuration data")
+parser.add_argument('--config-file',
+                    dest="configFile",
+                    action='store',
+                    default=None,
+                    help="configuration filename override")
 
 
 def launch(args):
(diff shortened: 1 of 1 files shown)


kentbull:
This resolved #456 for me.

kentbull:
Looks like I was wrong. It was the arguments I was passing in to the kli init command that were creating the file in ${CONFIG_DIR} :
```
kli init --name wan --salt 0AB3YW5uLXRoZS13aXRuZXNz --nopasscode \
  --config-dir "${KERI_SCRIPT_DIR}" --config-file wan-witness
```
I was missing a main/ path segment on the wan-witness file I was using.
Due to the default arguments for the Configer class it uses the path segment "main" when looking for configuration files.

```
class Configer(filing.Filer):
  ...
  TailDirPath = "keri/cf"
  ...
   def __init__(self, name="conf", base="main", filed=True, mode="r+b",
                 fext="json", human=True, **kwa):
    ...
```

The problem can be worked around (solved?) by just putting a prefix on the path given to the --config-file argument as so:

```
--config-file main/wan-witness
```

There are two potential --base arguments though only one exists in the CLI interface for kli witness start right now.
1. The base directory for the keystore. (in the .keri or /usr/local dir)
2. The base directory for the configuration files. (set to $KERI_SCRIPTS_DIR in the sample scripts)

In the future we could separate them out with another change to the `kli witness start` command by allowing `--base-keystore` and `--base-config` options.

src/keri/app/cli/commands/witness/start.py
```diff
@@ -54,13 +60,16 @@ def launch(args):
...
+               config_dir=args.config_dir,
+               config_file=args.config_file)
 
     logger.info("\n******* Ended Witness for %s listening: http/%s, tcp/%s"
                 ".******\n\n", args.name, args.http, args.tcp)
 
 
-def runWitness(name="witness", base="", alias="witness", bran="", tcp=5631, http=5632, expire=0.0):
+def runWitness(name="witness", base="", alias="witness", bran="", tcp=5631, http=5632, expire=0.0,
+               config_dir="", config_file=""):
```
pfeairheller, line 72: In KERIpy, camelcase is standard for parameters.  Can you update these two?

pfeairheller (changes requested):

src/keri/app/cli/commands/witness/start.py
kentbull: Yes, just pushed the update. I should get a pre-commit hook set up with a code style check.

pfeairheller (approved):

PR #458, opened 2023-03-03
Lock `libsodium` pwhash algo/params

jasoncolburne:
## Rationale

This PR locks the params we use for stretching keying material using libsodium. This will allow us to have more control over when the algorithm and params are updated, as they will be under ours. If the current Argon2 implementation is found to be flawed, it will likely be removed and the net effect will be the same. This just prevents libsodium from changing their defaults and breaking our implementation.

## Changes

Locks libsodium pwhash algorithm and params throughout.

## Testing

Created test vectors for all security tiers and verified they pass using the library constants and also the locked values.

### DIFF
--- a/tests/core/test_coring.py
+++ b/tests/core/test_coring.py
@@ -23,7 +23,7 @@ from keri.core import eventing
 from keri.core.coring import Ilkage, Ilks, Ids, Idents, Sadder
 from keri.core.coring import Seqner, NumDex, Number, Siger, Dater, Bexter
 from keri.core.coring import Serder, Tholder
-from keri.core.coring import Serialage, Serials, Vstrings
+from keri.core.coring import Serialage, Serials, Tiers, Vstrings
 from keri.core.coring import (Sizage, MtrDex, Matter, Xizage, IdrDex, IdxSigDex,
                               IdxCrtSigDex, IdxBthSigDex, Indexer,
                               CtrDex, Counter, sniff, ProDex)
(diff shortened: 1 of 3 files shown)


jasoncolburne:
I also had a thought to redefine our own constants somewhere but I wasn't sure what to name them or where to put them. In the end there weren't a lot of instances, so I'm not sure it's worth it.

tests/core/test_coring.py
```diff
@@ -4122,6 +4122,18 @@ def test_salter():
...
+    assert salter.stretch(temp=True) == b'\xd4@\xeb\xa6x\x86\xdf\x93\xd6C\xdc\xb8\xa6\x9b\x02\xafh\xc1m(L\xd6\xf6\x86YU>$[\xf9\xef\xc0'
+
+    salter = Salter(raw=raw)
+    assert salter.stretch(tier=Tiers.low) == b'\xf8e\x80\xbaX\x08\xb9\xba\xc6\x1e\x84\r\x1d\xac\xa7\\\x82Wc@`\x13\xfd\x024t\x8ct\xd3\x01\x19\xe9'
+
+    salter = Salter(raw=raw)
+    assert salter.stretch(tier=Tiers.med) == b',\xf3\x8c\xbb\xe9)\nSQ\xec\xad\x8c9?\xaf\xb8\xb0\xb3\xcdB\xda\xd8\xb6\xf7\r\xf6D}Z\xb9Y\x16'
+
+    salter = Salter(raw=raw)
+    assert salter.stretch(tier=Tiers.high) == b'(\xcd\xc4\xb85\xcd\xe8:\xfc\x00\x8b\xfd\xa6\tj.y\x98\x0b\x04\x1c\xe3hBc!I\xe49K\x16-'
```
jasoncolburne, line 4135: Re-reading this it's a bit ridiculous to instantiate the salter 4 times, I'll change that

PR #467, opened 2023-03-20
make kli errors easier to debug without changing code

dhh1128:


src/keri/app/cli/kli.py
```diff
@@ -25,7 +25,12 @@ def main():
...
     except Exception as ex:
-        print(f"ERR: {ex}")
+        import os
+        if os.getenv('DEBUG_KLI'):
+            import traceback
+            traceback.print_exc()
+        else:
+            print(f"ERR: {ex}")
         return -1
         # raise ex
```
pfeairheller, line 35: Minor nit... can we remove this commented line since its no longer needed!

pfeairheller (changes requested):
Been waiting for a clean solution to this.  If you could remove that one commented line, I'll merge ASAP.

src/keri/app/cli/kli.py
dhh1128: Done (commit 757b7cb)

pfeairheller (approved):

PR #497, opened 2023-05-08
Allow untargeted ACDCs before end of chain

jasoncolburne:
I think this could could be changed - I seems like it might be unintentional.

When I use this branch I can verify complex ACDC graphs we produce that have attestations in places other than the end of a chain. `development` is breaking my integration tests, and I'm hoping this was just an oversight.

If it makes sense please merge!

### DIFF
--- a/src/keri/vdr/verifying.py
+++ b/src/keri/vdr/verifying.py
@@ -401,6 +401,9 @@ class Verifier:
             op = 'I2I' if 'i' in creder.subject else 'NI2I'
 
         if op != 'NI2I':
+            if 'i' not in creder.subject:
+                return None
+
             iss = self.reger.subjs.get(keys=creder.subject['i'])
             if iss is None:
                 return None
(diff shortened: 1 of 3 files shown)


SmithSamuelM:
Do you have the NI2I operator on the edge?  The default if no operator given is I2I which requires targeted. NI2I allows untargeted so you have to specify that operator

jasoncolburne:
@SmithSamuelM Thanks for the reply! I forgot about those operators actually, let me add the correct one and see if KERIpy likes the chain.

pfeairheller:
Instead of changing the behavior to only check if available the proper check is to look at the operator of the edge and decide from there.  Let me know if you want to add that behavior in this PR or close it and open an issue for the full fix.

jasoncolburne:
I pushed some code. Let me know if it looks okay (default behaviour is [specified in the IETF KERI draft](https://trustoverip.github.io/tswg-acdc-specification/draft-ssmith-acdc.html#name-defaults-for-missing-operat)) and I'll try to write some tests. My integration succeeds with this operator code, and KERIpy tests still pass.

PR #498, opened 2023-05-09
Fix I2I operator

jasoncolburne:
I thought that `nodeSaid not in [i.qb64 for i in iss]` was actually checking something it wasn't. Turns out the issuer == node issuee check needs to be explicit. I found this when implementing parity in Rust. I noticed our credential graph failed to pass there but was passing here. This should fix things.

### DIFF
--- a/src/keri/vdr/verifying.py
+++ b/src/keri/vdr/verifying.py
@@ -196,7 +196,7 @@ class Verifier:
                     continue
                 nodeSaid = node["n"]
                 op = node['o'] if 'o' in node else None
-                state = self.verifyChain(nodeSaid, op)
+                state = self.verifyChain(nodeSaid, op, creder.issuer)
                 if state is None:
                     self.escrowMCE(creder, sadsigers, sadcigars)
                     self.cues.append(dict(kin="proof",  said=nodeSaid))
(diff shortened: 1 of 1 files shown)


SmithSamuelM:
@pfeairheller I can't really tell if this is OK or not. So waiting for you to OK the pull request.

src/keri/vdr/verifying.py
```diff
@@ -408,8 +408,9 @@ def verifyChain(self, nodeSaid, op):
             if iss is None:
                 return None
 
-            if op == 'I2I' and nodeSaid not in [i.qb64 for i in iss]:
-                return None
+            if op == 'I2I':
+                if issuer != creder.subject['i'] or nodeSaid not in [i.qb64 for i in iss]:
```
pfeairheller, line 412: I'm confused with the check for nodeSaid in iss wasn't enough.  That should be redundant to the check of issuer against creder.subject['i'].  If the subjs database is an index for look ups and not required to check for valid credential.  So just issuer != creder.subject['i'] is needed.

pfeairheller (changes requested):

src/keri/vdr/verifying.py
jasoncolburne: Good call, I thought subjs was validating more than it was. I've made the change, and tested against my acdc graphs and things look okay.

pfeairheller (approved):

PR #549, opened 2023-08-03
Dockerfile

tibor19:
This is the dockerfile branch that I created for keipy


tibor19:
@m00sey we can go with the multi-stage approach, however there are 2 things things in this pr missing from #563, the entry point doesn't start up the witness(which might not be needed, as this can be passed as parameter for the compose), and prameter filed was hard-coded, instead of being passed. I am happy with whatever choice you make.

pfeairheller:
I'm going to close this and submit a PR with the fix to configuring.py for the filed parameter.

PR #559, opened 2023-08-20
exn Message Upgrades.

pfeairheller:
This PR includes:

* Update to sign `exn` messages with full Seal to allow for long living exn messages.  Add sender, recipient and embeds to exn messages to support attaching signed event messages that don't get processed automatically.
* All multisig exn messages (icp, rot, ixn, vcp, iss, rvk) now contain embedded events with persistent exn messages.  Support for loading and verify persisted exn messages.
* Credential issuance now sends the credential embedded in the /credential/issue exn message with transposed signatures instead of streamed over.  There is also now a `kli vc accept` command to allow the issuee of a credential to accept or reject an issue credential.


SmithSamuelM:
@pfeairheller  

> credential embedded in the /credential/issue exn message with transposed signatures instead of streamed over.

Am I missing something. Because the credential issuance is anchored in a KEL which is signed, the credential itself does not need to be signed agains and signature attached, so there is no need for transposed signatures?

pfeairheller:
You are not missing anything, we have not removed the signatures from the credential, that is still an open issue.  I didn't want to tackle that issue with this PR.

PR #564, opened 2023-09-02
Peer to peer communication for passing events between group multisig participants

pfeairheller:
This PR includes:

* exn messages between group participants for initiating a group AID (inception)
* exn messages for credential registry inception between group participants.


m00sey (commented):
🚀

PR #567, opened 2023-09-05
IPEX handler

rodolfomiranda:


src/keri/vc/protocoling.py
```diff
@@ -405,3 +405,221 @@ def presentationExchangeExn(hab, reger, said):
...
+
+        super(IssueHandler, self).__init__(**kwa)
+
+    def recur(self, tyme):
+        if self.msgs:
+            msg = self.msgs.popleft()
+            serder = msg["serder"]
+            attrs = serder.ked["a"]
+
+            if not attrs["r"].startWith(self.resource):
```
pfeairheller, line 443: This check isn't necessary because messages won't end up here without the proper route, that't what the routing is for.  Think of it like an HTTP method handler, they don't need to check the path each time

src/keri/vc/protocoling.py
```diff
@@ -405,3 +405,221 @@ def presentationExchangeExn(hab, reger, said):
...
+    )
+
+    exn, end = exchanging.exchange(route="/ipex/apply", payload=data, sender=hab.pre)
+    ims = hab.endorse(serder=exn, last=False, pipelined=False)
+    del ims[:exn.size]
+    ims.extend(end)
+
+    return exn, ims
+
+def ipexOfferExn(hab, message, offer):
```
pfeairheller, line 482: I think the offer should include the ACDC and it should be signed and embedded.

pfeairheller (changes requested):
This is a great start!!  Just a few changes and we can merge this in as a starting point.  We will likely need to add embedded events some of the verbs, but time will tell.

src/keri/vc/protocoling.py
rodolfomiranda: removed

src/keri/vc/protocoling.py
rodolfomiranda: the offer is the metadata ACDC. I'll add as signed and embedded. Should it be called `offer` or `acdc`? 
We can also add the said in the payload.

src/keri/vc/protocoling.py
```diff
@@ -405,3 +405,221 @@ def presentationExchangeExn(hab, reger, said):
...
+    )
+
+    exn, end = exchanging.exchange(route="/ipex/agree", payload=data, sender=hab.pre)
+    ims = hab.endorse(serder=exn, last=False, pipelined=False)
+    del ims[:exn.size]
+    ims.extend(end)
+
+    return exn, ims
+    
+def ipexGrantExn(hab, message, acdc, iss):
```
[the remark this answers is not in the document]
rodolfomiranda: OK, that can be an `ixn` or a `rot`. Do you have a var name that covers both type of events or is it better to include the two and make them optional?

src/keri/vc/protocoling.py
pfeairheller: Either is fine but I like acdc better.

src/keri/vc/protocoling.py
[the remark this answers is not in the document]
pfeairheller: It doesn't matter to the protocol whether it's an ixn or rot so no need to make the caller choose. Let's go with anc for anchoring event and then it can be either. This avoids optional values in the exn.  

This isn't the approach I took with registry creation but it should be. I'll update that.

src/keri/vc/protocoling.py
rodolfomiranda: changed to acdc

src/keri/vc/protocoling.py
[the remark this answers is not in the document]
rodolfomiranda: `anc` added

pfeairheller (approved):

pfeairheller:
I think this PR is ready to merge as is.  @rodolfomiranda do you want to move it out off draft?

PR #569, opened 2023-09-06
Updates to multisig interaction and rotation protocols

pfeairheller:
This PR includes:

* Updates to multisig protocols to enable full interop with KERIA agents for creating, interacting and rotating multisig AIDs across client types.


m00sey (commented):
🚀

PR #570, opened 2023-09-07
Add witness TLS support

kentbull:
Adds the same args from `kli agent start` to `kli witness start`.

If the base function `createHttpServer` is accepted into KERIpy then we can change my PR to KERIA to use that function rather than duplicate it.

This is missing a positive test for creating a `tcp.ServerTls` instance. I just need to figure out how to generate self-signed TLS certs on the fly or add a test set of TLS certs+key to the test fixtures.

### DIFF
--- a/src/keri/app/cli/commands/witness/start.py
+++ b/src/keri/app/cli/commands/witness/start.py
@@ -44,6 +44,9 @@ parser.add_argument('--config-file',
                     action='store',
                     default=None,
                     help="configuration filename override")
+parser.add_argument("--keypath", action="store", required=False, default=None)
+parser.add_argument("--certpath", action="store", required=False, default=None)
+parser.add_argument("--cafilepath", action="store", required=False, default=None)
 
 
 def launch(args):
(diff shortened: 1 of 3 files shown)


m00sey:
@kentbull  generating certs could work, maybe someone like lets encrypt has a python lib.

Another option would be to use a mock library (like https://pypi.org/project/mockito/).

I'm happy to try put together a sample for you.

kentbull:
Sam has a test in his iofo/hio library I'm going to bring over here: https://github.com/ioflo/hio/blob/master/tests/core/tcp/test_tcp.py#L657

m00sey:
@kentbull you don't want to copy this test over, that would just mean we're retesting HIO in keripy.

You want to mock the `tcp.ServerTls` calls in keripy, as they're already tested in HIO

m00sey:
You might be able to get away with just monkey patching the tcp lib for this test

kentbull:
> You might be able to get away with just monkey patching the tcp lib for this test

I did just that:

```python
def test_createHttpServer(monkeypatch):
    port = 5632
    app = falcon.App()
    server = indirecting.createHttpServer(port, app)
    assert isinstance(server, http.Server)

    monkeypatch.setattr(hio.core.tcp, 'ServerTls', MockServerTls)
    monkeypatch.setattr(hio.core.http, 'Server', MockHttpServer)

    server = indirecting.createHttpServer(port, app, keypath='keypath', certpath='certpath', cafilepath='cafilepath')

    assert isinstance(server, MockHttpServer)
    assert isinstance(server.servant, MockServerTls)
```

PR #573, opened 2023-09-20
Interactive and non-Interactive multisig credential issuance with IPEX grant and admit

pfeairheller:
This PR includes:

* Multisig credental creation
* Multisig IPEX grant and admit message support
* Script (mutlisig-issuance.sh) exercising the functionality of multisig credential issuance with IPEX grant and admit messages having both participants agree out of band all all steps in the process (thus never executing the join command).
* Script (mutlisig-issuance-intractive.sh) exercising the functionality of multisig credential issuance with IPEX grant and admit messages having one participant lead each step and the second participant executing the join command.


m00sey (approved):
🚀

Issue #575, opened 2023-09-26
BadValsizeError (MBD_BAD_VALSIZE) error thrown when resolving a slightly incorrect oobi url


daidoji:
Not a big issue but I noticed that when I

1. Started the witness demo `kli witness demo`
2. Then retrieved the incorrect url `http://127.0.0.1:5642/oobi/` (notice the trailing slash that shouldn't be there)

I'd get a traceback Falcon error as below.

Ran this between the demo command and the traceback below.  `daidoji@darkstar:~$ curl http://127.0.0.1:5642/oobi/`

```bash
(keri) daidoji@darkstar:~/ssi/keripy$ kli witness demo &
[1] 31993
(keri) daidoji@darkstar:~/ssi/keripy$ Witness wan : BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha
Witness wil : BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM
Witness wes : BIKKuvBwpmDVA4Ds-EpL5bt9OqPzWPja2LigFYZN2YfX
Witness wit : BM35JN8XeJSEfpxopjn5jr7tAHCE5749f0OobhMLCorE
Witness wub : BIj15u5V11bkbtAxMA7gcNJZcax-7TgaBMLsQnMHpYHP
Witness wyz : BF2rZTW79z4IXocYRQnjjsOuvFUQv-ptCf8Yltd7PfsM
2023-09-26 09:44:16 [FALCON] [ERROR] GET /oobi/ => Traceback (most recent call last):
  File "/home/daidoji/ssi/keripy/src/keri/db/basing.py", line 62, in __getitem__
    return super(dbdict, self).__getitem__(k)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: ''

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "falcon/app.py", line 365, in falcon.app.App.__call__
  File "/home/daidoji/ssi/keripy/src/keri/end/ending.py", line 571, in on_get
    if aid not in self.hby.kevers:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/daidoji/ssi/keripy/src/keri/db/basing.py", line 78, in __contains__
    self.__getitem__(k)
  File "/home/daidoji/ssi/keripy/src/keri/db/basing.py", line 66, in __getitem__
[...]

PR #576, opened 2023-09-26
LMDBer methods throw KeyError on malformed keys.

daidoji:
_Note: I wish there was something a little more elegant to this fix, but it seemed better to keep the lmdb exceptions close to the classes and methods rather than trying something more fancy or a key_validation() type function to signal intent_

LMBD throws lmdb.BadValSizeError which isn't very helpful (or caught) in the codebase although KeyErrors are because of the dict abstraction elsewhere.  (Found while trying to resolve Issue #575)

So now the LMDB methods throw KeyError where appropriate and tests for all methods with empty keys now execute regardless of whether those methods had the issue or not.


m00sey (approved):
discussed on call

PR #577, opened 2023-09-26
Clean up aisle 6!

pfeairheller:
This PR includes:

* Updated script for issuing a credential from a group multisig in collaboration with a KERIA agent participant.
* Removal of the Mark I Agent (kiwiing, booting, kli commands and all references).  Much like Tony's original Mark I suit, this agent is now a relic placed in a case on the wall for posterity's sake, replaced by a technologically superior version... KERIA


m00sey (approved):
🧹

PR #596, opened 2023-11-08
Telquery fix

pfeairheller:
This PR includes:

* Update to the telquery feature to properly query when issuers only have Agents and not witnesses.


kentbull (approved):
LGTM

Issue #565, opened 2023-09-05
Liveness and readiness health check endpoints


kentbull:
Liveness checks and readiness checks facilitate stability and functionality for container deployment technology like Kubernetes for the following purposes:
1. Permitting routing of requests to a Kubernetes Pod only when it is ready to receive requests. This avoids unnecessarily causing client side errors or potential server side crashes.
2. Continuous health monitoring of pods. When a pod crashes or becomes unstable then Kubernetes will detect the poor health and automatically restart it. This may or may not make problems worse, though it usually makes them better.

Doing health checks in KERIpy for Witnesses and Watchers needs to address the following concerns:
1. A health check for the witness that it can receive HTTP requests

kentbull:
This can be accomplished with the blind OOBI with a GET to `https://witness_host:port/oobi`

PR #627, opened 2023-12-07
Adds simple healthcheck

kentbull:
Adds a simple healthcheck endpoint at each of the ports that is allowed through the SignifyTS header check.

This is not a fully involved healthcheck that has useful statistics on the Habery, Habs, or other things related to the witness. That would be cool and will happen in a future evolution of the healthcheck. This endpoint is just to support a simple runtime status reporting that the application is running.

This is useful for Kubernetes readinessProbes and livenessProbes.

### DIFF
--- a/src/keri/end/ending.py
+++ b/src/keri/end/ending.py
@@ -23,7 +23,7 @@ from .. import help
 from .. import kering
 from ..app import habbing
 from ..core import coring
-from ..help import helping
+from ..help import helping, nowIso8601
 
 logger = help.ogler.getLogger()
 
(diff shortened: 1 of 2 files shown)


kentbull:
Addresses #565

pfeairheller:
I believe you intend this for witnesses and I don't think its necessary.  Why not just use the blind OOBI?

kentbull:
I don't yet know how to use the blind OOBI. How would I use that? Is it just a GET to `https://witness_host:port/oobi?`

What I'm wanting is something that Kubernetes can call without knowing anything particular about an AID. I see the OOBIEnd.on_get and it looks like it would work for a healthcheck. Will the default AID always be set for a witness as shown below? If so then this would work as a healthcheck.

```python
class OOBIEnd:
        def __init__(self, hby: habbing.Habery, default=None):
        """ ...
            default (str) qb64 AID of the 'self' of the node for
        """
        self.hby = hby
        self.default = default


    def on_get(self, req, rep, aid=None, role=None, eid=None):
        if aid is None:
            if self.default is None:
                rep.status = falcon.HTTP_NOT_FOUND
                rep.text = "no blind oobi for this node"
                return

            aid = self.default
    ...
```

m00sey:
As discussed on 12/14 closing.

Thanks @kentbull !

PR #628, opened 2023-12-08
Cache docker layers in CI build

kentbull:
This should speed up successive Docker image builds.

### DIFF
--- a/.github/workflows/publish-keripy.yml
+++ b/.github/workflows/publish-keripy.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check out the repo
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Log in to Docker Hub
         uses: docker/login-action@v2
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-keripy.yml
```diff
@@ -26,6 +26,18 @@ jobs:
...
+        id: buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Cache Docker Layers
+        uses: actions/cache@v3
+        with:
+          path: /tmp/.buildx-cache
+          key: keri-${{ runner.os }}-buildx-${{ github.sha }}
+          restore-keys: |
+            keri-${{ runner.os }}-buildx-
```
pfeairheller, line 39: Is something missing from the end of this line?

pfeairheller (commented):
We don't deploy using this workflow so I have no way to test this.

.github/workflows/publish-keripy.yml
kentbull: No, it's just a match prefix. 

The reason it doesn't make sense to include the `${{ github.sha }}` value in match line (restore-keys) the line is because you only want to invalidate the Docker cache when the internals of the Docker Build command detect a change needed. 

The `${{ github.sha }}` value changes on every new commit which would mean an entirely new cache folder on every commit which defeats the purpose of caching. The value you can gain with caching only comes from the intermediary Docker image layers prior to the COPY command for copying source code in.

Typically this only happens when you make a change to the Dockerfile or when something copied in with a COPY command changes.

kentbull:
Which workflow is used for deployment? I'll copy this over.

kentbull:
I see the [python-app-ci.yaml](https://github.com/WebOfTrust/keripy/blob/development/.github/workflows/python-app-ci.yml)

It's not too much different to integrate it into a docker build command though you do have to change it to use `docker buildx` rather than docker build, which is pretty much the same.

I see the `--no-cache` option in the `docker build` command though:

```yaml
docker build -f .github/workflows/interop/Dockerfile --no-cache -t ${{ env.GITHUB_REPOSITORY_NAME }}/keripy-interop:${{ env.CACHE_KEY_INTEROP }} .
```
Is the `--no-cache` necessary? If so then we can close this PR as caching would be irrelevant.

PR #671, opened 2024-02-06
Use image python:3.10-alpine; Fix docker COPY build failure

clackwork:
By using a less specific base image we can be right more often while doing less maintenance grunt-work on our Docker image. By using 3.10 instead of 3.10.13, our code will still be correct when 3.10.14 is released. Similarly, by using "alpine" instead of "alpine 3.18", we are still presently using the latest stable version of Alpine 3 (3.18) while being upgraded automatically to the latest stable version when it becomes available.

Reference:

- https://hub.docker.com/layers/library/python/3.10-alpine/images/sha256-4b062a3861ac865436439499d9a03e055871a0f465833987d23b67364a35a38f?context=repo
- https://github.com/docker-library/python/blob/b34fc81375758e84dec797e9210a079b4889352e/3.10/alpine3.18/Dockerfile#L24

Also, fix docker build bug with COPY:

> Step 9/17 : COPY requirements.txt setup.py .
> When using COPY with more than one source file, the destination must be a directory and end with a /

### DIFF
--- a/images/keripy.dockerfile
+++ b/images/keripy.dockerfile
@@ -1,5 +1,5 @@
 # Builder layer
-FROM python:3.10.13-alpine3.18 as builder
+FROM python:3.10-alpine as builder
 
 # Install compilation dependencies
 RUN apk --no-cache add \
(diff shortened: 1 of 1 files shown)


pfeairheller:
Not sure why this was tagged pre-release, it is based off of development.  If you want it in pre-release, it needs to be changed to the release branch.

m00sey:
I thought I said during the call it was tagged for discussion.

pfeairheller (approved):

pfeairheller:
Apologies, I merged this into development instead of the release branch so I had to revert it.  I've included these changes in my current PR that I'll merge this morning

PR #701, opened 2024-03-06
Non deterministic prefixes in witnesses

daidoji:
Removed hardcoded salts as default behavior of openHby and openHab.

Also fixed a bug that hadn't been caught yet wherein whatever was passed to openHby was always used and a default salt was never created. Adjusted to match the intention of the code and no tests seemed to break from it so it must have been getting caught further down the stack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -140,3 +140,6 @@ config/
 
 # keripy virtualenv
 keripy/
+
+# ctags
+.tags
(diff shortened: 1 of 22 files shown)


SmithSamuelM (approved):
Yes this is better. The hardcoded Salts should be in the tests and not in the default code.  This is better for security. Thanks for changing all the tests.

PR #723, opened 2024-03-25
Catching socket exception when creating witness http client

rodolfomiranda:
This PR is a fix for **KERIA** issue [#214](https://github.com/WebOfTrust/keria/issues/214) where a `socket.getaddrinfo` exception is raised in certain conditions causing the witness http client creation to fail.

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -74,10 +74,13 @@ class Receiptor(doing.DoDoer):
         clients = dict()
         doers = []
         for wit in wits:
-            client, clientDoer = httpClient(hab, wit)
-            clients[wit] = client
-            doers.append(clientDoer)
-            self.extend([clientDoer])
+            try:
+                client, clientDoer = httpClient(hab, wit)
+                clients[wit] = client
+                doers.append(clientDoer)
+                self.extend([clientDoer])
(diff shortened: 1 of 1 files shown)


src/keri/app/agenting.py
```diff
@@ -74,10 +74,13 @@ def receipt(self, pre, sn=None):
...
-            client, clientDoer = httpClient(hab, wit)
-            clients[wit] = client
-            doers.append(clientDoer)
-            self.extend([clientDoer])
+            try:
+                client, clientDoer = httpClient(hab, wit)
+                clients[wit] = client
+                doers.append(clientDoer)
+                self.extend([clientDoer])
+            except Exception as e:
```
pfeairheller, line 82: Can we catch something more specific here?

pfeairheller (changes requested):

src/keri/app/agenting.py
rodolfomiranda: yes, but I need to import `socket`. Is that ok.

rodolfomiranda: I confined it to  two specific exceptions

pfeairheller (approved):

PR #738, opened 2024-04-04
Fix for failing Reply End Roles escrows

rodolfomiranda:
Sometimes when escrowing reply end roles, a race condition causes that the updated `end` in the db is interpreted as an old end, and then the validation failed because the datetime of the new message has not increased. When that happens, the escrow of the reply message can not be confirmed and fails until timeouts. 
In this PR I'm adding a validation to make sure that the end retrieved from the DB is actually an old end and not the same that we are trying to authorize.

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -4285,7 +4285,8 @@ class Kevery:
         aid = cid  # authorizing attribution id
         keys = (aid, role, eid)
         osaider = self.db.eans.get(keys=keys)  # get old said if any
-        if osaider is not None and osaider.qb64b == saider.qb64b: osaider = None
+        if osaider is not None and osaider.qb64b == saider.qb64b: 
+            osaider = None
         # BADA Logic
         accepted = self.rvy.acceptReply(serder=serder, saider=saider, route=route,
                                         aid=aid, osaider=osaider, cigars=cigars,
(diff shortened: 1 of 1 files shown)


src/keri/core/eventing.py
```diff
@@ -4285,6 +4285,7 @@ def processReplyEndRole(self, *, serder, saider, route,
         aid = cid  # authorizing attribution id
         keys = (aid, role, eid)
         osaider = self.db.eans.get(keys=keys)  # get old said if any
+        if osaider is not None and osaider.qb64b == saider.qb64b: osaider = None
```
pfeairheller, line 4288: 2 things...

1.  please wrap if statements like this otherwise it triggers a PEP warning.
2. We need to discuss this as I believe this breaks BADA RUN logic.

pfeairheller (changes requested):
Can we get a unit test that triggers the bug being fixed?

src/keri/core/eventing.py
rodolfomiranda: 1- done
2- Yes, I need your careful review. I think it is not breaking bada run since it's preventing to compare the same said (old = new), so you'll be updating/authorizing the same aid-role-eid value again.

rodolfomiranda:
> Can we get a unit test that triggers the bug being fixed?

I hit that problem working with a multisig in KERIA/Signify, and the problem is an escrow that never completes. Not sure if that can be part of a unit test. I'll take a look.

PR #740, opened 2024-04-05
Updated docker images to use python 3.12

daidoji:
Not sure how we test docker images but these build and seem to be the latest alpine images that matched the original image.  The witness demo uses the GLEIF images so wasn't sure where to update those.

### DIFF
--- a/images/keripy.dockerfile
+++ b/images/keripy.dockerfile
@@ -1,5 +1,5 @@
 # Builder layer
-FROM python:3.10-alpine as builder
+FROM python:3.12-alpine as builder
 
 # Install compilation dependencies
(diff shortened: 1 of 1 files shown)


2byrds:
> Not sure how we test docker images but these build and seem to be the latest alpine images that matched the original image. The witness demo uses the GLEIF images so wasn't sure where to update those.

Good stuff. The GLIEF images can hopefully be ignored by the community now that the WebOfTrust images are getting updated/used more often.

pfeairheller:
Thanks @daidoji.  The plan was to do this when we cut a release off of development since we don't publish docker images for this branch.

PR #749, opened 2024-04-09
Process only saved credentials during migration

Arsh-Sandhu:
- update migrate command to process only saved credentials
- while migrating keys, skip missing events


pfeairheller (approved):
Good stuff.  Thanks to Provenant for catching these and creating a fix.

PR #750, opened 2024-04-11
Secure Witness Support

pfeairheller:
This PR includes:

- Enhanced interact and rotate kli commands to allow for --authenticate parameter to ask user for a 2-factor auth code to send to witnesses for secure witness interactions.
- Enhancement to streaming CESR or HTTP to allow for path components in location records.
- Update to receipt endpoint interaction to honor path components in location records to allow for easier multi-tenant witness solutions.


SmithSamuelM (commented):
I noticed that you are using Saider to merely store a SAID. As we discussed some time ago unless one is going to use the Saider methods to create a SAID from some SAD dict. There is no benefit to serializaing and deserlizing via Saider. SuberBase is adequate for the string that is the SAID.  If one needs to convert from qb64 to qb2 or raw—code then Diger is more appropriate for that. Since Saids are digests.  I am trying to minimize the use of Saider because its functionality except for corner cases is now included in Serder.
So if we use it everywhere we store a SAID digest in a database then it will become much more difficult to pull out later.  
As far as I can tell in most instances the SAID stored as a value is just an index to be pulled out an then used in the keyspace for yet another database. So when db value use is to create a keyspace to reference the actual data. then serializing as a Saider obscures its purpose.  In LMDB there is really just one space. keyspace but its split between keys and values and so only when the value is some datastructure or a tuple of values is there any benefit to using CESRSuber or SerderSuber. If the value is just a string that is to be used as an index to another table then no reason to incur the serialization/deserialization cost of converting to/vrom a Matter subclass. Just keep it a string.

The purpose of CesrSuber was so we could parse a concatenation of a tuple of values. And to parse them we need the Matter classes who do the heavy lifting. But when we are just storing indexes as single values its a waste of effort. The point was not to convert everything to CESRSubers. The ORM like functionality of having a standard api of get pin put rem is provided just fine by Suber. The Orm is Suber not CESR Suber. I hope this helps.

pfeairheller:
Fixed.

m00sey (approved):
Impressive

Issue #690, opened 2024-02-23
Python 3.12 more strict on Regex: SyntaxWarning: invalid escape sequence '\-'  B64REX = b'^[A-Za-z0-9\-\_]*\Z' in src/keri/core/coring.py:142


kentbull:
Python 3.12 is more strict on regular expressions.
In src/keri/core/coring.py:142 the following regular expression shows a syntax warning in my console every time I run a witness:
```python
src/keri/core/coring.py:142: SyntaxWarning: invalid escape sequence '\-'
  B64REX = b'^[A-Za-z0-9\-\_]*\Z'
```

Some solutions here are:
1. Use a raw string (backslashes not needed):
```python
B64REX = rb'^[A-Za-z0-9-_]*\Z'
```
2. escape the slash
```python
B64REX = b'^[A-Za-z0-9\\-_]*\\Z'
```

SmithSamuelM:
```pythonh
b'^[A-Za-z0-9_-]*\Z'
```

Should work.  I dont have any problems with 3.12

As long as the minus - is the first or last character in a group it doesn’t have to be escaped

daidoji:
@kentbull this has been fixed.

PR #754, opened 2024-04-12
690 Fix SyntaxWarning: invalid escape sequence

dc7:
Use a raw byte string for the regex so that \Z is properly interpreted as an anchor representing the end of the input, not an escape sequence for the character \Z (which doesn't exist). The hyphen was already moved to the end of the character class so we don't need to worry about escaping that.

### DIFF
--- a/src/keri/help/helping.py
+++ b/src/keri/help/helping.py
@@ -264,7 +264,7 @@ B64IdxByChr = {char: index for index, char in B64ChrByIdx.items()}
 B64_CHARS = tuple(B64ChrByIdx.values())  # tuple of characters in Base64
 
 
-B64REX = b'^[0-9A-Za-z_-]*\Z'   # [A-Za-z0-9\-\_]
+B64REX = rb'^[0-9A-Za-z_-]*\Z'   # [A-Za-z0-9\-\_]
 Reb64 = re.compile(B64REX)  # compile is faster
 # to use if Reb64.match(bext):
 


dc7:
#690

SmithSamuelM:
See https://docs.python.org/3/library/re.html. Search for \Z. “Matches only at end of string”.  this ensures that the whole string is consumed. Not sure why you are getting an invalid escape sequence double check your IDE . I suspect it is in error. This code works with units tests that do not throuw errors. So not sure what system you are using that causes your error.

dc7:
> According to re docs if a regex escape sequence is not a valid escape sequence in python then it will be interpreted specially as regex and does not need to be doubly escaped as is the case for a valid python escape sequence in a bytes not raw bytes.

This behavior was [deprecated in Python 3.6](https://docs.python.org/dev/whatsnew/3.6.html#deprecated-python-behavior) and has been [upgraded to a warning as of Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#other-language-changes). Not IDE-related, you can see it in [the test logs here](https://github.com/WebOfTrust/keripy/actions/runs/8656442458/job/23736974567#step:6:76). Want me to fix the other SyntaxWarnings in the same PR? I just noticed there are two more. I'm Daniel Comisar by the way, I'm working with Charles Lanahan on his startup.

SmithSamuelM:
@dc7. thanks good to know.

SmithSamuelM (approved):

PR #761, opened 2024-04-17
Fixed missing verbose output from ipex list

pSchlarb:


src/keri/app/cli/commands/ipex/list.py
```diff
@@ -199,6 +200,12 @@ def grant(self, exn):
         if response is not None:
             print(f"    Response: {responseType} ({response.qb64})")
 
+        if self.verbose:
+            bsad = json.dumps(sad, indent=2)
+            print("    Full Credential:")
```
daidoji, line 205: You could do it this way but probably better to use pprint.  https://docs.python.org/3/library/pprint.html

pSchlarb: I just oriented myself on the list command as it is on the vc list, which can be found at `src/keri/app/cli/commands/vc/list.py`.
I can change to pprint, should i also change it there(vc list)?

daidoji: I don't think you should change it there.  Most of the commands are made with the intention of consumption from shell-style scripts and pretty printing there is probably just making life harder for the consumers of keripy commands.  In fact one might say that its probably inappropriate in the log file too, but logging isn't exactly 100% in this project so its probably fine there.  

I was just commenting from a python library standpoint pprint is the right way to go if printing in a visually easy way is what you're trying to accomplish.

pSchlarb: Just thinking: with the terminal and shell scripts in mind wouldn't then for example a `--oneline` argument be a meaningful improvement (printing the credential in oneline)?
That would make the output also more grep-friendly.

Also with that in mind i am unsure if the conversion of pprint to put everything in single quotes is a wanted feature here? 
 
I am open to ideas and input.

daidoji: Yeah, that's probably a wider discussion above my paygrade but I'd be for it.  Since the kli tool is mostly consumed by shell scripts making it easier for script writers seems like a reasonable thing to do.  I'm actually not sure what all the current commands are doing and outputting but harmony among them would be good too.

pSchlarb: Hm maybe it is a thing that needs to be discussed in the implementors call?
I will gladly join again next week for that.

PR #782, opened 2024-05-14
feat: multisig join --group arg

kentbull:
This adds the capability to specify an alias during multisig join. Previously the alias was always set to "test alias" which would cause alias collision problems when doing successive multisig join commands for different identifiers.

### DIFF
--- a/src/keri/app/cli/commands/multisig/join.py
+++ b/src/keri/app/cli/commands/multisig/join.py
@@ -21,47 +21,52 @@ from keri.vdr import verifying, credentialing
 logger = help.ogler.getLogger()
 
 parser = argparse.ArgumentParser(description='Join group multisig inception, rotation or interaction event.')
-parser.set_defaults(handler=lambda args: confirm(args))
+parser.set_defaults(handler=lambda args: join(args))
 parser.add_argument('--name', '-n', help='keystore name and file location of KERI keystore', required=True)
 parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
                     required=False, default="")
+parser.add_argument('--alias', '-a', help='human readable alias for the multisig identifier prefix', required=False, default=None)
 parser.add_argument('--passcode', '-p', help='22 character encryption passcode for keystore (is not saved)',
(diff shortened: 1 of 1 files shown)


src/keri/app/cli/commands/multisig/join.py
```diff
@@ -196,7 +201,12 @@ def incept(self, attrs):
 
         if approve:
             if self.auto:
-                alias = "test alias"
+                alias = self.alias
```
pfeairheller, line 204: This does not account for auto being true and alias being None which will cause an exception below. (As happened in the scripts run for this PR).

src/keri/app/cli/commands/multisig/join.py
```diff
@@ -196,7 +201,12 @@ def incept(self, attrs):
...
+            elif self.alias:
```
pfeairheller, line 205: self.alias should only be used if self.auto is True.

pfeairheller (changes requested):
Please fix the logic so that the alias is only used if self.auto is True.  I would also recommend using a different argument name as --alias has specific meaning in most other commands.

src/keri/app/cli/commands/multisig/join.py
kentbull: This was changed in the most recent commit to have a nested if statement checking of the alias (now "group") is set.
For example:
```python
            if self.auto:
                if self.group is None:
                    group = "default-group"
                else:
                    group = self.group
```

src/keri/app/cli/commands/multisig/join.py
kentbull: This was changed in the most recent commit to have a nested if statement checking of the alias (now "group") is set.
For example:
```python
            if self.auto:
                if self.group is None:
                    group = "default-group"
                else:
                    group = self.group
```

kentbull:
I changed the argument name from `--alias` to be `--group` to align with the argument name sent in to `self.hby.makeGroupHab`.

pfeairheller (approved):

PR #783, opened 2024-05-14
feat: multisig join --group arg

kentbull:
This adds the capability to specify a multisig group name during `kli multisig join`. Previously the group name was always set to "test alias" which would cause a name collision problems when doing successive `multisig join` commands for different identifiers.

This is similar to the PR to v1.1.14: #782 

There were merge conflicts so the commits couldn't be directly cherry picked.


src/keri/app/cli/commands/multisig/join.py
```diff
@@ -196,17 +201,20 @@ def incept(self, attrs):
 
         if approve:
             if self.auto:
-                alias = "test alias"
+                if self.group is None:
```
pfeairheller, line 204: This is so much better than what we had here!

pfeairheller (approved):

PR #789, opened 2024-05-29
fixes MissingDelegableApprovalError logging bug

2byrds:
Adds a message to the MissingDelegableApprovalError so the logging doesn't fail.
Updated test to exercise the exception and logging.

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -5806,7 +5806,7 @@ class Kevery:
                     self.processEvent(serder=eserder, sigers=sigers, wigers=wigers, delseqner=seqner,
                                       delsaider=saider, local=esr.local)
                 else:
-                    raise MissingDelegableApprovalError()
+                    raise MissingDelegableApprovalError("No delegation seal found for event.")
 
             except MissingDelegableApprovalError as ex:
                 # still waiting on missing delegation approval
--- a/tests/core/test_delegating.py
+++ b/tests/core/test_delegating.py
@@ -739,6 +739,8 @@ def test_delegables_escrow():
         parsing.Parser().parse(ims=bytearray(gateIcp), kvy=torKvy, local=True)
         assert gateHab.pre not in torKvy.kevers
         assert len(torHab.db.delegables.get(keys=snKey(gateHab.kever.serder.preb, gateHab.kever.serder.sn))) == 1
+        # Exercise the MissingDelegableApprovalError case
+        torKvy.processEscrowDelegables()
 
         # Now create delegating interaction event
         seal = eventing.SealEvent(i=gateHab.pre,


tests/core/test_delegating.py
```diff
@@ -739,6 +739,8 @@ def test_delegables_escrow():
         parsing.Parser().parse(ims=bytearray(gateIcp), kvy=torKvy, local=True)
         assert gateHab.pre not in torKvy.kevers
         assert len(torHab.db.delegables.get(keys=snKey(gateHab.kever.serder.preb, gateHab.kever.serder.sn))) == 1
+        # Exercise the MissingDelegableApprovalError case
+        torKvy.processEscrowDelegables()
```
m00sey, line 743: Is there expected to be additional assertions here to catch `MissingDelegableApprovalError`?

m00sey: or should there be additional assertions to verify there _wasn't_ a missing delegable approval error?

m00sey: seems odd to add the processing line but no assertions in the test

src/keri/core/eventing.py
```diff
@@ -5806,7 +5806,7 @@ def processEscrowDelegables(self):
                     self.processEvent(serder=eserder, sigers=sigers, wigers=wigers, delseqner=seqner,
                                       delsaider=saider, local=esr.local)
                 else:
-                    raise MissingDelegableApprovalError()
+                    raise MissingDelegableApprovalError("No delegation seal found for event.")
```
m00sey, line 5809: might be good to include the event digest?



```MissingDelegableApprovalError("Missing Delegable Approval for evt at dig = {}.""".format(bytes(edig)))```

tests/core/test_delegating.py
2byrds: > Is there expected to be additional assertions here to catch `MissingDelegableApprovalError`?

No, the error is part of the escrow processing (the exception doesn't make it out of the processing).

2byrds: > or should there be additional assertions to verify there _wasn't_ a missing delegable approval error?

The test was valid originally, but it never exercised the MissingDelegableApprovalError so it did discover the logging error. This small addition does a better job of simulating that processEscrowDelegables is being called multiple times during the flow (becuase it is part of asynchronous escrow processing)

2byrds: > seems odd to add the processing line but no assertions in the test

The constantly running escrow processors are difficult to capture in a small test. this is the shortest path to exercising the case when the processor runs before the user managed flow is complete.

src/keri/core/eventing.py
2byrds: Yes, that would be better

PR #793, opened 2024-06-02
adds reconfigured documentation, make file additions - needs to be in…

m00sey:
…coperated into future builds


m00sey:
I'll demo this on Tuesdays dev call.

PR #797, opened 2024-06-05
fix: re-cue unmatched cues in Respondant

iFergal:
Should resolve https://github.com/WebOfTrust/keria/issues/184.

In KERIA, the Respondant processes certain cues for a kevery, but it drops other cues like `keyStateSaved` and Phil has suggested this to be the safest change to make to resolve the issue.

Since we are pushing the cue back onto the Deck, I needed to process one item per tock instead of looping the entire Deck and then yielding. An alternative is to keep another list of items and re-add them all after `while self.cues` is finished but I find this risky in case there are a lot of items and the process dies half way through.

### DIFF
--- a/src/keri/app/storing.py
+++ b/src/keri/app/storing.py
@@ -219,65 +219,70 @@ class Respondant(doing.DoDoer):
         _ = (yield self.tock)
 
         while True:
-            while self.cues:  # iteratively process each cue in cues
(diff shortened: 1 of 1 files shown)


pfeairheller (approved):
Looks good to me, exactly what we discussed in Discord.

PR #798, opened 2024-06-07
fix alias renaming

SabbirIrfan:
when i was trying to rename with the command `kli rename --name test --base "${KERI_TEMP_DIR}"  --alias <old-alias> <new-alias>` i didn't get the expected result. I think this PR can fix this issue.


m00sey:
Would be great to see a test added for this.

pfeairheller (changes requested):
Please add a script to the scripts/demo/basic directory that tests this and add that test to test_scripts.sh so it runs as part of our CI.

SabbirIrfan:
sorry! there was a unnecessary echo.

pfeairheller:
Can you please add execute permission to the new script.

Issue #799, opened 2024-06-12
Credential received are not validated if the registry is local


rodolfomiranda:
### Version

All

### Expected behavior

Use case: the QARs of a multisig QVI issuing a LE credential to a legal entity multisig conformed by 4 LARs, where  two of the LARs  are also QARs. We expect that all LARs receive and store the credential in their wallet. 

### Actual behavior

The credential is received by the LAR that is also a QAR but is rejected by this part of the code:
```
if self.local:
   if regk not in self.registries:  # nonlocal event when in local mode
      raise ValueError("Nonlocal event regk={} when local mode for registries={}."
         "".format(regk, self.registries))
```
Quote from Phil as a clue: "we should probably figure out how to reconcile a credential registry that is being used to issue to an AID in its own agent/wallet"
The workaround it to create separate wallets for QARs and LARs

### Steps to reproduce

I'll add a script.

rodolfomiranda:
PR submitted: #806

PR #806, opened 2024-06-23
Allow local tever events when not in local mode

rodolfomiranda:
This PR solves issue https://github.com/WebOfTrust/keripy/issues/799
In KERIA, when a credential is issued to someone that is also part of the issuer, the credential is rejected by the issuee with `Local event regk={} when nonlocal mode`. 
I understand that this should be a valid case when the event is received as non local, so I'm proposing to remove the check that prevents to continue with the event processing.

### DIFF
--- a/src/keri/vdr/eventing.py
+++ b/src/keri/vdr/eventing.py
@@ -1559,8 +1559,7 @@ class Tevery:
                                      "".format(regk, self.registries))
             else:
                 if regk in self.registries:  # local event when not in local mode
-                    raise ValueError("Local event regk={} when nonlocal mode."
-                                     "".format(regk))
+                    pass
 
         if regk not in self.tevers:  # first seen for this registry
             if ilk in [Ilks.vcp]:


src/keri/vdr/eventing.py
```diff
@@ -1559,8 +1559,7 @@ def processEvent(self, serder, seqner=None, saider=None, wigers=None):
                                      "".format(regk, self.registries))
             else:
                 if regk in self.registries:  # local event when not in local mode
-                    raise ValueError("Local event regk={} when nonlocal mode."
-                                     "".format(regk))
+                    pass
```
pfeairheller, line 1562: This is not the proper fix for this as it disables the features provided by both the `lax` and the `local` properties of the Tevery.  Please use those flags when launching the Tevery to accomplish this

pfeairheller (changes requested):
This should not be a KERIpy PR but changes to the use of Tevery in other repos, most likely KERIA

src/keri/vdr/eventing.py
rodolfomiranda: Ok, the problem is that in KERIA we have one global `Tevery` process that is shared and used in the escrows. It's instantiated as non local. Do we need to check if the event is local or non local before launching a Tevery? I'll take a look on how that can be implemented in KERIA  if that's the way to go.

pfeairheller: Yes, you can create 2 Tevery objects.  One to use for local events that you create and one for remote messages that is used to process incoming messages.

rodolfomiranda: Ok. I'll look into that direction. The main problem to solve is that the incoming message received is the one that has the local event. We need to detect that before deciding which Tevery object should be called.

pfeairheller:
Can this PR be closed?

PR #810, opened 2024-06-28
Externalize the size limit for LMDB map size for Baser to an environment variable.

pfeairheller:
This PR exposes the KERI_BASER_MAP_SIZE environment variable to allow setting the maximum file size for the main KERIpy LMDB database.  By default it is set to 100MB, but LMDB documentation suggests on 64-bit systems it is safe to set it as high as a terabyte or more as long as you have the disk space to handle.

You can also change this value and relaunch any process using KERIpy and it will take effect.


m00sey (approved):
🥔

PR #812, opened 2024-06-29
Fixed KERI_BASER_MAP_SIZE not integer bug.

### LINKED ISSUE #811: KERI_BASER_MAP_SIZE isn't an integer tests fail
### Version

-e git+https://github.com/WebOfTrust/keripy.git@eb43ff833ce940a94615e325da8932a842c10c86#egg=keri

### Environment

Ubuntu, python 3.12.2, local

### Expected behavior

KERI_BASER_MAP_SIZE passed a bad value keripy shouldn't fall over.  A log message should be displayed and a fallback value should be chosen (or alternatively program shutdown should occur).

### Actual behavior

when say KERI_BASER_MAP_SIZE="foo" or KERI_BASER_MAP_SIZE="1.0" tests fail to pass and program functionality impacted because of exception raised.

### Steps to reproduce

```
kli init --name baz --nopass
kli incept --name foo --alias foo_alias --icount 1 --isith 1 --ncount 0 --nsith 0 --toad 0
```

daidoji:
Added log message and exception check for when KERI_BASER_MAP_SIZE not an integer.

Fixes #811

### DIFF
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -942,7 +942,8 @@ class Baser(dbing.LMDBer):
             try:
                 self.MapSize = int(mapSize)
             except ValueError:
-                logger.exception("KERI_BASER_MAP_SIZE must be an integer value!  Will continue to use default keripy value of 104857600!")
+                logger.error("KERI_BASER_MAP_SIZE must be an integer value >1!")
+                raise
 
         super(Baser, self).__init__(headDirPath=headDirPath, reopen=reopen, **kwa)
 
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -939,7 +939,10 @@ class Baser(dbing.LMDBer):
         self._kevers.db = self  # assign db for read through cache of kevers
 
         if (mapSize := os.getenv(KERIBaserMapSizeKey)) is not None:
-            self.MapSize = int(mapSize)
+            try:
+                self.MapSize = int(mapSize)
+            except ValueError:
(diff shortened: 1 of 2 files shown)


src/keri/db/basing.py
```diff
@@ -939,7 +939,10 @@ def __init__(self, headDirPath=None, reopen=False, **kwa):
         self._kevers.db = self  # assign db for read through cache of kevers
 
         if (mapSize := os.getenv(KERIBaserMapSizeKey)) is not None:
-            self.MapSize = int(mapSize)
+            try:
+                self.MapSize = int(mapSize)
+            except ValueError:
+                logger.exception("KERI_BASER_MAP_SIZE must be an integer value!  Will continue to use default keripy value of 104857600!")
```
pfeairheller, line 945: I do not believe this is an appropriate change.  It is not a bug to fail fast when an illegal value is entered for MapSize.  It prevents an application from running with a different value and failing some time later (perhaps much later) when a database map file blows up.  Trying to track that down would be very difficult.

If you want a better error message you can rethrow a new exception with a clearer message or just leave the code as is.

pfeairheller (changes requested):

src/keri/db/basing.py
daidoji: Sure.  I'll rethrow with a clearer message.

daidoji: @pfeairheller fixed

pfeairheller (approved):

PR #816, opened 2024-07-10
Add filters and reverse ordering in Noter and Notifier

rodolfomiranda:
This PR address a feature requested in KERIA (https://github.com/WebOfTrust/keria/issues/266) to allow notifications to be ordered in reverse, and be filtered by the `read` status and the `route` parameter in the `attrs` (`note.a.r`) if exists.

To get the notes in reversed order I added a new function in `LMDBer` called `getAllItemRvsdIter` that returns the reversed iterator.

### DIFF
--- a/src/keri/app/notifying.py
+++ b/src/keri/app/notifying.py
@@ -188,6 +188,20 @@ class DicterSuber(subing.Suber):
         for key, val in self.db.getAllItemIter(db=self.sdb, key=self._tokey(keys), split=False):
             yield self._tokeys(key), self.klas(raw=bytes(val))
 
+    def getItemRvsdIter(self, keys: Union[str, Iterable] = b""):
+        """ Return reversed iterator over the all the items in subdb
+
+        Parameters:
+            keys (tuple): of key strs to be combined in order to form key
+
+        Returns:
+            iterator: of tuples of keys tuple and val serdering.SerderKERI for
+            each entry in db
+
+        """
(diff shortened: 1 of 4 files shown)


src/keri/db/dbing.py
```diff
@@ -581,6 +581,37 @@ def getAllItemIter(self, db, key=b'', split=True, sep=b'.'):
                     splits = (bytes(key), val)
                 yield tuple(splits)
 
+    def getAllItemRvsdIter(self, db, key=b'', split=True, sep=b'.'):
```
pfeairheller, line 584: We already have a method that implements something similar to this and is called getIoValsAllPreBackIter.  This name should match.

src/keri/app/notifying.py
```diff
@@ -353,7 +367,49 @@ def getNotes(self, start=0, end=25):
                 break
 
         return notes
+    
+    def getFltNotes(self, start=0, end=25, rvsd=False, read=None, route=None):
```
pfeairheller, line 371: I don't think this is an appropriate approach for filtering messages in LMDB.  It requires deserializing the data and analyzing it and it hard codes the fields you can search for.  

If you look at the Seeker class in KERIA you'll see an approach to creating LMDB indexes that are more flexible and more performant than this approach.

pfeairheller (changes requested):

pfeairheller:
Can this PR be closed?

PR #817, opened 2024-07-15
New kli command for introductions and a few fixes

pfeairheller:
This PR includes:

- A new kli command to send an `/introduce` message with an AID and OOBI.
- Bug fix to watching to check against the correct key state.
- Logging additions, fixes and typo fixes.

### DIFF
--- a/src/keri/app/cli/commands/introduce.py
+++ b/src/keri/app/cli/commands/introduce.py
@@ -1,0 +1,137 @@
+# -*- encoding: utf-8 -*-
+"""
+keri.kli.commands module
+
+"""
+import argparse
+
+from hio.base import doing
+
+from keri.core import serdering
+from ..common import existing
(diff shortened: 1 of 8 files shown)


src/keri/app/cli/commands/introduce.py
```diff
@@ -0,0 +1,137 @@
...
+            raise ValueError(f"Unable to find OOBI with role {self.role} for {introducee}")
+
+        msg = hab.reply(route="/introduce", data=dict(
+            cid=hab.pre,
+            oobi=oobi
+        ))
+
+        serder = serdering.SerderKERI(raw=msg)
+        atc = msg[serder.size:]
+        self.postman.send(src=hab.pre, dest=recp, topic="credential", serder=serder,
```
kentbull, line 126: why would the topic be "credential" here?

PR #826, opened 2024-07-26
Now uses versioned Counter class in core.counting everywhere. Old unversioned coring.Counter has been removed.

SmithSamuelM:
This is the first major step in supporting verisoned CESR with versioned count codes.  
Next steps,

- Change parser and other places that parse count codes to dynamically determine version of incoming message stream and apply the correct version of Counter with count codes.
- Make default generation of count codes in output streams use the version 2 codes which requires that generation logic convert to always counting quadlets not groups of primitives.


SmithSamuelM:
@pfeairheller I didn't merge this yet because wanted you to be appriased of it.  I checked and it does not break anything in Keria, because Keria does not directly reference Counter or CtrDex.  I didn't know if something else might break.

SmithSamuelM:
The last commit breaks Keria so I fixed Keria as well

PR #827, opened 2024-07-27
Fix notes datetime ISO 8601 format

rodolfomiranda:
Notes that are not initiated with a `dt` parameter, are being created with a default `dt` that don't follows the ISO 8601 format. The class use the function `datetime.datetime.now().isoformat()` to create the string timestamp that result in a value without timezone or offset such as `2023-10-05T14:48:23.123456`.
Instead, `keripy` normal use `helping.nowIso8601()` to create the timestamp that result in values like `2021-06-27T21:26:21.233257+00:00`

### DIFF
--- a/src/keri/app/signaling.py
+++ b/src/keri/app/signaling.py
@@ -27,7 +27,7 @@ def signal(attrs, topic, ckey=None, dt=None):
         Notice:  Notice instance
 
     """
-    dt = dt if dt is not None else datetime.datetime.now().isoformat()
+    dt = dt if dt is not None else helping.nowIso8601()
 
     if hasattr(dt, "isoformat"):
         dt = dt.isoformat()
@@ -117,7 +117,7 @@ class Signaler(doing.DoDoer):
         Returns:
 
         """
-        dt = dt if dt is not None else datetime.datetime.now()
+        dt = dt if dt is not None else helping.nowIso8601()
         sig = signal(attrs=attrs, topic=topic, ckey=ckey, dt=dt)
 
         if sig.ckey is not None:
@@ -146,7 +146,7 @@ class Signaler(doing.DoDoer):
         _ = (yield self.tock)
 
         while True:  # loop checking for expired messages
-            now = datetime.datetime.now()
+            now = helping.nowUTC()
             toRemove = []
             for sig in self.signals:
                 if now - helping.fromIso8601(sig.dt) > self.SignalTimeout:  # Expire messages that are too old
--- a/tests/app/test_notifying.py
+++ b/tests/app/test_notifying.py
@@ -203,7 +203,7 @@ def test_notifier(mockHelpingNowUTC):
         assert notifier.rem(note.rid) is True
         assert notifier.getNotes() == []
 
-        dt = datetime.datetime.now()
+        dt = helping.nowIso8601()
         assert notifier.add(attrs=dict(a=1)) is True
         assert notifier.add(attrs=dict(a=2)) is True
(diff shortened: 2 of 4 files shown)


m00sey (approved):

tests/app/test_notifying.py
```diff
@@ -208,6 +210,8 @@ def test_notifier():
         notes = notifier.getNotes()
         assert len(notes) == 3
 
+        assert notes[2].datetime[-6:] == "+00:00"
```
m00sey, line 213: Instead of just truncating, using `mockHelpingNowUTC` and asserting the full timestamp.

m00sey:
Any interest in fixing the other usages of datetime vs helping.nowISO() @rodolfomiranda 😬

tests/app/test_notifying.py
rodolfomiranda: I didn't know about `mockHelpingNowUTC`. Code updated with the mock.

rodolfomiranda:
> Any interest in fixing the other usages of datetime vs helping.nowISO()

I could only find  `datetime.datetime` used in `signaling.py`. I updated it. The rest of `keripy` seems to be ok since `datetime` is only used for `timedelta`.

src/keri/app/signaling.py
```diff
@@ -146,7 +146,7 @@ def expireDo(self, tymth=None, tock=0.0):
         _ = (yield self.tock)
 
         while True:  # loop checking for expired messages
-            now = datetime.datetime.now()
+            now = helping.nowUTC()
```
m00sey, line 149: datetime.datetime.now(): Returns the current local date and time.
helping.nowUTC(): Returns the current date and time in UTC.

maybe this needs to be `helping.nowIso8601()` instead?

m00sey:
> > Any interest in fixing the other usages of datetime vs helping.nowISO()
> 
> I could only find `datetime.datetime` used in `signaling.py`. I updated it. The rest of `keripy` seems to be ok since `datetime` is only used for `timedelta`.

Thanks, appreciate you taking the extra time.

src/keri/app/signaling.py
rodolfomiranda: I tried `helping.nowIso8601()` first but `test_signaling.py` throws the error `TypeError: unsupported operand type(s) for -: 'str' and 'datetime.datetime'` on line 152 of `signaling.py`.
We need a datetime object as returned by `helping.nowUTC()`:
`Returns timezone aware datetime of current UTC time Convenience function that allows monkeypatching in tests to mock time`

m00sey: ahhh OK  - got it and this pairs with the `dt = dt if dt is not None else helping.nowIso8601()` update as they both now use UTC.

pfeairheller:
Can we please explain in the description of the PR why this change is being made?  The helping functions are used in many other places in KERIpy, specifically for time in KERI events and credentials.  Why are you proposing we change them for only this small subset of date usages?

If the date functions are wrong, why not fix them?

rodolfomiranda:
> If the date functions are wrong, why not fix them?
date functions are correct, but were not applied correctly in a couple of lines in `notifying.py` and `signaling.py`

pfeairheller (approved):
Wow, I definitely reviewed this PR too early in the morning... I read it backwards!!

Looks good to me!

SmithSamuelM:
Should be using helping.nowIso8601() or helping.toIso8601 everywhere there is a datetime string (non CESR encoded). Or Dater instance if want CESR encoded datetime

Issue #856, opened 2024-09-04
Stale read of smids - on multisig rotate neither GroupHab.smids nor HabitatRecord.smids are updated to latest signing members


kentbull:
### Version

1.1.18

### Environment

MacOS Sonoma 14.6.1, Python 3.12.3, local deployment

### Expected behavior

Calling `GroupHab.smids` should return the latest `smids` sent in to a multisig rotate command. 

### Actual behavior

Currently the `smids` coming back is exactly what was sent in during the inception event, or for a member being rotated in, the `smids` as of them being rotated into the group.

### Steps to reproduce

Perform a rotation on any multisig AID where you either rotate out one of the original group or rotate a new member in. 

If you rotate a member out then both the remaining members will still have the old `smids` values in their `HabitatRecord.smids` value.

If you rotate a member in then the original members will not have the new member's AID in their `HabitatRecord.smids` value, which is what `GroupHab.smids` is populated from during the `Habery.loadHabs()` call. The new member will have the correct `GroupHab.smids` value because it was provided the correct value in the rotation event that brought the new member in, yet on the next rotation if there are any members removed or added then that `smids` value will then be out of date.

kentbull:
This is fixed in
- main by https://github.com/WebOfTrust/keripy/pull/858
- v1.1.18 by https://github.com/WebOfTrust/keripy/pull/857

PR #857, opened 2024-09-05
fix: update smids and rmids on rotate

kentbull:
This updates the database HabitatRecord smids and rmids properties when a group rotation happens.

Fixes https://github.com/WebOfTrust/keripy/issues/856

### DIFF
--- a/src/keri/app/habbing.py
+++ b/src/keri/app/habbing.py
@@ -330,7 +330,7 @@ class Habery:
             if habord.mid and not habord.sid:
                 hab = GroupHab(ks=self.ks, db=self.db, cf=self.cf, mgr=self.mgr,
                                rtr=self.rtr, rvy=self.rvy, kvy=self.kvy, psr=self.psr,
-                               name=name, pre=pre, temp=self.temp, smids=habord.smids)
+                               name=name, pre=pre, temp=self.temp, smids=habord.smids, rmids=habord.rmids)
                 groups.append(habord)
             elif habord.sid and not habord.mid:
                 hab = SignifyHab(ks=self.ks, db=self.db, cf=self.cf, mgr=self.mgr,
@@ -2736,24 +2736,34 @@ class GroupHab(BaseHab):
 
         self.inited = True
 
-    def rotate(self, serder=None, **kwargs):
+    def rotate(self, serder=None, smids=None, rmids=None, **kwargs):
+
+        if (habord := self.db.habs.get(keys=(self.name,))) is None:
+            raise kering.ValidationError(f"Missing HabitatRecord for name={self.name}")
 
         if serder is None:
-            return super(GroupHab, self).rotate(**kwargs)
+            msg = super(GroupHab, self).rotate(**kwargs)
+        else:
 
-        # sign handles group hab with .mhab case
-        sigers = self.sign(ser=serder.raw, verfers=serder.verfers, rotated=True)
+            # sign handles group hab with .mhab case
+            sigers = self.sign(ser=serder.raw, verfers=serder.verfers, rotated=True)
 
-        # update own key event verifier state
-        msg = eventing.messagize(serder, sigers=sigers)
+            # update own key event verifier state
+            msg = eventing.messagize(serder, sigers=sigers)
 
-        try:
-            self.kvy.processEvent(serder=serder, sigers=sigers)
-        except MissingSignatureError:
-            pass
-        except Exception as ex:
-            raise kering.ValidationError("Improper Habitat rotation for "
-                                         "pre={self.pre}.") from ex
+            try:
+                self.kvy.processEvent(serder=serder, sigers=sigers)
+            except MissingSignatureError:
+                pass
+            except Exception as ex:
+                raise kering.ValidationError("Improper Habitat rotation for "
+                                             "pre={self.pre}.") from ex
+
+        self.smids = smids
+        self.rmids = rmids
+        habord.smids = smids
+        habord.rmids = rmids
+        self.db.habs.pin(keys=(self.name,), val=habord)
 
         return msg
 
@@ -2869,7 +2879,6 @@ class GroupHab(BaseHab):
         serder = eventing.query(query=query, **kwa)
 
         return self.mhab.endorse(serder, last=True)
-
 
     def witnesser(self):
         """This method name does not match logic???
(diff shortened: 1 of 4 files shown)


kentbull:
It would seem as though doing a name check on `BaseHab.save` will prevent any updates because the name argument will already be populated in the `self.db.names` database and thus the `self.db.names.get(keys=(ns, self.name))` will always return a value and raise the following value error:

```python
class BaseHab:
    ...
    def save(self, habord):
        self.db.habs.pin(keys=self.pre,
                         val=habord)
        ns = "" if self.ns is None else self.ns
        if self.db.names.get(keys=(ns, self.name)) is not None:
            raise ValueError("AID already exists with that name")

        self.db.names.pin(keys=(ns, self.name),
                          val=self.pre)
```

It seems like the solution is to have some sort of `BaseHab.update` that allows updating an existing `HabitatRecord` only if the record already exists. I will submit that.

src/keri/app/habbing.py
```diff
@@ -2632,7 +2632,7 @@ def __init__(self, smids, mhab=None, rmids=None, **kwa):
             pre (str | None): qb64 identifier prefix of own local controller else None
             mhab (Hab | None): group member hab (local) when this Hab is multisig group
                            else None. The mhab.pre aid could be a member of
-                           .smids, or .rmids, or both.
+                           .smids, or .rmids, or both.                                                                                                                      6-0606060606
```
pfeairheller, line 2635: What is this change?

pfeairheller (changes requested):
I think you can reorganize the rotate method without introducing the hidden method and calling save.  See my suggestion in Discord.

src/keri/app/habbing.py
kentbull: Looks like this was my little boy. I'll revert haha

SmithSamuelM:
I am confused by what is happening here:

the habbing.GroupHab.rotate clearly updates the rmids and smids when it is called.

```def rotate(self, smids=None, rmids=None, serder=None, **kwargs):

        if serder is None:
            return super(GroupHab, self).rotate(**kwargs)

        if (habord := self.db.habs.get(keys=(self.pre,))) is None:
            raise kering.ValidationError(f"Missing HabitatRecord for pre={self.pre}")

        # sign handles group hab with .mhab case
        sigers = self.sign(ser=serder.raw, verfers=serder.verfers, rotated=True)

        # update own key event verifier state
        msg = eventing.messagize(serder, sigers=sigers)

        try:
            self.kvy.processEvent(serder=serder, sigers=sigers)
        except MissingSignatureError:
            pass
        except Exception as ex:
            raise kering.ValidationError("Improper Habitat rotation for "
                                         "pre={self.pre}.") from ex

        self.smids = smids
        self.rmids = rmids
        habord.smids = smids
        habord.rmids = rmids
        self.db.habs.pin(keys=(self.pre,), val=habord)

        return msg
```

This called by JoinDoer.rotate which is called by joinDo  when the rotation completes sucessfully. So unless there is bug in how the logic for JoinDoer.joinDo and what leads up to that. Then there should be no fix required.  I am fearful that this is brute forcing a rotate but without completing the dependencies of verification that have to happen in order for it to be successful.  

Since I did not write this code, I am not familiar enough to have a definitive opinion, but I am concerned,   joinDo is waiting for all the group multisig events to occur, this seems to be the correct approach.  Otherwise one would be brute forcing a rotate that actually was not approved by the group, and therefore one would have to have a way to rollback. 

This is antithetical to the architecture of KERI in general. KERI is safe because it doesn't accept or change state until and unless the dependencies are met. It doesn't provisionally change state and then after the fact rollback. I understand that the latter is a common architecture for database based applications, but these are in general insecure, because of that. The former is the proper way to implement state changes that are secure. What this means is that a provisional state of waiting for depedencies to resolve and then possibly retrying interaction to refresh them when they time out or are unavailable is the correct approach. But the rmid, smid record is the state record not the provisional state record. 

If the application logic requires provisional state then that is a different database that does not exist yet.

SmithSamuelM:
A provisional source of truth is not the same as the primary source of truth.  Escrows are an example of a provisional source of truth. But escrows are ephemeral so that are meant to buffer the asynchronous nature of faulty unreliable network transports. A non-ephemeral but provisional source of truth usually resides in its own sandbox in the application. This allows some transaction to live persistently attempting to enact a change in the primary source of truth, but that change to the primary source of truth requires the cooperation of other actors in the distributed application space. Thus all the dependencies must be met prior to updating the primary source of truth to reflect a desired provisional source of truth.

Unfortunately many applications builders do not like this extra complication but then they must implement roll backs. The problem is that in distributed applications, its virtually impossible to rollback state that has propagated downstream. Event sourcing is one approach to a universal rollback mechanism, where you delete the database and recreate it by replaying all the events, albeit with the bad events elided.   While the KERI design has elements that feel like event sourcing, these do not exist at the application layer, and are not meant to protect from applications that brute force changes in key state or even application state such as the multi-sig group aid state (rmids and smids) Habord record.

kentbull:
I see what you are saying here. So the `smids` and `rmids` updates should go in a provisional storage area in order to abide by the rule of 
 > [KERI] doesn't accept or change state until and unless the dependencies are met
where in this case the completion of the multisig rotation is critical dependency that must be met. 

This way of waiting for dependencies makes a lot of sense.

Based on my reading of the JoinDoer it appears we need to change when `smids` and `rmids` are updated to occur after the call to `yield from self.startCounselor(...)` in order to wait until the multisig operation completes. If I understand correctly, currently the `smids` and `rmids` are sent into `ghab.rotate(...)` which causes them to be updated in the `HabitatRecord` before the completion of the multisig.

It will be straightforward to correct the code in JoinDoer. 

The original reason why I wrote the code like I did here was that I was mimicking something I saw in `GroupHab.rotate()`. I misspoke on the call yesterday when I said I changed the code to get a test to work. I was a bit distracted on the call.

It looks like the code I was mimicking I either did not understand fully or that code was not accounting for the need to wait until after the rotation is fully signed.

> If the application logic requires provisional state then that is a different database that does not exist yet.

The application code I am writing does not need provisional state. As long as the `smids` and `rmids` are only ever updated after the multisig rotation completes then that would meet my need.

I will update this PR with what I understand would be meeting the design goal of waiting to accept or change state until the dependencies are met.

pfeairheller (approved):

PR #867, opened 2024-09-23
Removed --alias from verify command.

daidoji:
Prefixes are 1-1 with aliases which are already required for this command and thus redundant.  Removed for ease of use.


pfeairheller:
The `--prefix` option is meant to specify the signer of the data, not the one verifying it.  The tests in `demo-script.sh` are misleading because they use the same AID for signing and verifying.

If anything, the `--alias` argument can be removed because you don't need a Hab context to verify someone else's signature, you just need the Habery database so you load their key state in the Kevers.

daidoji:
> The `--prefix` option is meant to specify the signer of the data, not the one verifying it. The tests in `demo-script.sh` are misleading because they use the same AID for signing and verifying.
> 
> If anything, the `--alias` argument can be removed because you don't need a Hab context to verify someone else's signature, you just need the Habery database so you load their key state in the Kevers.

@pfeairheller so how would I restructure this not to have to use an alias to verify?  Nothing in the verify command looks like its making a remote call to a prefix KEL in the wild so aren't aliases and prefixes 1-1 in the current implementation?

pfeairheller:
The appropriate change here is to remove the `--alias` argument because a local Hab is not needed to verifiy a signature.  The `--prefix` parameter is to allow the verification of signatures from other folks AIDs.  If we rely on `--alias` and a Hab, we can only verify our own signatures.

To remove `--alias` you need to change 

```
with existing.existingHab(name=name, alias=alias, base=base, bran=bran) as (_, hab):
```


to:

```
with existing.existingHby(name=name, base=base, bran=bran) as hby:
```

And then change `hab.kevers` to `hby.kevers`

pfeairheller (approved):

Issue #865, opened 2024-09-18
processCredential telquery messages missing src identifier (fix or remove?)


iFergal:
### Version

1.2.0.dev11 (still there on main)

### Expected behavior

If tel events are missing when processing a credential, the credential is escrowed and a telquery is initiated.

e.g. [here](https://github.com/WebOfTrust/keripy/blob/1.2.0-dev11/src/keri/vdr/verifying.py#L109)
```
self.cues.append(dict(kin="telquery", q=dict(ri=regk, i=vcid, issr=creder.issuer)))
```

Given the current code, expected would be the telquery would be sent to the issuer, so the credential can get out of escrow.

However, yesterday we had some discussions on a TEL observer role and in many ways retrieving TELs may vary based on use case and not directly involve the issuer agent, so I wonder if we should remove the telquery and have application specific code decide how to retrieve the TELs if missing?

### Actual behavior

Not only should this be `pre=creder.issuer` instead of `issr` but even then it's still missing the `src` field so the WitnessInquisitor doesn't know from which prefix to send the message - see query function [here](https://github.com/WebOfTrust/keripy/blob/1.2.0-dev11/src/keri/app/agenting.py#L475-L478).

### Steps to reproduce

I've added an endpoint in KERIA temporarily locally to call `processCredential` given an ACDC SAD. This happens if I don't do an explicit (properly formed) telquery beforehand.

An easier way might be to comment out the streaming of TEL events when disclosing an ACDC via IPEX: https://github.com/WebOfTrust/keripy/blob/1.2.0-dev11/src/keri/vdr/credentialing.py#L981-L990

PR #868, opened 2024-09-24
feat: tel state noticer to query for tel updates

iFergal:
Working towards TEL Observers - made some changes to make TSN behave same as KSN with a noticer that will fetch any remaining items.

This isn't _fully_ complete as I need some feedback - I'm really not sure if we are meant to escrow notices of new events that we haven't received yet.

Looking back in the history, we did this for KSN too but changed it in #441 and #538 so thinking it might make sense for TSN to behave the same way so that we can have efficient TEL observers. So I've made some changes there to be more like KSN.

If this isn't the right approach, we need to do something to trigger fetching of tels - the `telquery` calls for escrows aren't right atm for the same reason as #865 - no src so messages are never sent.

### DIFF
--- a/src/keri/vdr/eventing.py
+++ b/src/keri/vdr/eventing.py
@@ -1648,11 +1648,11 @@ class Tevery:
             if ri in self.tevers:
                 tever = self.tevers[ri]
                 tsn = tever.state()
-                self.cues.push(dict(kin="reply", route=f"/tsn/registry/{src}", data=asdict(tsn), dest=source.qb64))
+                self.cues.push(dict(kin="reply", route=f"/tsn/registry/{src}", src=src, data=asdict(tsn), dest=source.qb64))
 
                 if vcpre := qry["i"]:
                     tsn = tever.vcState(vci=vcpre)
-                    self.cues.push(dict(kin="reply", route=f"/tsn/credential/{src}", data=asdict(tsn), dest=source.qb64))
+                    self.cues.push(dict(kin="reply", route=f"/tsn/credential/{src}", src=src, data=asdict(tsn), dest=source.qb64))
 
         else:
             raise ValidationError("invalid query message {} for evt = {}".format(ilk, ked))
--- a/src/keri/vdr/eventing.py
+++ b/src/keri/vdr/eventing.py
@@ -1644,14 +1644,15 @@ class Tevery:
                 self.cues.append(dict(kin="replay", src=src, dest=source.qb64, msgs=msgs))
         elif route == "tsn":
             ri = qry["ri"]
+            src = qry["src"]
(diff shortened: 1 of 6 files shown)


src/keri/vdr/eventing.py
```diff
@@ -1644,14 +1644,15 @@ def processQuery(self, serder, source=None, sigers=None, cigars=None):
                 self.cues.append(dict(kin="replay", src=src, dest=source.qb64, msgs=msgs))
         elif route == "tsn":
             ri = qry["ri"]
+            src = qry["src"]
             if ri in self.tevers:
                 tever = self.tevers[ri]
                 tsn = tever.state()
-                self.cues.push(dict(kin="reply", route="/tsn/registry", data=asdict(tsn), dest=source))
+                self.cues.push(dict(kin="reply", route=f"/tsn/registry/{src}", data=asdict(tsn), dest=source.qb64))
```
iFergal, line 1651: Curious if we should just skip this if `vci` is set below, depends on if we need to escrow or not

src/keri/vdr/eventing.py
```diff
@@ -2007,12 +1994,8 @@ def processEscrows(self):
                                                 extype=kering.MissingRegistryError)
             self.reger.txnsb.processEscrowState(typ="credential-mae", processReply=self.processReplyCredentialTxnState,
                                                 extype=kering.MissingAnchorError)
-            self.reger.txnsb.processEscrowState(typ="credential-ooo", processReply=self.processReplyCredentialTxnState,
```
iFergal, line 2010: If this change is right, maybe we should also remove the `credential-mre` escrow, not sure.

src/keri/vdr/eventing.py
```diff
@@ -1784,24 +1785,17 @@ def processReplyRegistryTxnState(self, *, serder, saider, route, cigars=None, ts
...
-
-            raise kering.OutOfOrderTxnStateError("Out of order txn state={}.".format(rsr))
-
         tsaider = coring.Saider(qb64=rsr.d)
-        ldig = bytes(ldig)
-        # retrieve last event itself of signer given sdig
-        sraw = self.reger.getTvt(key=dgKey(pre=regk, dig=ldig))
-        # assumes db ensures that sraw must not be none because sdig was in KE
-        sserder = serdering.SerderKERI(raw=bytes(sraw))
+        if ldig is not None:  # escrow because event does not yet exist in database
```
iFergal, line 1789: aligning with ksn

iFergal:
@pfeairheller can you advise if this change to make tsn match ksn is on the right path? thanks!

pfeairheller:
This PR has aged out.  Please reopen when ready.

Issue #820, opened 2024-07-18
Semver comparison issues for migrations in same dev release set


iFergal:
### Version

1.2.0-dev1 - 1.2.0-dev10

### Environment

Mac OS, Python 3.12.2

### Expected behavior

I think we should be able to check migrations on a higher dev tag for the same version.

Not sure if we would plan on consolidating migrations prior to the next patch release instead, but I'd imagine that might cause headaches in workflows if stakeholders are involved in using test environments that are persisted.

### Actual behavior

This error is raised, despite `dev4` being older.
```
keri.kering.ConfigurationError: Database version=1.2.0-dev4 is ahead of library version=1.2.0-dev10
```

The reason for this is the Semver library will compare `dev4` and `dev10` as **strings** lexographically, and consider 4 bigger than 1 (ignoring the 0).
```
>>> semver.compare('1.2.0-dev4', '1.2.0-dev10')
1
>>> semver.compare('1.2.1', '1.2.0')
1
>>> semver.compare('1.2.0-4', '1.2.0-10')
-1
```
If the build number is just an integer, it works as intended as 4 < 10.

I quite like having `dev` in the version, so maybe could call `.replace('dev', '')` being semver comparing?

### Steps to reproduce

- Create an agent at a lower `1.2.0` dev tag
- Move to a higher `1.2.0` dev tag
- Call `kli migrate` on that agent

PR #872, opened 2024-10-08
fix: strip dev from version when semver comparing for migrations

iFergal:
This is to address #820 - creating the PR as I'm hitting this issue and now cannot upgrade versions without wiping the DBs.

Also, `rc` can be in the version too. I haven't covered that in this PR though because then `dev1` and `rc1` are the same version. Perhaps we can ignore it and just test rc versions in release environments? Open to suggestions!

### DIFF
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -1347,7 +1347,8 @@ class Baser(dbing.LMDBer):
         """
         for (version, migrations) in MIGRATIONS:
             # Check to see if this is for an older version
-            if self.version is not None and semver.compare(version, self.version) != 1:
+            if self.version is not None and semver.compare(version.replace('dev', ''),
+                                                           self.version.replace('dev', '')) != 1:
                 continue
 
             for migration in migrations:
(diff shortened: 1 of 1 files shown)


kentbull:
I addressed this in a different way my migration PRs [here](https://github.com/WebOfTrust/keripy/pull/875/files#diff-ad9a4110967a79613f9b5444dd8aa3092ca785ea9269e240ed72e247765e5932R1353).

Thanks for catching the second version check, I only had the first. My migrations PR is updated to include both.

iFergal:
Nice @kentbull that looks like a better approach!

I wonder if we should capture the `dev` or `rc` using it too. On the other hand, if there's a migration within the same patch version at the dev or rc level, maybe this means the minor or patch should be being bumped!

kentbull:
> if there's a migration within the same patch version at the dev or rc level, maybe this means the minor or patch should be being bumped!

That makes sense to me. Anything significant enough to require a database migration qualifies for a patch bump.

iFergal:
Cool - I prefer your approach to my quick hack, so I'm going to close this

PR #874, opened 2024-10-08
feat: add database migrations up to 1.2.0

kentbull:
Adds database migrations for:
- `0.6.7` (Jun 16, 2022) -> `0.6.8` (Nov 2, 2022): `hab_data_rename` (new)
- `0.6.8` (Nov 2, 2022) -> `1.0.0` (Mar 31, 2023): `add_key_and_reg_state_schemas` (new)
- `1.0.0` (Mar 31, 2023) -> `1.2.0` (Oct 4, 2024): `rekey_habs` (already existed)

I discovered the need for the `hab_data_rename` migration while testing migration for 0.6.7 test Root GAR databases similar to production as we believe v0.6.7 is what is in production for Root GARs.

The `add_key_and_reg_state_schemas` we initially thought was specific to Provenant yet in my testing it turns out it is needed to take a Root GAR 0.6.7 database up to 1.0.0, so not specific to Provenant.

I added the `name`, `base`, and `temp` args to `db.migrate(...)` so that the `viring.Reger` database could be initialized without having to open up the `Habery`.

## Important

The `rekey_habs` migration is not run even though it is included in this PR because it's version of `1.2.0` is above the KERI version of `1.1.19` stated in KERIpy. That migration is only needed for the `v1.2.x` series of releases. I include it in this PR only for completeness to remove any question of whether it is needed or not. It's not needed, though it is useful to see in this PR for reference.


kentbull:
This includes the semver cleanup idea from https://github.com/WebOfTrust/keripy/pull/872

PR #875, opened 2024-10-08
feat: migrations for 0.6.7 to 1.2.0

kentbull:
Add database migrations for:
- `0.6.7` (Jun 16, 2022) -> `0.6.8` (Nov 2, 2022): `hab_data_rename` (new)
- `0.6.8` (Nov 2, 2022) -> `1.0.0` (Mar 31, 2023): `add_key_and_reg_state_schemas` (new)
- `1.0.0` (Mar 31, 2023) -> `1.2.0` (Oct 4, 2024): `rekey_habs` (already existed)

I discovered the need for the `hab_data_rename` migration while testing migration for 0.6.7 test Root GAR databases similar to production as we believe v0.6.7 is what is in production for Root GARs.

The `add_key_and_reg_state_schemas` we initially thought was specific to Provenant yet in my testing it turns out it is needed to take a Root GAR 0.6.7 database up to 1.0.0, so not specific to Provenant.

I added the `name`, `base`, and `temp` args to `db.migrate(...)` so that the `viring.Reger` database could be initialized without having to open up the `Habery`.


kentbull:
This includes the semver cleanup idea from https://github.com/WebOfTrust/keripy/pull/872

src/keri/db/basing.py
```diff
@@ -1335,7 +1337,7 @@ def reload(self):
         for keys in removes:  # remove bare .habs records
             self.habs.rem(keys=keys)
 
-    def migrate(self):
+    def migrate(self, name, base, temp):
```
pfeairheller, line 1340: These three parameters (name, base, temp) are available from the database object itself.  Please remove them and use them from `self`

pfeairheller (changes requested):
Just the one change to remove the unneeded parameters and we should be good to merge.

src/keri/db/basing.py
kentbull: Those three arguments are now removed.

pfeairheller (approved):

PR #877, opened 2024-10-11
Version bump to 1.1.20

kentbull:
version bump


pfeairheller:
We don't do these with PRs and also you missed the version number in the Makefile

PR #880, opened 2024-10-15
docs: fix Sphinx build

kentbull:
Upgrade Sphinx to 8.1.3 and myst-parser to 4.1.3.

1. doing `make clean html` builds the docs.
2. Open the docs on OSx with `open _build/html/index.html`
3. Profit

 There are a bunch of warnings yet they don't stop the docs from building.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ except ImportError:
     sphinx_rtd_theme = None
 
 project = 'keri'
-copyright = '2022, Dr. Samuel Smith and contributors'
+copyright = '2024, Dr. Samuel Smith and contributors'
 author = 'Dr. Samuel Smith'
 
 version = release = keri.__version__
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -28,7 +28,7 @@ except ImportError:
     sphinx_rtd_theme = None
 
 project = 'keri'
-copyright = '2024, Dr. Samuel Smith and contributors'
+copyright = '2022 - 2024, Dr. Samuel Smith and contributors'
 author = 'Dr. Samuel Smith'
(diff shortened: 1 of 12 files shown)


kentbull:
took about 15 mins of work. What else, if anything, is missing from this? the docs build into the static assets `_build` dir and can be published. It seems complete to me.

m00sey (commented):
seems to be missing the vdr package

m00sey:
Appreciate it only took 15 minutes.

The local rendering for me does not have any detail, screenshot attached.

I assume this is just a templating issue, but it is what I referred to on the call.

Screenshot 2024-10-15 at 4 31 07 PM

kentbull:
I'm also seeing the VDR API:

image

And the details of the VDR API:

image

This is when I do an `open ./_build/html/index.html` from the OSx command line after a `make clean html`

m00sey:
LGTM - however I dislike the theme lol

docs/conf.py
m00sey, line 1: Think the copyright needs to be 2022 - 2024 inclusive of previous years, not exclude them?

docs/index.rst
```diff
@@ -35,11 +35,13 @@ API Reference
     keri_app
     keri_core
     keri_db
+    keri_demo
```
m00sey, line 38: I think we should remove this

kentbull: This is removed.

docs/conf.py
kentbull: Should it be copyright or should it be creative commons or Apache 2 or something?

kentbull: for now I changed it to 2022 - 2024

PR #882, opened 2024-10-17
feat: de-couple contact creation from OOBI resolution

iFergal:
This will resolve WebOfTrust/keria#291 and also avoid the lack of clarity raised in WebOfTrust/keria#292. As dicussed on [Discord](https://discord.com/channels/1148629222647148624/1154416772401860648/1295408495801401396).

kli command added to add the alias using `org.replace`. I added it to the demo scripts that seemed to need the contact to be in place. Many don't require it, but I hope I didn't miss too many.


m00sey:
Removed GLEIF so this PR can make progress - we don't have time to address it now.

We have not upgraded to 1.2 yet so we can address our own concerns when we do.

iFergal:
Cool thanks! @pfeairheller let me know if the kli command is OK on your side.

I think this change should be OK for the KERIA side @2byrds @lenkan

pfeairheller (approved):

PR #894, opened 2024-11-22
Clear both KEL and TEL escrows; list all events in escrow, or their counts

kentbull:
This backports `kli escrow clear` and `kli escrow list` into the 1.1.x release branch.
I also added clearing for every escrow I could find as there seemed to be some missing from the 1.2.x branch. I upgraded the `kli escrow list` command to show a count of all existing escrows.

### DIFF
--- a/src/keri/app/cli/commands/escrow/clear.py
+++ b/src/keri/app/cli/commands/escrow/clear.py
@@ -1,0 +1,57 @@
+# -*- encoding: utf-8 -*-
+"""
+KERI
+keri.kli.commands.escrow module
+
+"""
+import argparse
+
+from hio import help
+from hio.base import doing
+from keri.app.cli.common import existing
+from keri.vdr import viring
+
+logger = help.ogler.getLogger()
+
+parser = argparse.ArgumentParser(description='Clear escrows')
+parser.set_defaults(handler=lambda args: handler(args),
+                    transferable=True)
+parser.add_argument('--name', '-n', help='keystore name and file location of KERI keystore', required=True)
+parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
+                    required=False, default="")
+parser.add_argument('--passcode', '-p', help='21 character encryption passcode for keystore (is not saved)',
+                    dest="bran", default=None)  # passcode => bran
+parser.add_argument('--force', '-f', action="store_true", required=False,
+                    help='True means perform clear without prompting the user')
+
+
+def handler(args):
+    if not args.force:
+        print()
+        print("This command will clear all escrows and is not reversible.")
+        print()
+        yn = input("Are you sure you want to continue? [y|N]: ")
+
+        if yn not in ("y", "Y"):
+            print("...exiting")
+            return []
+
+    kwa = dict(args=args)
+    return [doing.doify(clear, **kwa)]
+
+
+def clear(tymth, tock=0.0, **opts):
+    """ Command line clear handler
+    """
+    _ = (yield tock)
+    args = opts["args"]
+    name = args.name
+    base = args.base
+    bran = args.bran
+    logger.setLevel("INFO")
+
+    with existing.existingHby(name=name, base=base, bran=bran) as hby:
+        hby.db.clearEscrows()
+        reger = viring.Reger(name=hby.name, db=hby.db, temp=False)
+        reger.clearEscrows()
+
--- a/src/keri/app/cli/commands/escrow/list.py
+++ b/src/keri/app/cli/commands/escrow/list.py
@@ -54,16 +54,10 @@ def escrows(tymth, tock=0.0, **opts):
             # KEL / Baser escrows
 
             if (not escrow) or escrow == "unverified-receipts":
-                count = 0
-                for key, _ in hby.db.getUreItemIter():
-                    count += 1
-                escrows["unverified-receipts"] = count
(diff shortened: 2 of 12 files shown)


src/keri/app/cli/commands/escrow/clear.py
```diff
@@ -0,0 +1,57 @@
...
+
+logger = help.ogler.getLogger()
+
+parser = argparse.ArgumentParser(description='Clear escrows')
+parser.set_defaults(handler=lambda args: handler(args),
+                    transferable=True)
+parser.add_argument('--name', '-n', help='keystore name and file location of KERI keystore', required=True)
+parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
+                    required=False, default="")
+parser.add_argument('--passcode', '-p', help='21 character encryption passcode for keystore (is not saved)',
```
daidoji, line 22: Do you need `--nopasscode` here too in order to match the other commands?

src/keri/app/cli/commands/escrow/list.py
```diff
@@ -0,0 +1,279 @@
...
+
+logger = help.ogler.getLogger()
+
+parser = argparse.ArgumentParser(description='Views events in escrow state.')
+parser.set_defaults(handler=lambda args: handler(args),
+                    transferable=True)
+parser.add_argument('--name', '-n', help='keystore name and file location of KERI keystore', required=True)
+parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',
+                    required=False, default="")
+parser.add_argument('--passcode', '-p', help='21 character encryption passcode for keystore (is not saved)',
```
daidoji, line 27: Likewise `--nopasscode` as above.

src/keri/app/cli/commands/escrow/list.py
```diff
@@ -0,0 +1,279 @@
...
+    escrow = args.escrow
+
+    try:
+        with existing.existingHby(name=name, base=base, bran=bran) as hby:
+            escrows = dict()
+
+            # KEL / Baser escrows
+
+            if (not escrow) or escrow == "unverified-receipts":
+                count = 0
```
daidoji, line 57: Also good is `sum(1 for k, v in <iter>)` in python if you want to save some lines.

src/keri/app/cli/commands/escrow/list.py
```diff
@@ -0,0 +1,279 @@
...
+            if (not escrow) or escrow == 'tel-anchorless':
+                count = 0
+                for key, _ in reger.getAllItemIter(reger.taes):
+                    count += 1
+                escrows["tel-anchorless"] = count
+
+            if (not escrow) or escrow == "missing-registry-escrow":
+                creds = list()
+                for (said,), dater in reger.mre.getItemIter():
+                    creder, *_ = reger.cloneCred(said)
```
daidoji, line 230: Why `*_` here?

src/keri/db/dbing.py
```diff
@@ -137,17 +137,20 @@ def splitKey(key, sep=b'.'):
...
+    top, on = splitKey(key, sep=sep)
     on = int(on, 16)
-    return (pre, on)
+    return (top, on)
+
+splitSnKey = splitKeyON # alias so intent is clear, sn vs fn; backport of 1.2.x alias
+splitFnKey = splitKeyON # alias so intent is clear, sn vs fn; backport of 1.2.x alias
 
 splitKeySN = splitKeyON  # alias so intent is clear, sn vs fn
 splitKeyFN = splitKeyON  # alias so intent is clear, sn vs fn
```
daidoji, line 156: Why are these seemingly repeated?  Also why are they global?

daidoji (approved):
Mostly just reviewing for curiosity's sake but looks good to me.  Maybe a few comments of suggestions but nothing crucial.

src/keri/app/cli/commands/escrow/clear.py
kentbull: As far as I can tell then `--nopasscode` is only needed in the `kli init` command because that is where the Habery, keeping.Manager are set up based on whether a passcode is supplied or not. If `--nopasscode` is set when `kli init` is called then you don't need to provide a passcode on later commands, so it is okay that `kli escrow clear` and `kli escrow list` do not have this argument.

src/keri/app/cli/commands/escrow/list.py
kentbull: Same response as above.

src/keri/app/cli/commands/escrow/clear.py
daidoji: Oh cool.

src/keri/app/cli/commands/escrow/list.py
kentbull: Yes, I do want to save some lines. I was thinking there was a reduction function I could use like this, so thanks. This has been addressed in the new code

src/keri/app/cli/commands/escrow/list.py
kentbull: Likely brevity as there are four return arguments in the tuple and only the first was needed.. Someone else wrote this code.

src/keri/db/dbing.py
kentbull: It is repeated because I ported the two names from the 1.2.x release version (now `main`). Those two are global to the `keri.db.dbing` module because that's how Sam wrote them.

daidoji (approved):

m00sey (approved):

PR #902, opened 2024-12-10
pass through salts for vc create to facilitate command line multisig

m00sey:


src/keri/vc/proving.py
```diff
@@ -4,52 +4,51 @@
...
-               data,
-               recipient=None,
-               private=False,
-               salt=None,
-               status=None,
-               source=None,
-               rules=None,
-               version=Version,
-               kind=Kinds.json):
+def credential(schema:str,
```
pfeairheller, line 21: Thank you for fixing the type hints in the PR too, I'm surprised I didn't have to remind you here as well...  nicely done.

pfeairheller (approved):

kentbull:
Looks like the `store_true` in the two new command line arguments `--private-credential-nonce` and `--private-subject-nonce` here should have had their `"action"` prop set to `store` so that they can accept string values rather than be set only to True or False based on presence or absence of the command line arguments like `--private`.

I'll fix this in a PR.

PR #905, opened 2024-12-13
adds rp to exn to work with >1.2.0-dev4

m00sey:


src/keri/core/serdering.py
```diff
@@ -304,7 +304,7 @@ class Serder:
                     Ilks.bar: Fieldage(saids={Saids.d: DigDex.Blake3_256},
                         alls=dict(v='', t='',d='', i='', dt='', r='',a=[])),
                     Ilks.exn: Fieldage(saids={Saids.d: DigDex.Blake3_256},
-                        alls=dict(v='', t='', d='', i="", p="", dt='', r='', q={},
+                        alls=dict(v='', t='', d='', i="", rp="", p="", dt='', r='', q={},
```
daidoji, line 307: What is the RP field?

daidoji (commented):
What is "newer KERI"?

src/keri/core/serdering.py
m00sey: "newer KERI" this is pre 1.2.0

Rp field was added 1.2.0-dev4

pfeairheller (approved):

src/keri/core/serdering.py
daidoji: > The PR is against 1.1 - renamed to be clear.
> 
> Rp field was added 1.2.0-dev4

Sure, but what is it?

m00sey: Top level recipient field. This was discussed some time ago, hence its addition to 1.2.+
It was added to save unpacking the `a.r` field.

daidoji: > Top level recipient field. This was discussed some time ago, hence its addition to 1.2.+ It was added to save unpacking the `a.r` field.

Do you remember when?  It hasn't been added to the specs yet.

m00sey: The specs show Ver_2_0 but it looks like it’s `ri` in there, these are against Ver1__0 and Ver_1_0.

SmithSamuelM:
The fix should be that an `ri` field for receiver identifier (AID) be added to the EXN and XIP messages in version 2.  This is what is in the specification for KERI.  The reason for `ri` and not `rp` is that the convention when the field value is an AID is for the field label to either be `i` or if there is already and `i` to use a two character field label where the second character is `i` and the first character is the modifier. For example `di` is the delegator identifier field.

`rp` was put in at some point in the current code before we had the discussion of using the convention. 
If you are going to update version 1 of keri, then I suggest replacing  the `rp` with the `ri` field label. Then it would already be version 2 compliant.

PR #906, opened 2024-12-13
V1.2.1

m00sey:


src/keri/vc/proving.py
```diff
@@ -25,8 +22,9 @@ def credential(schema,
                issuer,
                data,
                recipient=None,
-               private=False,
-               salt=None,
+               private:bool=False,
+               private_credential_nonce:Optional[str]=None,
+               private_subject_nonce:Optional[str]=None,
```
pfeairheller, line 27: If you are going to introduce type hints to parameters in a method (which is a good thing), please update the entire signature of the method, not just the few you are touching.

pfeairheller (changes requested):
One change requested, and it is minor.

src/keri/vc/proving.py
m00sey: my bad!

pfeairheller (approved):

Issue #886, opened 2024-11-05
Way to embed an endpoint for .well-known discovery in an OOBI


iFergal:
### Feature request description/rationale

In controlled deployments of KERI for particular use cases, it can be easier to bootstrap agents with schemas. (e.g. all the signify tests use the vLEI test server for vLEI related schemas)

I would like a generic way for a mobile wallet (in my case) to discover schemas from new contacts. Since schemas are cryptographically bound to their SAID, OOBIs work well and don't have the overhead of signing etc. @pfeairheller has suggested using .well-known at an endpoint.

From this point, I still need to discover the endpoint hosting these schemas for a particular issuer, which may be different to the controller/agent URL, especially if this is KERIA and not a web service built on top of keripy.

Discovery of this service endpoint via the OOBI makes sense to me, but then this service endpoint wouldn't really have a KERI role since there isn't a KERI agent behind that URL. It's just for discovery. Not sure if I'm going down the wrong path here. It feels like an anti pattern for a "service endpoint" to not have a KERI identifier at the domain.

Alternatively (just thought of this now), KERIA could automatically serve these so the domain would be the same. Would need to be controlled more tightly, but that's a general problem for KERIA to figure out.

pfeairheller:
As long as the service endpoint is attested to in a signed rpy message using BADA-RUN, I don't believe there is a requirement that they endpoint itself only contain information directly related to the AID.  Exposing a "schema" endpoint is a perfectly valid use case.  

From the KERI spec:

> An important use case for BADA-RUN is to process OOBIs that provide service endpoint discovery of the AIDS of KERI components. These components include but are not limited to, Controllers, Agents, Backers (Witness or Registrar), Watchers, Jurors, Judges, and Forwarders.

Even though KERIpy uses an enumeration (sort of) for role names, the spec is very clear that we are not limited to those roles.  To me, this falls under the KAPI (KERI API) domain and should be a simple API defined to start with a service endpoint with a given role (I really like `registrar` but that is overloaded in the spec already) and a defined set of routes off of that service endpoint for discovering schema.  It _could_ be off of .well-known, but I don't think it has to be (and I'm not sure it makes sense when it would be just as easy to define a few routes for schema discovery).

That would define the rules for schema discovery.  Have any issuer expose a service endpoint with a role of `schema-discovery-thingie` then any verifier knows what API to hit against the URL of that service endpoint to get a list of exposed schema and how to resolve each schema exposed.

[...]

PR #920, opened 2025-01-06
feat: new role for schema discovery

iFergal:
To be honest, I'm not sure what to call this new role.

It's to address #886. This role is meant to allow a controller to designate a service endpoint (which may be on a different URL than the controller/agent) for discovery of schemas. Perhaps it could be used for other things, which is why I tried to leave it broad as `indexer` - like an index of OOBIs (or other things, like public ACDCs...) by SAIDs.

Open to suggestions.

---

Tested with KERIA/Signify to add a new loc scheme and end role auth where the `cid=eid`, and the service endpoint appears in the OOBI format `http://keria.com/oobi/<cid>`.

### DIFF
--- a/src/keri/kering.py
+++ b/src/keri/kering.py
@@ -342,9 +342,9 @@ SEPARATOR_BYTES = SEPARATOR.encode("utf-8")
 Schemage = namedtuple("Schemage", 'tcp http https')
 Schemes = Schemage(tcp='tcp', http='http', https='https')
 
-Rolage = namedtuple("Rolage", 'controller witness registrar watcher judge juror peer mailbox agent')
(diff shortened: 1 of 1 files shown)


daidoji:
@iFergal the `platonic` role https://en.wikipedia.org/wiki/Platonism or if that word is overloaded in modern english, the `Parminedian` role after Parmenides. :-)

2byrds:
> @iFergal the `platonic` role https://en.wikipedia.org/wiki/Platonism or if that word is overloaded in modern english, the `Parminedian` role after Parmenides. :-)

please no 😆

daidoji:
`zhenxing` role https://en.wikipedia.org/wiki/True_form_(Taoism)

pfeairheller (approved):
We can change the name later if we are inspired by something else.

PR #923, opened 2025-01-12
Fix OOBI contact generation

### LINKED ISSUE #922: Incorrect fix in PR #882
The fix to completely remove contact creation in OOBI resolution  in PR #882 was incorrect.  The previous code that was changed allowed for OOBI resolution that does not effect the contact by not providing an alias during resolution.  Upstream code should be changed to use OOBI resolution correctly instead of removing behavior from KERIpy.

pfeairheller:
This PR fixes #922 by reintroducing contact creation since it is optional.  It also includes a few minor logging fixes.

### DIFF
--- a/scripts/demo/basic/challenge.sh
+++ b/scripts/demo/basic/challenge.sh
@@ -14,12 +14,6 @@ cha2_oobi="$(kli oobi generate --name cha2 --alias cha2 --role witness | sed -n 
 kli oobi resolve --name cha1 --oobi-alias cha2 --oobi "${cha2_oobi}"
 kli oobi resolve --name cha2 --oobi-alias cha1 --oobi "${cha1_oobi}"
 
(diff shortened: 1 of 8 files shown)


kentbull:
From one of our calls we discussed changing:
```python
self.org.replace(pre=obr.cid, data=dict(alias=obr.oobialias, oobi=url))
```

to
```python
self.org.update(pre=obr.cid, data=dict(alias=obr.oobialias, oobi=url))
```

Because `Organizer.update()` performs a `Organizer.replace()` if the contact data doesn't exist yet if it does exist then it performs a merge, which is what Fergal was after anyway. What was causing people problems is that `Organizer.replace() wipes out any contact data that is there, which will always erase any aliases you have for someone if you re-oobi with them without specifying the alias.

I'll open up A PR for this.

PR #926, opened 2025-01-17
adds escrow clear

m00sey:
We have it in main and 1.1.30, just sharing the love


kentbull (approved):
sure

yah, sure, looks good

PR #931, opened 2025-01-28
uwes Refactor

arilieb:
Refactor of the sub db of unverified event indexed escrowed couples from witness signers (uwes) to use B64OnIoDupSuber instead of bespoke methods.


pfeairheller (approved):
Nice first PR, Ari!

PR #935, opened 2025-01-30
feat: TSP message types and optionally encrypted StreamPoster

iFergal:
Allows the `StreamPoster` to wrap the message stream in ESSR. Default is not to use it. I thought about doing this in KERIA but I think it's quite useful here. Tested e2e with full IPEX grant/admit in KERIA/Signify and works.

For now doesn't work with forwarded messages to mailboxes or witnesses but can be extended, at least once this has been reviewed. (I also want to avoid the list copy on line 360)

This will prepend the stream with the proposed TSP message type `SCS` from Sam: https://github.com/WebOfTrust/keripy/discussions/612#discussioncomment-11986131

I created a `Tsper` class similar to `Ilker` for the 3 new TSP message types.

### DIFF
--- a/src/keri/app/forwarding.py
+++ b/src/keri/app/forwarding.py
@@ -6,6 +6,7 @@ keri.app.forwarding module
 module for enveloping and forwarding KERI message
 """
 import random
+import pysodium
 from ordered_set import OrderedSet as oset
 
 from hio.base import doing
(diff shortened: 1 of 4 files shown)


pfeairheller (commented):
My only concern is that this approach does not consider very large message bodies with chunking.  Perhaps that doesn't matter for a first pass.

We will discuss in the ToIP call on March 11th.

iFergal:
@pfeairheller If we're happy with the general approach, I can try to add chunking, and extend it to support forwarding! Whether in this PR or another. I won't make today's call as I'm OOO and traveling.

pfeairheller:
We discussed this PR in both the community dev call and the KERIpy maintainer's call.  The decision was made to move this functionality into another repo, not KERIpy.

iFergal:
@pfeairheller @SmithSamuelM Was there a decision on which repo it could be moved to? thanks!

PR #939, opened 2025-02-27
adds a naive count all to SuberBase

m00sey:
During a recent issue we had a old "un-parsable" event stuck in escrow.

The current (1.1.31 and 1.2.4) code includes a "count".
The count was only used in a log message, but utilized "getItemIter" which attempts to serialize the message into a serder. Which then throws an exception because it does not have enough required fields.

The `cntAll` is naive as it doesn't account for dupes etc - but it servers the purpose for a count at the most basic level.


kentbull:
LGTM

pfeairheller (approved):

PR #957, opened 2025-03-31
fix some incorrect help docs for kli vc

lenkan:
- Fixes some incorrect command descriptions (probably from old copy-paste)
- Fixes some incorrect argument descriptions
- Remove `transferable=True` default args, also probably from old copy-paste
- Moved `--base` to follow `--name` in the arguments declaration to be consistent with many of the other commands in the kli.


kentbull:
Looked at this during the dev call, looks okay.

m00sey (approved):

PR #960, opened 2025-04-14
Update Doer method signatures to support new Hio multiprocessing safety

kentbull:
ported from Sam's commit to main in PR #945
These changes are required in order to upgrade to HIO 0.6.18 or higher. Without these, or similar changes, then some KLI commands break.


m00sey:
ah this got close because I deleted the 1.2.5 branch

m00sey:
pleas reopen against 1.2.6

Issue #973, opened 2025-04-22
Log messages with bare bytes() data structures need an encoding or will fail when throwing an error


kentbull:
### Version

1.1.32, 1.2.6, main

### Environment

All

### Expected behavior

Log messages like the one made below should not throw encoding errors:
```python
class Kevery:
    ...
    def processEscrowPartialDels(self):
        ...
                dgkey = dgKey(epre, edig)
                if not (esr := self.db.esrs.get(keys=dgkey)):  # get event source, otherwise error
                    # no local source so raise ValidationError which unescrows below
                    msg = f"PDE Missing escrowed event source at dig = {bytes(edig)}"
                    logger.info("Kevery unescrow error: %s", msg)
                    raise ValidationError(msg)
```

### Actual behavior

Many of my recent log messages introduced with the logging overhauls have a bare `bytes(...)` data structure in them and as such are causing the wrong errors to be thrown during escrow object validation, and likely other places.

All places such log messages are created in 1.1.32, 1.2.6, and main need to be fixed.

For example, the following error needs to be resolved:
```python
2025-04-22 21:22:48 [sally] ERROR    Kevery PDE unescrowed: string argument without an encoding
Traceback (most recent call last):
  File "/keripy/src/keri/core/eventing.py", line 5941, in processEscrowPartialDels
    msg = f"PDE Stale event escrow at dig = {bytes(edig)}"
                                             ^^^^^^^^^^^
TypeError: string argument without an encoding
```

By specifying an encoding like below:
```python
    msg = f"PDE Stale event escrow at dig = {bytes(edig).encode('utf-8')}"
```

### Steps to reproduce

Send a credential presentation to Sally that does not have a delegation fully complete and this error will be triggered.

kentbull:
Resolved by https://github.com/WebOfTrust/keripy/pull/976

PR #976, opened 2025-04-23

kentbull:
Using "{bytes(edig)}" in f-strings was causing the incorrect log messages to trigger because `bytes(edig)` failed when `edig` was a string and not bytes. An error `TypeError: string argument without an encoding` would be thrown as shown below. The solution this PR suggests is to just use the raw `edig` in the f-strings. This works because f-strings auto-convert a data type to its string representation.

```python
2025-04-22 21:22:48 [sally] ERROR    Kevery PDE unescrowed: string argument without an encoding
Traceback (most recent call last):
  File "/keripy/src/keri/core/eventing.py", line 5941, in processEscrowPartialDels
    msg = f"PDE Stale event escrow at dig = {bytes(edig)}"
                                             ^^^^^^^^^^^
TypeError: string argument without an encoding
```

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -5595,7 +5595,7 @@ class Kevery:
                 dgkey = dgKey(pre, bytes(edig))
                 if not (esr := self.db.esrs.get(keys=dgkey)):  # get event source, otherwise error
                     # no local source so raise ValidationError which unescrows below
-                    msg = f"PSE Missing escrowed event source at dig = {bytes(edig)}"
+                    msg = f"PSE Missing escrowed event source at dig = {edig}"
                     logger.info("Kevery unescrow error: %s", msg)
                     raise ValidationError(msg)
 
@@ -5603,7 +5603,7 @@ class Kevery:
                 dtb = self.db.getDts(dgkey)
                 if dtb is None:  # othewise is a datetime as bytes
                     # no date time so raise ValidationError which unescrows below
-                    msg = f"PSE Missing escrowed event datetime at dig = {bytes(edig)}"
+                    msg = f"PSE Missing escrowed event datetime at dig = {edig}"
                     logger.trace("Kevery unescrow error: %s", msg)
                     raise ValidationError(msg)
 
@@ -5612,7 +5612,7 @@ class Kevery:
                 dte = helping.fromIso8601(bytes(dtb))
                 if (dtnow - dte) > datetime.timedelta(seconds=self.TimeoutPSE):
                     # escrow stale so raise ValidationError which unescrows below
-                    msg = f"PSE Stale event escrow at dig = {bytes(edig)}"
+                    msg = f"PSE Stale event escrow at dig = {edig}"
                     logger.trace("Kevery unescrow error: %s", msg)
                     raise ValidationError(msg)
 
@@ -5620,7 +5620,7 @@ class Kevery:
                 eraw = self.db.getEvt(dgkey)
                 if eraw is None:
                     # no event so so raise ValidationError which unescrows below
-                    msg = f"PSE Missing escrowed evt at dig = {bytes(edig)}"
+                    msg = f"PSE Missing escrowed evt at dig = {edig}"
                     logger.trace("Kevery unescrow error: %s", msg)
                     raise ValidationError(msg)
 
@@ -5629,7 +5629,7 @@ class Kevery:
                 sigs = self.db.getSigs(dgkey)
                 if not sigs:  # otherwise its a list of sigs
                     # no sigs so raise ValidationError which unescrows below
-                    msg = f"PSE Missing escrowed evt sigs at dig = {bytes(edig)}"
+                    msg = f"PSE Missing escrowed evt sigs at dig = {edig}"
                     logger.trace("Kevery unescrow error: %s", msg)
                     raise ValidationError(msg)
                 wigs = self.db.getWigs(dgKey(pre, bytes(edig)))  # list of wigs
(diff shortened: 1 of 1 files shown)


pfeairheller (changes requested):
The effects of the previous change were not evident until some folks were running in their environment meaning no unit tests exist that trigger this.  Please submit a unit test with these changes that prove they work as expected.

src/keri/core/eventing.py
```diff
@@ -5956,7 +5956,7 @@ def processEscrowPartialDels(self):
                 sigs = self.db.getSigs(dgkey)  # list of sigs
                 if not sigs:  # empty list
                     # no sigs so raise ValidationError which unescrows below
-                    msg = f"PDE Missing escrowed evt sigs at dig = {bytes(edig)}"
+                    msg = f"PDE Missing escrowed evt sigs at dig = {edig}"
                     logger.info("Kevery unescrow error: %s", bytes(edig))
```
m00sey, line 5960: You either want `msg` or `edig` here but `bytes(edig)` is wrong.

kentbull:
Do we have any example tests that trigger escrows I could start from? All the prior change did was change the way we logged things, meaning there were no unit tests prior to my prior change either.

m00sey (changes requested):

src/keri/core/eventing.py
kentbull: yes, the `bytes(edig)` is a holdover from the old logging statements. For example, from KERIpy 1.1.25 the logging statements Sam wrote look like this:
```python
logger.info("Kevery unescrow error: Missing event sigs at."
                                    "dig = %s\n", bytes(edig))
```

kentbull: f-strings are capable to convert the `edig` to the `__repr__` format for the given datatype, whatever it is.

kentbull: The reason line 5960 does not cause an error is because the log substitution with the `%s` operator auto-converts to the `__repr__` of whatever data type is there, thus a string encoding is not needed. All the old logging statements used `%s` and log substitutions, not f-string interpolations.

The problem I am trying to fix is when the call to `bytes(edig)` fails due to the lack of an encoding. It just so happens that this is occurring inside of an f-string.

kentbull: okay, I am wrong, you are right. The `bytes(edig)` call would cause an error if it were a string.

SmithSamuelM:
The LMDB function return Python memoryview objects, which are not usable in f"" strings are in general with most operations like logging. If it is a memoryview than converting it to bytes is accomplished by bytes(mv). But formated f"" want unicode strings not bytes so to to convert to regular strings you want. bytes(mv).decode()

some of the functions do the work of converting to str by checking the type and then decoding if it needed it. 

But when trying to somehow optimize the logging or simplify the logic, taking shortcuts becomes counter productive when one does not actually understand the types one is encountering.  So the logging worked before. Now its broken. The main problem was it was too verbose but it was correct. So instead of just fixing the verbosity, there was an attempt to somehow make it more 'efficient' without doing the work of understanding it.    Better to just do the one thing and not the other.

kentbull:
Sam, thank you for clarifying that the return value for edig in statements like
```python
for ekey, edig in self.db.getOoeItemIter(key=key):
```
can sometimes be a memoryview. When I dug into the `LMDBer.gettopItemIter` function to discover the type I saw the `ckey = bytes(ckey)` in the below code and assumed all returns of `edig` would be cast to bytes like the one in the below function is.

```python
class LMDBer(filing.Filer):
        ...
        def getTopItemIter(self, db, top=b''):
        ...
        with self.env.begin(db=db, write=False, buffers=True) as txn:
            cursor = txn.cursor()
            if cursor.set_range(top):  # move to val at key >= key if any
                for ckey, cval in cursor.iternext():  # get key, val at cursor
                    ckey = bytes(ckey)        # <----------- referring to this line ---------------
                    if not ckey.startswith(top): #  prev entry if any last in branch
                        break  # done
                    yield (ckey, cval)  # another entry in branch startswith key
            return  # done raises StopIteration
```

I made the assumption that `bytes(edig)` in the logger statement was extraneous since it appeared to always be done in the LMDBer function. It sounds like you are saying that in some instances the `edig` return value is a memoryview. I will go and manually check each iterator function to confirm this just to double check.

My goal wasn't to make it more efficient, I just wanted to get past the "string argument without encoding" TypeError which was confusing. It confused me because I thought that `edig` would always be a `bytes` object due to the cast I saw in LMDBer. I didn't take the time at first to look at every LMDBer function that is used in `eventing.py` in these escrow loops, though now I will.

From what I can tell the problem I am trying to solve occurs when the `bytes(edig)` function is called with something Python thinks is a string.

kentbull:
After reading the code just now I found out the following, which will be reflected in the PRs shortly, regarding the value for `edig` in the various escrow functions.

- `processEscrowOutOfOrders` -> bytes
- `processEscrowPartialSigs` -> bytes
- `processEscrowPartialWigs` -> bytes
- `processEscrowPartialDels` -> string
- `processEscrowDelegables` -> bytes, yet comes from DB as deserialized string, then cast to bytes
- `processQueryNotFound` -> string
- `processEscrowDuplicitous` -> bytes

I will update these escrow functions accordingly. This is the kind of deeper code analysis I should have done before making a bulk change like I did when I wrote this PR. In the future I will be more diligent in validating assumptions I make about data types in the code.

src/keri/core/eventing.py
kentbull: My latest commit addresses the underlying issue by me going and doing the work I should have done the first time around, reading all of the Baser or LMDBer functions to see their specific return types and then ensuring the type used in the log statement is correct.

kentbull:
The latest commit fixes the type used in log messages by properly converting memory view types to `bytes` and then using `.decode()` to transform it into a string. For consistency I used a `dig` variable to get it right once and then reuse it in all relevant log statements in each `processEscrow*` function.

kentbull:
@pfeairheller I added unit tests proving that the log statements work in the OOO, PSE, PWE, and PDEL escrows. I will have the rest tested later this week. I will add the same tests to 1.1.32 and main as well.

kentbull:
Resolves https://github.com/WebOfTrust/keripy/issues/973

PR #977, opened 2025-04-23

kentbull:
Using "{bytes(edig)}" in f-strings was causing the incorrect log messages to trigger because `bytes(edig)` failed when `edig` was a string and not bytes. An error `TypeError: string argument without an encoding` would be thrown as shown below. The solution this PR suggests is to just use the raw `edig` in the f-strings. This works because f-strings auto-convert a data type to its string representation.

```python
2025-04-22 21:22:48 [sally] ERROR    Kevery PDE unescrowed: string argument without an encoding
Traceback (most recent call last):
  File "/keripy/src/keri/core/eventing.py", line 5941, in processEscrowPartialDels
    msg = f"PDE Stale event escrow at dig = {bytes(edig)}"
                                             ^^^^^^^^^^^
TypeError: string argument without an encoding
```


kentbull:
The latest commit fixes the type used in log messages by properly converting memory view types to `bytes` and then using `.decode()` to transform it into a string. For consistency I used a `dig` variable to get it right once and then reuse it in all relevant log statements in each `processEscrow*` function.

kentbull:
Remove `dig` variables and replace with inline `bytes(edig).decode()` calls.

Make sure strings passed to `bytes()` cast have encodings specified.

PR #978, opened 2025-04-23

kentbull:
Edited:
When I made some changes to the logging I incorrectly assumed that `edig` was always a memoryview in the `processEscrow*` functions and thus able to be converted to bytes with `bytes(edig)`. This was incorrect because sometimes the `edig` argument is a string, so running `bytes(edig)` without specifying an encoding threw an error, thus creating the "string argument without an encoding" error.

This PR fixes the logging in each `processEscrow*` function so that each type, memoryview, bytes, or string, is appropriately converted to a printable value before being interpolated in a log message.


kentbull:
Remove `dig` variables and replace with inline `bytes(edig).decode()` calls.

Make sure strings passed to `bytes()` cast have encodings specified.

PR #986, opened 2025-04-29
Updated Temp File for Windows, Tests and CI/CD 

arilieb:
### This PR is dependent on https://github.com/ioflo/hio/pull/80. Windows CI/CD is expected to fail until https://github.com/ioflo/hio/pull/80 has been merged.
- Updated temp file access such that it is os agnostic
- Removed \tmp file creation in windows CI/CD

PR #991, opened 2025-05-02
feat: configure keystore parameters with environment variables

### LINKED ISSUE #955: Feature request: ability to specify keystore parameters as environment variables
### Feature request description/rationale

I would like to be able to call kli commands that require keystore parameters such as `--name`, `--base` and possibly `--passcode`  without having to supply them on the command line.

Example use:

```
export KERIPY_KEYSTORE_NAME="my_keystore"
export KERIPY_KEYSTORE_BASE="/home/user/.keri-dir/my_keystore"

# Now I can call kli incept without having to supply --name and --base parameters because they are stored in my environment.
kli incept --alias test --transferable --icount 1 --isith 1 --ncount 1 --nsith 1

# I can do more commands here without having to specify the keystore parameters
kli contacts list
```

The environment variables can be stored in files that I can source or load whichever way I want. If I run kli in a docker container, I would set the environment variables as part of the docker runtime.

# Implementation

Implementation would be rather straightforward I think. We can just add default variables to the arguments.

```
parser.add_argument('--name', '-n', help='keystore name and file location of KERI keystore', required=True, default=os.getenv("KERIPY_KEYSTORE_NAME"))
parser.add_argument('--base', '-b', help='additional optional prefix to file location of KERI keystore',                     required=False, default=os.getenv("KERIPY_KEYSTORE_BASE") or "")
```

If you accept this feature request, I can open up a PR. Let me know your opinion. Also, whether or not you think the passcode can be added to the environment as well? If you think it is unsafe, we can add call it `KERIPY_UNSAFE_PASSCODE` or something, or skip it.

lenkan:
This PR enables you to configure which keystore to access using environment variables.

Summary of changes:

- Extracted --name, --base and --passcode into a common "Parsery.keystore()" method. Not sure about the module or naming here, tried to follow https://github.com/WebOfTrust/keripy/blob/main/ref/naming.md, but please advise. Should be easy search replace to modify.
- Removed redundant "transferable=True" default option which seemed to have been left over from copy-paste. Please advise if not correct and I can change back.
- Inlined the "addInceptingArgs" as it was only used in one place. I thought we could define one pattern for common arguments and use that everywhere.

Closes #955


lenkan:
I believe the failed windows tests are fixed by #986

lenkan:
Note that if you are hesitant about the environment variables being enables, we can disable it. It is trivial to add now when the keystore parameters are extracted to its own module. We can also hide it being a feature flag.

src/keri/app/cli/commands/witness/start.py
```diff
@@ -12,10 +12,11 @@
 from keri import help
 from keri.app import directing, indirecting, habbing, keeping, configing
 from keri.app.cli.common import existing
+from keri.app.cli.common.parsing import Parsery
 
 d = "Runs KERI witness controller.\n"
 d += "Example:\nwitness -H 5631 -t 5632\n"
-parser = argparse.ArgumentParser(description=d)
+parser = argparse.ArgumentParser(description=d, parents=[Parsery.keystore()], conflict_handler='resolve')
```
pfeairheller, line 19: Very interesting, I didn't know about the `conflict_handler` parameter.

Issue #981, opened 2025-04-28
Multisig delegator to multisig delegate does not work


lenkan:
### Version

1.2.5-dev0 (main branch)

### Environment

Ubuntu 24.04 WSL, Python 3.12

### Expected behavior

The test script that does delegation with multisig delegator and delegate should work.Ukk

### Actual behavior

Script waits indefinitely on the delegation to succeed. Here is the log:

```
Creating delegate's first local identifier in delegate1 keystore
KERI Keystore created at: /home/lenkan/.keri/ks/delegate1
KERI Database created at: /home/lenkan/.keri/db/delegate1
KERI Credential Store created at: /home/lenkan/.keri/reg/delegate1

Loading 3 OOBIs...
http://127.0.0.1:5642/oobi/BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha/controller succeeded
http://127.0.0.1:5643/oobi/BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM/controller succeeded
http://127.0.0.1:5644/oobi/BIKKuvBwpmDVA4Ds-EpL5bt9OqPzWPja2LigFYZN2YfX/controller succeeded
Waiting for witness receipts...
Prefix  EJ97lUuRH3xz0OMKhdMAU6V2TcSF9X6m1CKyIbIUcRxp
        Public key 1:  DFnhkUJRim8dlcXVHHIki3ObI37qwaZgKFyiHY2VOJsJ

Creating delegate's second local identifier in delegate2 keystore
KERI Keystore created at: /home/lenkan/.keri/ks/delegate2
KERI Database created at: /home/lenkan/.keri/db/delegate2
KERI Credential Store created at: /home/lenkan/.keri/reg/delegate2

Loading 3 OOBIs...
http://127.0.0.1:5642/oobi/BBilc4-L3tFUnfM_wJr4S4OJanAv_VmF_dJNN6vkf2Ha/controller succeeded
http://127.0.0.1:5643/oobi/BLskRTInXnMxWaGqcpSyMgo0nYbalW99cGZESrz3zapM/controller succeeded
http://127.0.0.1:5644/oobi/BIKKuvBwpmDVA4Ds-EpL5bt9OqPzWPja2LigFYZN2YfX/controller succeeded
Waiting for witness receipts...
Prefix  ELZyCjnSL2Haors35LKM19T4qWT4K8Gfz1FPDD9oJN33
[...]

PR #997, opened 2025-05-06
changed locallyDelegated method to include locally membered group AID…

SmithSamuelM:
… for multisig delegator

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -1700,7 +1700,7 @@ class Kever:
         of self. Unaccepted dip events do not have self.delpre set yet.
 
         Returns:
-            (bool): True if pre is local hab but not group hab
+            (bool): True if pre is local hab or group hab that has a local member
                         When pre="" empty or None then returns False
 
         Parameters:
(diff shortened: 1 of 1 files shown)


SmithSamuelM:
This should fix #981 on latest main branch by fixing underlying locallyDelegated method in Kever.
@lenkan @m00sey @kentbull

src/keri/core/eventing.py
```diff
@@ -1695,14 +1700,16 @@ def locallyDelegated(self, pre: str):
...
+            (bool): True if pre is local hab or group hab that has a local member
                         When pre="" empty or None then returns False
 
         Parameters:
             pre (str): qb64 identifier prefix if any.
         """
+
         pre = pre if pre is not None else ""
-        return self.locallyOwned(pre=pre)
+        return pre in self.prefixes or pre in self.groups
```
lenkan, line 1711: I thought `self.prefixes` also includes `self.groups`, so the `or` clause is redundant?

SmithSamuelM: Yes your correct, my bad.

lenkan:
I have rebased my PR that has the e2e test script that triggers this. I think this needs test coverage to ensure we don't get further regressions down the line. It is probably better with unit tests, but at least the e2e is something.

kentbull:
I will be writing that test. I must ensure this works and continues to work since GLEIF has a critical dependency on this functionality.

PR #999, opened 2025-05-07
Implement script for multisig delegation from v1.1 to current version

lenkan:
Perhaps this type of test is better suited in GLEIF-IT organization. Opening as Draft for discussion


pfeairheller:
@m00sey will look at where this can be moved and comment back here.

PR #1009, opened 2025-05-14
Fixes to how parses incomplete CESR streams

SmithSamuelM:


SmithSamuelM:
@m00sey @iFergal Cherry pick these back to whatever branch you are using and then in the witness doer processing change the call to Parser.parsator  to include framed==True

lenkan (commented):
Does this fix usage of parseOne in ReceiptEnd? If so, that is great!

I feel there is a lack of test coverage here if no tests had to change or be added for this change. Perhaps we should open an issue for that?

SmithSamuelM:
@lenkan I don't know about other use case. And yes we need test coverage for malformed CESR streams so please open an issue.

lenkan:
> @lenkan I don't know about other use case.

Sorry, I was referring to the usage here in ReceiptEnd: https://github.com/WebOfTrust/keripy/blob/6b402bf16383bb6d3a56d799f8b6ff8c2fd34293/src/keri/app/indirecting.py#L1082

If a KERI server registers that endpoint handler and receives an incomplete attachment in the cesr attachment header, the process is blocked by the parser, causing the server to become unresponsive. At least prior to this change. Currently, `kli witness start` uses that endpoint.

PR #1012, opened 2025-05-19
feat: configure LMDB db sizes in config file

kentbull:
Adds ability to configure LMDB sizes using the bootstrap configuration file.
Also adds some INFO level logging on the config file location and database sizes set.

Adds the `.mapSize` property to LMDBer and the `LMDBer.ConfigKey` that is overridden in each subclass. This config key is used to select a subset of the overall config file to use to configure each LMDBer subclass.

For backwards compatibility the `KERI_BASER_MAP_SIZE` environment variable still works, yet may be deprecated in the future. Moving forward using the configuration file as the single source of truth for configuration is the recommended path.

### DIFF
--- a/scripts/keri/cf/main/wan.json
+++ b/scripts/keri/cf/main/wan.json
@@ -1,5 +1,25 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
+  "baser": {
+    // LMDB Map size - 1 GB
+    "mapSize": 1073741824
+  },
+  "keeper": {
+    // LMDB Map Size - 25 MB
+    "mapSize": 26214400
+  },
+  "mailboxer": {
+    // LMDB Map Size - 16.384 GB
+    "mapSize": 17179869184
+  },
+  "reger": {
+    // LMDB Map Size - 512 MB
+    "mapSize": 536870912
+  },
+  "noter": {
+    // LMDB Map Size - 128 MB
+    "mapSize": 134217728
+  },
   "wan": {
     "dt": "2022-01-20T12:57:59.823350+00:00",
     "curls": ["tcp://127.0.0.1:5632/", "http://127.0.0.1:5642/"]
--- a/scripts/keri/cf/main/wan.json
+++ b/scripts/keri/cf/main/wan.json
@@ -1,23 +1,18 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
   "baser": {
-    // LMDB Map size - 1 GB
     "mapSize": 1073741824
   },
   "keeper": {
-    // LMDB Map Size - 25 MB
     "mapSize": 26214400
   },
   "mailboxer": {
-    // LMDB Map Size - 16.384 GB
     "mapSize": 17179869184
   },
   "reger": {
-    // LMDB Map Size - 512 MB
     "mapSize": 536870912
   },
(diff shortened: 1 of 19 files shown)


kentbull:
There is both a global `lmdber` config block and the subclass specific blocks shown below illustrate how to configure map sizes either globally or per class.

```hjson
{
  "dt": "2022-01-20T12:57:59.823350+00:00",
  "lmdber": {
    // LMDB Map Size - 50 MB
    "mapSize": 52428800
  },
  "baser": {
    // LMDB Map size - 1 GB
    "mapSize": 1073741824
  },
  "keeper": {
    // LMDB Map Size - 25 MB
    "mapSize": 26214400
  },
  "mailboxer": {
    // LMDB Map Size - 16.384 GB
    "mapSize": 17179869184
  },
  "reger": {
    // LMDB Map Size - 512 MB
    "mapSize": 536870912
  },
  "noter": {
    // LMDB Map Size - 128 MB
    "mapSize": 134217728
  },
  "wan": {
    "dt": "2022-01-20T12:57:59.823350+00:00",
    "curls": ["tcp://127.0.0.1:5632/", "http://127.0.0.1:5642/"]
  },
  "iurls": [
  ]
}
```

scripts/keri/cf/main/wan.json
```diff
@@ -1,5 +1,25 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
+  "baser": {
+    // LMDB Map size - 1 GB
```
pfeairheller, line 4: Please remove these lines from all JSON files.

pfeairheller (changes requested):
You need to remove the comments from the JSON files. The comments make them invalid.

scripts/keri/cf/main/wan.json
kentbull: these lines have been removed.

src/keri/app/keeping.py
```diff
@@ -259,6 +262,7 @@ def reopen(self, **kwa):
         Open sub databases
         """
         opened = super(Keeper, self).reopen(**kwa)
+        logger.info("[%s] Keeper map size set to %s", self.name, self.mapSize)
```
pfeairheller, line 265: These log statement should be at debug, like the one in Baser.

src/keri/db/dbing.py
```diff
@@ -408,15 +412,25 @@ def reopen(self, readonly=False, **kwa):
...
 
+        cf = cf if cf is not None else self.cf
+        config = cf.get() if cf is not None else None
+        lmdberConf = config.get(LMDBer.ConfigKey) if config is not None else None
+        dbconf = config.get(self.ConfigKey, None) if config is not None else None
+        if dbconf is not None:
+            self.mapSize = dbconf.get("mapSize", 104_857_600)  # 10MB default
+        elif config and lmdberConf: # fall back to LMDBer size if present
+            self.mapSize = lmdberConf.get("mapSize", 104_857_600)  # 10MB default
+
```
pfeairheller, line 430: We discussed in an implementers call that the dependency here is not correct.  Lets discuss at the next call.

pfeairheller (changes requested):

kentbull (commented):
I have addressed the requested changes.

PR #1013, opened 2025-05-19
feat: configure LMDB db sizes in config file

kentbull:
Adds ability to configure LMDB sizes using the bootstrap configuration file.
Also adds some INFO level logging on the config file location and database sizes set.

Adds the `.mapSize` property to LMDBer and the `LMDBer.ConfigKey` that is overridden in each subclass. This config key is used to select a subset of the overall config file to use to configure each LMDBer subclass.

For backwards compatibility the `KERI_BASER_MAP_SIZE` environment variable still works, yet may be deprecated in the future. Moving forward using the configuration file as the single source of truth for configuration is the recommended path.

### DIFF
--- a/scripts/keri/cf/main/wan.json
+++ b/scripts/keri/cf/main/wan.json
@@ -1,5 +1,25 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
+  "baser": {
+    // LMDB Map size - 1 GB
+    "mapSize": 1073741824
+  },
+  "keeper": {
+    // LMDB Map Size - 25 MB
+    "mapSize": 26214400
+  },
+  "mailboxer": {
+    // LMDB Map Size - 16.384 GB
+    "mapSize": 17179869184
+  },
+  "reger": {
+    // LMDB Map Size - 512 MB
+    "mapSize": 536870912
+  },
+  "noter": {
+    // LMDB Map Size - 128 MB
+    "mapSize": 134217728
+  },
   "wan": {
     "dt": "2022-01-20T12:57:59.823350+00:00",
     "curls": ["tcp://127.0.0.1:5632/", "http://127.0.0.1:5642/"]
--- a/scripts/keri/cf/main/wan.json
+++ b/scripts/keri/cf/main/wan.json
@@ -1,23 +1,18 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
   "baser": {
-    // LMDB Map size - 1 GB
     "mapSize": 1073741824
   },
   "keeper": {
-    // LMDB Map Size - 25 MB
     "mapSize": 26214400
   },
   "mailboxer": {
-    // LMDB Map Size - 16.384 GB
     "mapSize": 17179869184
   },
   "reger": {
-    // LMDB Map Size - 512 MB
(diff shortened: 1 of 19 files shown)


kentbull:
There is both a global `lmdber` config block and the subclass specific blocks shown below illustrate how to configure map sizes either globally or per class.

```hjson
{
  "dt": "2022-01-20T12:57:59.823350+00:00",
  "lmdber": {
    // LMDB Map Size - 50 MB
    "mapSize": 52428800
  },
  "baser": {
    // LMDB Map size - 1 GB
    "mapSize": 1073741824
  },
  "keeper": {
    // LMDB Map Size - 25 MB
    "mapSize": 26214400
  },
  "mailboxer": {
    // LMDB Map Size - 16.384 GB
    "mapSize": 17179869184
  },
  "reger": {
    // LMDB Map Size - 512 MB
    "mapSize": 536870912
  },
  "noter": {
    // LMDB Map Size - 128 MB
    "mapSize": 134217728
  },
  "wan": {
    "dt": "2022-01-20T12:57:59.823350+00:00",
    "curls": ["tcp://127.0.0.1:5632/", "http://127.0.0.1:5642/"]
  },
  "iurls": [
  ]
}
```

src/keri/db/dbing.py
```diff
@@ -378,18 +379,20 @@ def __init__(self, readonly=False, **kwa):
...
                                 False means open database in read/write mode
-
+            cf (Configer): optional Configer to configure the opened LMDB database
         """
 
         self.env = None
         self._version = None
         self.readonly = True if readonly else False
+        self._mapSize = 104_857_600  # 10MB fallback default
+        self.cf = kwa["cf"] if "cf" in kwa else None
```
kentbull, line 389: pass it in as a regular key word argument to `LMDBer.__init__`

daidoji: not important just noticed as I was glancing at this, you can use a ternary evaluation here, but I think `kwa.get("cf")` is a little nicer and a few less key strokes to get the same result.

kentbull: Good point if I were leaving it in `kwa`, yet the  output of the maintainer call was to pass cf as an argument to both `LMDBer.__init__` and `.reopen()` so it will be a proper key word argument rather than something passed into `kwa`. The rationale articulated is that `kwa` should not be used to pass init args to objects, rather only be used to pass function args around, so `kwa` is not appropriate in this case and a proper key word arg is best here.

kentbull: This has been done.

scripts/keri/cf/main/wan.json
```diff
@@ -1,5 +1,25 @@
 {
   "dt": "2022-01-20T12:57:59.823350+00:00",
+  "baser": {
+    // LMDB Map size - 1 GB
+    "mapSize": 1073741824
```
pfeairheller, line 5: Please remove these lines from all JSON, they are not valid json.

pfeairheller (changes requested):
You need to remove the comments from the JSON files.  The comments make them invalid.

kentbull:
All comments have been addressed and this is now ready for re-review.

PR #1019, opened 2025-06-03
Mapper class nested  field maps in CESR Native messages. Implemented query and reply with tests.

SmithSamuelM:


src/keri/core/mapping.py
```diff
@@ -4,8 +4,343 @@
...
+        elif isinstance(val, (str, bytes, bytearray)):
+            try:
+                primitive = Matter(qb64=val)
+            except Exception as ex:  # not valid primitive
+                ser.extend(Labeler(label=val).qb64b)  # so serialize as label
+            else:  # valid primitive in qb64 format
+                if len(primitive.qb64) != len(val):  # not complete so not really valid
+                    ser.extend(Labeler(label=val).qb64b)  # so serialize as label
+                else:  # really valid complete primitive in qb64
+                    ser.extend(primitive.qb64b)
```
lenkan, line 329: Is this behaviour defined in spec?

So, if I have a field that is a string value, but by coincidence, it also parses as a valid primitive in the text domain, it gets serialized as that primitive, rather than an arbitrary string.

See this test case:

```python
def test_mapper_roundtrip_with_primitive():
    mapper = Mapper(mad=dict(a="1AAM"))
    assert mapper.qb64 == '-IAC0J_a1AAM'

    mapper = Mapper(qb64="-IAC0J_a1AAM")
    assert mapper.mad == {"a": "1AAM"}
```

Which fails with


```
    def test_mapper_with_primitive():
        mapper = Mapper(mad=dict(a="1AAM"))
        assert mapper.qb64 == '-IAC0J_a1AAM'

        mapper = Mapper(qb64="-IAC0J_a1AAM")
>       assert mapper.mad == {"a": "1AAM"}
E       AssertionError: assert {'a': True} == {'a': '1AAM'}
E
E         Differing items:
E         {'a': True} != {'a': '1AAM'}
E         Use -v to get more diff
```

PR #1068, opened 2025-07-29
feat: TSP message types and optionally encrypted StreamPoster

iFergal:
Updated version of #935 

Supports chunking and mailboxes. Also keeps TSP related functions in a separate subpackage.


src/keri/spac/essr/tsping.py
iFergal, line 1: From what I remember from the meeting Sam suggested `spac -> essr` subpackages, happy to revise as needed. Not sure on how I'm meant to define the interface for the sub-package though - I think in python we need IDE/tooling to enforce that as you can access anything private technically? I might be wrong.

src/keri/app/forwarding.py
```diff
@@ -269,14 +271,32 @@ def deliver(self):
...
+            # bext field can be randomized to reduce correlation based on packet size, empty for now
+            msg.extend(coring.Bexter(bext="").qb64b)
+
         while self.evts:
             evt = self.evts.popleft()
 
             serder = evt["serder"]
             atc = evt["attachment"] if "attachment" in evt else b''
 
+            if self.essr and len(msg) + len(serder.raw) + len(atc) > 16384:
```
iFergal, line 296: 16kb chunk size if this makes sense. I chunked it based on the next message going over the limit, instead of strict chunking would not work with the parser

src/keri/spac/essr/tsping.py
iFergal: From a discussion with Sam yesterday, as this is closed mode - I've removed TSP references to avoid confusion as this is more of an implementation of SPAC.

pfeairheller (approved):

PR #1070, opened 2025-07-29
KRAM Module and Point Implementation for QRY messages

arilieb:
First pass at KRAM implementation
- KRAM module (kraming.py)
- KRAM tests (test_kraming.py)
- Point solution KRAM implementation for query messages
>- Since witness aids are non-transferrable, they will not need to collect signers, and as such an escrow is unnecessary

### DIFF
--- a/src/keri/core/kraming.py
+++ b/src/keri/core/kraming.py
@@ -1,4 +1,3 @@
-
 from keri import kering
 from keri.help import helping
 
@@ -6,12 +5,10 @@ from typing import NamedTuple
 from keri import help
 from hio.base import Doer
 import json
-from hio.help import decking
-from keri.core.serdering import Serder
-from keri.kering import Ilks
 
 logger = help.ogler.getLogger()
 
+# TODO: Implement message type as part of TimelinessCache module for more granular control
 class MessageType(NamedTuple):
     """Named tuple for KERI message types, not yet used."""
     QRY: str = "qry"
@@ -31,14 +28,14 @@ class TimelinessCache:
     1. A Lagging Window Size Table - to determine validity windows for different message types
     2. A Replay Cache Table - to store timestamps of previously validated messages
     """
-    #change to float time, 3 seconds, 1 second
-    def __init__(self, db, defaultWindowSize=300_000_000, defaultDriftSkew=60_000_000, cues=None, msgs=None):
+
+    def __init__(self, db, defaultWindowSize=3.0, defaultDriftSkew=1.0):
         """Initialize the TimelinessCache.
 
         Parameters:
             db: Database instance that contains the IoSetSuber at db.time
-            defaultWindowSize (int): Default window size in microseconds
-            defaultDriftSkew (int): Default drift skew in microseconds
+            defaultWindowSize (float): Default window size in seconds
+            defaultDriftSkew (float): Default drift skew in seconds
         """
         self.defaultWindowSize = defaultWindowSize
         self.defaultDriftSkew = defaultDriftSkew
@@ -48,6 +45,7 @@ class TimelinessCache:
     def setWindowParameters(self, aid, windowSize=None, driftSkew=None, messageType=None):
         """Set window parameters for given autonomic identifier and message type.
         """
+        # TODO: Implement message type as part of the window parameters
         windowSize = windowSize or self.defaultWindowSize
         driftSkew = driftSkew or self.defaultDriftSkew
 
@@ -68,8 +66,9 @@ class TimelinessCache:
                 be used to determine the window size for a given message type
 
         Returns:
-            tuple: (windowSize, driftSkew) as integers in microseconds
+            tuple: (windowSize, driftSkew) as floats in seconds
         """
+        # TODO: Implement message type as part of the window parameters
         try:
             # Try to get the stored parameters
             storedData = self.db.kram.getLast(aid)
@@ -93,11 +92,10 @@ class TimelinessCache:
         Returns:
             str: The key for the Replay Cache Table
         """
+        # TODO: Implement message type as part of the key (serder.ilk)
         sad = serder.sad
-        sourceAid = sad.get("i", "")  # 'i' for identifier in KERI messages
-        # messageType = serder.ilk  # Use serder.ilk for message type
+        sourceAid = sad.get("i", "")
 
-        # return (sourceAid, messageType)
         return sourceAid
 
     def _getCachedTimestamp(self, key):
@@ -107,12 +105,12 @@ class TimelinessCache:
             key (str): The cache key
 
         Returns:
-            int | None: The cached timestamp in microseconds or None if not found
+            float | None: The cached timestamp in seconds or None if not found
         """
         try:
             storedTimestamp = self.db.time.getLast(key)
             if storedTimestamp is not None:
-                return int(storedTimestamp)
+                return float(storedTimestamp)
         except Exception as e:
             print(e)
             pass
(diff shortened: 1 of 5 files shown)


tests/core/test_kraming.py
```diff
@@ -0,0 +1,76 @@
...
+
+        tc = kraming.TimelinessCache(db=hby.db, defaultDriftSkew=1.0, defaultWindowSize=1.0)
+
+        assert tc.defaultWindowSize == 1.0
+        assert tc.defaultDriftSkew == 1.0
+
+        # Set window parameters for the test AIDs
+        tc.setWindowParameters(hab.pre, windowSize=1.0, driftSkew=1.0)
+
+        def create_test_serder(ilk, timestamp=None, route=None, sourceAid=hab.pre, qBlock=None, routeParams=None):
```
kentbull, line 25: Should this `create_test_serder` fn be lifted out into conftest or some other fixtures location?

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Parameters:
+            aid (str): autonomic identifier
+            messageType (str | None): message type identifier. None for now, but will
+                be used to determine the window size for a given message type
+
+        Returns:
+            tuple: (windowSize, driftSkew) as floats in seconds
+        """
+        # TODO: Implement message type as part of the window parameters
```
kentbull, line 71: You could accomplish this with a key of `(aid, messageType)`

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+    def setWindowParameters(self, aid, windowSize=None, driftSkew=None, messageType=None):
+        """Set window parameters for given autonomic identifier and message type.
+        """
+        # TODO: Implement message type as part of the window parameters
+        windowSize = windowSize or self.defaultWindowSize
+        driftSkew = driftSkew or self.defaultDriftSkew
+
+        # Serialize the tuple as JSON bytes for storage
+        windowTuple = (windowSize, driftSkew)
+        serialized = json.dumps(windowTuple).encode('utf-8')
```
kentbull, line 54: Should we be serializing as JSON or as a set of CESR encoded floats?

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+
+        self.db.kram.pin(aid, [serialized])
```
kentbull, line 56: If you `.pin()` with the keys of `(aid, messageType)` for a key that would look like "aid.messageType" then you could support window parameters by message type.

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+
+        Parameters:
+            serder: The SerderKERI instance containing the message
+
+        Returns:
+            str: The key for the Replay Cache Table
+        """
+        # TODO: Implement message type as part of the key (serder.ilk)
+        sad = serder.sad
+        sourceAid = sad.get("i", "")
```
kentbull, line 97: If you do `ilk = sad.get("t", "")` and then return `(sourceAid, ilk)` then you can get support the message type.

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+        windowSize, driftSkew = self.getWindowParameters(sourceAid, messageType)
+
+        # Convert both timestamps to seconds since epoch for comparison
+        currentTime = helping.fromIso8601(helping.nowIso8601()).timestamp()
+
+        messageTime = helping.fromIso8601(timestamp).timestamp()
+
+        if (messageTime < currentTime - driftSkew - windowSize or
+                messageTime > currentTime + driftSkew):
+            raise kering.ValidationError(f"Message is out of time window {serder.pretty()}")
```
kentbull, line 162: In some places you do `return False, <str>` and in other places you raise an error. It would make more sense to use one approach for everything. If you want the `checkMessageTimeliness` to never throw then only ever return a `True, msg` or `False, msg` tuple and never throw an exception or only ever return `True` and then throw exceptions for all timeliness validation failures.

And, what will be catching the errors for timeliness failures? Should it halt existing message processing at the parser level, as in should a KRAM failure bubble up to the Parser and just cause the message to be discarded? If so then there should be logging in the Parser when a KRAM failure occurs so that INFO or DEBUG level logging can expose when KRAM failures are causing messages to not be accepted by the parser.

And, if you do want to use exceptions then I suggest adding a new exception class for KRAM timeliness errors, maybe a KRAMError or KRAMTimelinessError. Something like that.

kentbull: Using a generic `kering.ValidationError` prevents the caller of `checkMessageTimeliness` from being able to have a more focused response to what, in my mind, could be a KRAM error.  We can give the caller more information by using a custom error subclass if error handling is the way we want to go. It does seem more idiomatic in Python to throw errors when KRAM time window validation fails.

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+            self._storeTimestamp(cacheKey, messageTime)
+            # Message accepted, new entry
+            return True
+
+        if messageTime > cachedTimestamp:
+            self._storeTimestamp(cacheKey, messageTime)
+            # Message accepted, updated cached entry
+            return True
+
+        if messageTime == cachedTimestamp:
```
kentbull, line 178: Is this too strict? As in, if any two messages have the same timestamp should they really be considered a replay? It would seem comparing the message digests in addition to the timestamp is also necessary to determine something is a replay.

src/keri/core/kraming.py
```diff
@@ -0,0 +1,231 @@
...
+        for said, serder in self.db.krms.getFullItemIter():
+            try:
+                # assumes serder is a SerderKERI, more processing may be needed
+                if self.checkMessageTimeliness(serder):
+                    # TODO: Implement escrowing functionality
+                    self.db.krms.rem(said)
+                    logger.info(f"Message accepted: {serder.pretty()}")
+            except kering.ValidationError as e:
+                logger.error(f"Invalid message: {e}")
+            self.pruneCache()
```
kentbull, line 218: Should this `pruneCache()` call occur no matter what is in the `.krms` DB? As in, should it be shifted to the left by one indentation level?

Another way of saying it is should prune cache only occur when there are items in the `.krms` DB or should it always occur on each recur of the KramDoer?

src/keri/core/kraming.py
arilieb: That's actually how I've done it in my working copy! Haven't written tests for this feature yet.

src/keri/core/kraming.py
arilieb: I am not opposed to changing it to fit whatever the best practice is; there other places in the codebase where we serialize this way when not handling KERI event data.

src/keri/core/kraming.py
arilieb: My understanding is if they have the same timestamp they should be treated as a replay by the time they are validated by the kraming module. The KRAM white paper describes the cache as monotonically ordered, which seems to imply that this is the appropriate level of strictness. Open to corrections/discussion here.

src/keri/core/parsing.py
```diff
@@ -132,7 +133,7 @@ class Parser:
 
     def __init__(self, ims=None, framed=True, piped=False, kvy=None,
                  tvy=None, exc=None, rvy=None, vry=None, local=False,
-                 version=Vrsn_2_0):
+                 krm=None, version=Vrsn_2_0):
```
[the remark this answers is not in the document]
arilieb: Full implementation plan is to have a KRAM escrow since point KRAM only works for non-trans aids. Open to discussion, just worth considering that the current parsing integration will probably be replaced with something more robust and non-interdependent.

tests/core/test_kraming.py
arilieb: Made this change

Issue #1051, opened 2025-07-07
Ability to store metadata with an identifier


iFergal:
### Feature request description/rationale

I think it would be useful to have a way to store metadata about an identifier which is created beyond the alias, such as a profile picture, or user selected identifier theme etc.

The Organizer for contacts already has functionality for regex and chunking images, so I wonder if it could be generalized to organize both a contacts DB and an identifier metadata DB.

Or perhaps there's a better place for it.

pfeairheller:
Discussed in July 22, 2025 dev call.  General agreement to include this functionality in KERIpy.

PR #1084, opened 2025-09-03
Ability to store metadata with an identifier #1051

Sotatek-Patrick-Vu:
Feature request description/rationale here [WebOfTrust/keripy/issues/1051](https://github.com/WebOfTrust/keripy/issues/1051)

### DIFF
--- a/src/keri/app/__init__.py
+++ b/src/keri/app/__init__.py
@@ -5,3 +5,5 @@ KERI
 keri.app package
 
 """
+
+from .organizing import Organizer, IdentifierOrganizer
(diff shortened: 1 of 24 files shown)


iFergal:
Thanks @m00sey it's currently `keri.app.connecting`. So we can remove from the `app` package for sure.

While we can have `keri.organizing`, maybe `keri.wallet.organizing` might be better in case we have further keripy features that are related? Not sure. This kind of metadata feels "wallet" related, but I know there is already `keri.vc.walleting` w.r.t to credentials. I'm happy with whatever you guys think.

Sotatek-Patrick-Vu:
> @iFergal We discussed this at the 12/2 KERIpy maintainers meeting and came up with the following:
> 
> rename connecting.py to organizing.py and export the Organizer from the keri.app **init**.py file. So externally folks can refer to the Organizer as keri.app.Organizer and internally as keri.app.organizing.Organizer.

Thank you, I changed as your suggestion.

Issue #503, opened 2023-05-16
Fix order of witness receipts relative to delegation approval.


pfeairheller:
In the current implementation of delegated inception and rotation (including in multisig delegation), the delegate controller waits for delegation approval before publishing the event to the witnesses.  Since the witnesses are under the control of the controller, this order should be inverted.

To accomplish this, witnesses should accept delegated AID events without requiring a delegator approval anchor but not publish the events externally until the delegator's approval anchor is received.  Once watchers are in place the witnesses can publish the event externally without the approval because the watcher's won't accept the event until the approval shows up.

pfeairheller:
No, considering the issue you just opened, it is not complete for multisig delegation and needs work there, I believe.

kentbull:
This is fixed by https://github.com/WebOfTrust/keripy/pull/1100. Once we get that merged then we can close this issue.

Issue #1087, opened 2025-09-08
kli delegate confirm hangs on KERIpy 1.2.x+


kentbull:
### Version

KERIpy 1.2.x+

### Environment

Mac OS, Python 3.12.6

### Expected behavior

`kli delegate confirm` should complete after a reasonable amount of time and not hang forever.

### Actual behavior

`kli delegate confirm` hangs forever in KERIpy 1.2.x+. This does not happen with KERIpy 1.1.32, hinting that the problem has to do with the delegation logic change between 1.1.x and 1.2.x.

### Steps to reproduce

1. Check out the qvi-software repository
2. Run the `qvi-workflow/kli_only` workflow (along with witnesses and vLEI-server running in the background)
3. The process will hang when the delegate members query keystate from the multisig delegator.

PR #1100, opened 2025-11-20
fix: kli delegate confirm no longer hangs; and other delegated ixn fixes

kentbull:
This PR fixes the `kli delegate confirm` hanging by loosening the delegation approval check to include group multisig events, specifically `dip` and `drt` events.

Also, group ixns (interaction events) now skip unneeded approval check and OOBI resolution of a delegate no longer re-escrows already approved dip events.

Prior to this PR when performing OOBI resolution then any approved dip events were being re-added to the delegables escrow for another approval, which is incorrect because the dip was already approved. Adding a check during Kever.processEvent to look up the sequence number and digest of the seal fixes this.

And group interaction events do not need delegation approval and thus should not be sent to the delegables escrow. This PR puts a type check for dip and drt around the logic that adds items to the delegables delegation approval escrow.

There was also a third bug in the addition of dip and drt events to the delegables escrow. Only single sig events were being added to the escrow because the self.locallyDelegated() check was too restrictive and did not work for group multisig events. Adding the `Kever.locallyDelegated` implementation check of `self.prefixes` loosened the check enough to work for group multisig events.

All existing tests pass, yet this PR is needed because the existing tests around multisig ixn and dip event escrowing  do not cover multisig delegation. I will add those tests to this PR so we can avoid breakages here in the future.

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -2421,25 +2421,16 @@ class Kever:
 
         if serder.ilk in (Ilks.dip, Ilks.drt) and self.locallyDelegated(delpre) and not self.locallyOwned():  # local delegator of delegated event
             if delseqner is None or delsaider is None:  # missing delegation seal
-                # Before escrowing, check if we already have the seal in our KEL
-                # This handles the case where delegation was already approved and
-                # we're receiving the event via OOBI resolution or query
-                seal = dict(i=serder.pre, s=serder.snh, d=serder.said)
-                dserder = self.db.fetchLastSealingEventByEventSeal(pre=delpre, seal=seal)
-                if dserder is not None:  # found seal - use it instead of escrowing
-                    delseqner = coring.Seqner(sn=dserder.sn)
-                    delsaider = coring.Saider(qb64=dserder.said)
-                else:
-                    # Seal not found - escrow delegable. So local delegator can approve OOB.
-                    # and create delegator event with valid event seal of this
-                    # delegated event and then reprocess event with attached source
-                    # seal to delegating event, i.e. delseqner, delsaider.
-                    self.escrowDelegableEvent(serder=serder, sigers=sigers,
-                                              wigers=wigers, local=local)
-                    msg = f"Missing approval for delegation by {delpre} of event = {serder.said}"
-                    logger.info(msg)
-                    logger.debug("Event Body=\n%s\n", serder.pretty())
-                    raise MissingDelegableApprovalError(msg)
+                # so escrow delegable. So local delegator can approve OOB.
+                # and create delegator event with valid event seal of this
+                # delegated event and then reprocess event with attached source
+                # seal to delegating event, i.e. delseqner, delsaider.
+                self.escrowDelegableEvent(serder=serder, sigers=sigers,
+                                          wigers=wigers, local=local)
+                msg = f"Missing approval for delegation by {delpre} of event = {serder.said}"
+                logger.info(msg)
+                logger.debug("Event Body=\n%s\n", serder.pretty())
+                raise MissingDelegableApprovalError(msg)
 
         # validateDelegation returns (None, None) when delegation validation
         # does not apply. Raises ValidationError if validation applies but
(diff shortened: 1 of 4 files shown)


kentbull:
Fixes https://github.com/WebOfTrust/keripy/issues/503

src/keri/core/eventing.py
```diff
@@ -2388,7 +2419,10 @@ def valSigsWigsDel(self, serder, sigers, verfers, tholder,
         # seal in this case can't be malicious since sourced locally.
         # Doesn't get to here until fully signed and witnessed.
 
-        if self.locallyDelegated(delpre) and not self.locallyOwned():  # local delegator
+        # should only run for delegated inception and rotation, not interaction. Ixn does not require approval.
+        if serder.ilk in (Ilks.dip, Ilks.drt) \
```
kentbull: Have this be a `not Ilks.ixn`

src/keri/app/cli/commands/delegate/confirm.py
```diff
@@ -170,11 +168,11 @@ def confirmDo(self, tymth, tock=0.0):
                             saider = self.hby.db.cgms.get(keys=(prefixer.qb64, sner.qb64))
                             if saider is not None:
                                 break
-
                             yield self.tock
 
-                        print(f"Delegate {eserder.pre} {typ} event committed.")
+                        print(f"Delegate {typ} event {eserder.pre} committed.")
```
kentbull: Print statements in this file were reordered for readability.

src/keri/app/cli/commands/delegate/confirm.py
```diff
@@ -211,27 +209,23 @@ def confirmDo(self, tymth, tock=0.0):
...
                             self.witq.query(src=hab.pre, pre=eserder.pre, sn=eserder.sn, wits=wits)
-
                             while eserder.pre not in self.hby.kevers:
                                 yield self.tock
 
-                            print(f"Delegate {eserder.pre} {typ} event committed.")
+                        print(f"Delegate {typ} event {eserder.pre} committed.")
 
-                        self.hby.db.delegables.rem(keys=(pre, sn))
+                        self.hby.db.delegables.rem(keys=(pre, sn), val=edig)
```
kentbull: The `.rem()` logic was changed here to mirror what is found in the escrow logic in `core/eventing.py`

src/keri/core/eventing.py
```diff
@@ -2388,19 +2419,27 @@ def valSigsWigsDel(self, serder, sigers, verfers, tholder,
         # seal in this case can't be malicious since sourced locally.
         # Doesn't get to here until fully signed and witnessed.
 
-        if self.locallyDelegated(delpre) and not self.locallyOwned():  # local delegator
-            # must be local if locallyDelegated or caught above as misfit
+        if serder.ilk in (Ilks.dip, Ilks.drt) and self.locallyDelegated(delpre) and not self.locallyOwned():  # local delegator of delegated event
```
kentbull: This is the big fix to `kli delegate confirm` that solves https://github.com/WebOfTrust/keripy/issues/1087. The reason this solves #1087 is because interaction events would incorrectly pass through this block to be escrowed before the type guard `if serder.ilk in (Ilks.dip, Ilks.drt)` was added.

Only delegated inception and delegated rotation should be going to the delegables escrow.

src/keri/core/eventing.py
```diff
@@ -1682,27 +1682,58 @@ def locallyOwned(self, pre: str | None = None):
...
+        if pre in self.groups:  # local group delegator
+            habord = self.db.habs.get(keys=(pre,))
+            return habord.mid in habord.smids  # True not stale, False stale
+
+        return pre in self.prefixes  # otherwise local non-group delegator
+
         """
         pre = pre if pre is not None else ""
-        return self.locallyOwned(pre=pre)
+        return pre in self.prefixes
```
kentbull: This is part of the big delegation fix. `self.locallyOwned(...)` was too strict of a check and excluded multisig events.

`pre in self.prefixes` works because `self.prefixes` includes multisig identifiers.

kentbull:
Closed in favor of https://github.com/WebOfTrust/keripy/pull/1111 and https://github.com/WebOfTrust/keripy/pull/1114

Issue #307, opened 2022-09-21
Have to share delegated AID with delegator after completion


pfeairheller:

lenkan:
Is this still an issue? 

I if I understand correctly, the delegate inception event is fully signed and receipted before the delegator anchors the approval. Therefore this is a non-issue.

kentbull:
This is still an issue, as in the delegator does not have the receipts of the event from the delegate.

The delegate should send the equivalent of an introduction (OOBI resolve) back to the delegator once the delegation approval occurs. And, in my opinion, it would be useful if the delegator were to send a keystate update to the delegate to automate notifying the delegate of delegation approval, maybe in an exn. For multisig then each of the delegator members would send a notification of delegation approval to the delegate so that the delegate would not have to query the delegator's KEL. This could also be done with exns.

PR #1109, opened 2026-01-09
fix,test: add multisig delegation test and test Doers; fix incorrect dip,drt re-escrow

kentbull:
This commit contains one big thing and a few small things, with heavy use of Claude Opus 4.5:

## Big thing - add tests for the multisig delegation process

- Fix earlier this year: Core multisig delegation escrow fix in eventing.py: added type guard `if serder.ilk in (Ilks.dip, Ilks.drt)` to ensure interaction events are not sent for approval and only `dip` and `drt` events are.
  - Tests for this:
    - tests/app/test_grouping.py - Updated with test_multisig_delegate() which covers the full end-to-end workflow:
    - tests/app/app_helpers.py - Contains:
         - EscrowDoer - Fast escrow processing for tests
         - openWit() / openCtrlWited() - Context managers for witness and controller setup
         - HabHelpers - OOBI resolution, witness receipts, and delegation seals
         - MultisigInceptLeader / MultisigInceptFollower - DoDoers for multisig inception coordination
         - MultisigDelegationApprover - DoDoer for delegation approval workflow
         - KeystateQueryDoer - Keystate discovery for anchor completion

The test is the majority of this PR. 

This test is a new kind of test for the KERIpy codebase, one that operates at the level of Doers rather than using the internals of Parser, event handlers like Kevery, Tevery, and escrow processing. It uses the code similar to how the KLI does.

The multisig delegation test includes:
- one witness to use to transmit messages back and forth
- Delegator multisig formation
- Delegate participants setup
- OOBI exchanges
- Delegation approval by both delegator participants
- Keystate queries to discover the approval anchor
- OOBI resolution of the delegate by delegator
- Final verification assertions

## Medium things

- In `core/eventing.py` this adds a delegation seal check prior to escrowing a `dip` or `drt` to cover the case of a delegator OOBIing with the delegate AFTER approving the delegate's `dip` or `drt`. 
  - What was happening was that even AFTER approving the delegation the `dip` was ending up back on the `.delegables` escrow following an OOBI resolution of the delegate by the delegator *even though* the delegator had already approved the `dip`. Looking up the seal again here is necessary to avoid re-escrowing an already approved `dip` or `drt` event.
  - This prepares the codebase for the delegatee KEL sharing mechanism Phil mentions in #307.
- `kli multisig join` fix with `if "di" in ked else None` to `if "di" in oicp.ked else None`

## Small Things
- pass cf through in habbing.py so that `temp=True` works in tests.
- delegables.rem() fix in delegate/confirm.py

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -2636,25 +2636,16 @@ class Kever:
         if (serder.ilk in (Ilks.dip, Ilks.drt) and self.locallyDelegated(delpre)
                 and not self.locallyOwned()):  # local delegator of delegated event
             if delseqner is None or delsaider is None: # missing delegation seal
-                # Before escrowing, check if we already have the seal in our KEL
-                # This handles the case where delegation was already approved and
-                # we're receiving the event via OOBI resolution or query
-                seal = dict(i=serder.pre, s=serder.snh, d=serder.said)
-                dserder = self.db.fetchLastSealingEventByEventSeal(pre=delpre, seal=seal)
-                if dserder is not None:  # found seal - use it instead of escrowing
-                    delseqner = coring.Seqner(sn=dserder.sn)
-                    delsaider = coring.Saider(qb64=dserder.said)
-                else:
-                    # Seal not found - escrow delegable. So local delegator can approve OOB.
-                    # and create delegator event with valid event seal of this
-                    # delegated event and then reprocess event with attached source
-                    # seal to delegating event, i.e. delseqner, delsaider.
-                    self.escrowDelegableEvent(serder=serder, sigers=sigers,
-                                              wigers=wigers, local=local)
-                    msg = f"Missing approval for delegation by {delpre} of event = {serder.said}"
-                    logger.info(msg)
-                    logger.debug("Event Body=\n%s\n", serder.pretty())
-                    raise MissingDelegableApprovalError(msg)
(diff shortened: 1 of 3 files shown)


src/keri/core/eventing.py
```diff
@@ -2636,16 +2636,25 @@ def valSigsWigsDel(self, serder, sigers, verfers, tholder,
...
-                # and create delegator event with valid event seal of this
-                # delegated event and then reprocess event with attached source
-                # seal to delegating event, i.e. delseqner, delsaider.
-                self.escrowDelegableEvent(serder=serder, sigers=sigers,
-                                          wigers=wigers, local=local)
-                msg = f"Missing approval for delegation by {delpre} of event = {serder.said}"
-                logger.info(msg)
-                logger.debug("Event Body=\n%s\n", serder.pretty())
-                raise MissingDelegableApprovalError(msg)
+                # Before escrowing, check if we already have the seal in our KEL
```
kentbull: This is a more subtle error I encountered while writing the test. The bug shows up for a delegator only after delegation completes and the delegator is trying to OOBI with the delegate in order to see the delegate's KEL in .kevers.

What was happening was that the dip event was incorrectly getting put back in to the .delegables escrow in response to the KEL replay from the OOBI resolve even though the dip was already approved.

kentbull:
Closed in favor of https://github.com/WebOfTrust/keripy/pull/1123

PR #1117, opened 2026-01-18
fix: remove dip and drt from escrows once approved

kentbull:
Prior to this fix then only single sig workflows removed items from the delegables escrow. Somehow multisig dip and drt were missed. Both dip and drt should be removed from the delegables escrow once they are approved.


kentbull:
Similar to https://github.com/WebOfTrust/keripy/pull/1111 on main and https://github.com/WebOfTrust/keripy/pull/1118 on v1.3.2

PR #1125, opened 2026-01-19
test: fix the flakiness of kli tests by reloading parser context

kentbull:
KLI tests have been flaky on recent CI test runs and I found out how to fix this by resetting the argument parsing context.

Reload commands module to ensure fresh parser objects


kentbull:
This may be a multicommand bug

PR #1141, opened 2026-01-21
Two new kli utilities and a minor improvement to IPEX grant messages

pfeairheller:
This PR includes the following:

* A new credential schema kli utility to import schema from a file.
* A new AID kli utility to delete an identifier from your KERI database.
* Modification to the IPEX grant method to allow for including the credential registry inception event as an embed to the ipex grant message.
* Pin multicommand to 1.0.0 to avoid a bug in the newly released 2.0.0 (this was discovered and already fixed on main).

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -16,14 +16,27 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ macos-13, ubuntu-latest, windows-latest ]
+        os: [ macos-latest, ubuntu-latest, windows-latest ]
 
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


src/keri/app/cli/commands/vc/create.py
```diff
@@ -218,8 +218,6 @@ def createDo(self, tymth, tock=0.0, **kwa):
         dt = self.creder.attrib["dt"] if "dt" in self.creder.attrib else helping.nowIso8601()
         iserder = registry.issue(said=self.creder.said, dt=dt)
 
-        #vcid = iserder.ked["i"]
```
m00sey: thank you for leaving the campsite cleaner than you found it

m00sey (approved):

Issue #1163, opened 2026-02-02
Remove technical debt. Upgrade all databases in Baser to use Suber subclasses.


SmithSamuelM:
### Feature request description/rationale

# Migrate Baser subdbs  to all use Suber

self.aess = self.env.open_db(key=b'aess.')  CatCesrSuber
Later refactor to replace Saider with Diger since never use .saidify on sealsource
of a key event. This is done by SerderKERI

self.evts = self.env.open_db(key=b'evts.') 
replace with SerderSuber() assumes default klas=SerderKERI

self.fels = self.env.open_db(key=b'fels.')
replace with OnSuber.  Note the returned value is SAID of an event msg as
qb64 str not qb64b bytes nor a Diger or Saider instance.
But this value is only ever used as an index so in this case we do not want to
serdes as Diger. So return value of OnSuber.get in this case is qb64 already

self.dtss = self.env.open_db(key=b'dtss.')
Replace with CesrSuber() with klas set to coring.Dater
This enables use of Dater methods  .dts and .datetime to remove redundant
code. Note everywhere the CesrSuber.get returns a Dater instance not a dts string
so call dater.dts to get the string and dater.datetime to get the datetime object
for datetime calculations

self.kels = self.env.open_db(key=b'kels.', dupsort=True)
Replace with IoDupSuber. Note the returned value is SAID of an event msg as
qb64 str not qb64b bytes nor a Diger or Saider instance.
But this value is only ever used as an index so in this case we do not want to
serdes as Diger. So return value of IoDupSuber.get in this case is qb64 already

self.sigs = self.env.open_db(key=b'sigs.', dupsort=True)
Replace with CesrDupSuber with klas set to coring.Siger
Note everywhere the CesrDupSuber.get returns a Siger instance not qb64b signature
But most cases the signature is converted to a Siger in order to verify signature

self.wigs = self.env.open_db(key=b'wigs.', dupsort=True)
Replace with CesrDupSuber with klas set to coring.Siger
Note everywhere the CesrDupSuber.get returns a Siger instance not qb64b signature
But most cases the signature is converted to a Siger in order to verify signature

self.rcts = self.env.open_db(key=b'rcts.', dupsort=True)
Replace with CatCesrDupSuber with klas set to (coring.Prefixer, coring.Cigar)
Note that the return value is a duple (two tuple) of matter subclasses
not a concatenated couple of qb64b values. This means everywhere this is replaced
[...]

PR #1171, opened 2026-02-06
Feature refractor baser domains

KeaxD:
For issue #1163 

Refractor Ooes and Pses domain:
- Deleted basing.py functions related to those domains
- Refractored the baser api to delegate to ooes and pses suber classes
- Adapted the code to support return str val instead of bytes 
- Added test cases to ensure correct IoDupSuber behavior for both Ooes and Pses( could be shorten to just 1 test suite that each subclass could use)
- Added 1 test case for Ooes where out of order escrows were missing records (evt, sigs, dts) to test proper clean up.

### DIFF
--- a/src/keri/app/grouping.py
+++ b/src/keri/app/grouping.py
@@ -534,14 +534,13 @@ def getEscrowedEvent(db, pre, sn):
 
     dig = bytes(dig)
     key = dbing.dgKey(pre, dig)  # digest key
-    msg = db.getEvt(key)
-    serder = serdering.SerderKERI(raw=bytes(msg))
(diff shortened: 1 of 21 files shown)


src/keri/app/grouping.py
SmithSamuelM: Looks  like OnIoDupSuber is missing a method. GetLastOn.  Until we add that method the best approach is to use the super classe IoDupSuber's  getLast method and include the the serialized (onKey) on in the key . Looks like there are several plases where this needs to be fixed to use the super class method. Until I can add a getLastOn

PR #1198, opened 2026-02-16
Feature refractor taes twes oots

KeaxD:
For issue #1163 

Refractor taes, twes, and oots domains to OnIoDupSuber:

- Deleted viring.py functions related to those domains
- Refractored the viring api to delegate to each suber classes
- Adapted the tests for proper testing of OnIoDupSuber behavior

I also messed up my branch which is why the kels refractor commit is included but it should be the same changes as the kels refractor in my other PR


SmithSamuelM (commented):
Se comment on the keys pull earlier as it applies to this one as well.

PR #1199, opened 2026-02-16
Refactor ures to suber

alecforston:


SmithSamuelM (commented):
Everywhere you see inline dgKey replace with just tuple  of keys.

PR #1210, opened 2026-02-20
CesrSuber and CatCesrSuber serialization interface

### LINKED ISSUE #752: Modify serialization interface to CesrSuber and its subclasses CatCesrSuber etc.
Currently the interface for serializaton requires passing in primitive instance in order to serialize. However when creating the Instance the .klas variable holds the class. So it would be possible to also allow simply passing in an already serialized value or tuple of values. 

Also the ._ser method does not check that the .klas matches the ._ser instance types. This would allow someone to write the wrong serialization. 

So might want to add a strict attribute that enforces subclass of ._ser to match .klas
finally with strict then one could force des/ser using .klas when passed in already serialized. The last would avoid having to 
pass around the instances everywhere when the only purpose of the instances is to satisfy the interface for CatCesrSuber

evanja57:
In this PR:
- Added a strict boolean parameter to CESR suber initialization (CesrSuber/CatCesrSuber via base classes), defaulting to False for backward compatibility
- When strict=True, _ser validates write-time type conformance against .klas (and arity/order for Cat subers), raising TypeError/ValueError before bad data is written
- Added focused tests for fail on write behavior and inherited strict enforcement in cat derivatives

Fixes #752

### DIFF
--- a/src/keri/db/subing.py
+++ b/src/keri/db/subing.py
@@ -1052,9 +1052,7 @@ class CatCesrSuberBase(CesrSuberBase):
 
         vals = tuple(val) if self.strict else val
         if self.strict:
-            if len(vals) != len(self.klas):
-                raise ValueError(f"Expected {len(self.klas)} values, got {len(vals)}.")
-            for klas, item in zip(self.klas, vals):
+            for klas, item in zip(self.klas, vals, strict=self.strict):
                 if not isinstance(item, klas):
                     raise TypeError(f"Expected {klas}, got {type(item)}.")
 


SmithSamuelM (commented):
The strict init setup looks good but there seems to be some confusion about how the _ser method should be refactored.  there is a  a lot of unexpected code.  I will have to download the full code to be able to understand it.

evanja57:
Thanks Sam, that makes sense.

For _ser, I was trying to stick to the two points in #752:

1. allow already serialized inputs (single value, or tuple for CatCesrSuber)
2. when strict=True, enforce .klas compatibility at write time (including tuple slot order for CatCesrSuber)

So the extra branches are just for those two paths, not meant as a broader API change. If this feels too wide, I’m happy to trim it down to the smallest version you want.

Issue #420, opened 2022-12-31
Convert all ints JSON representations of sequence numbers for events to hex str or CESR us Number


SmithSamuelM:
The largest int number possible in Javascript  and hence any JSON to be processed by javascript is 2^53 -1  with is the max int for 64 bit floating point.
The largest sequence number in KERI is 2^128-1 which is practically infinite but this number is much larger than a  javascript int.
Large numbers like this (2^128-1) is not a problem for Python with employs variably sized ints with expand to accomodate the number value. 

When JSON is deserialized by python on both ends ,json.dumps to  json.loads, the big numbers round trip just fine. But when the JSON is meant to be processed by Javascript (or a transpiled Javascript like TypeScript) the big numbers do not round trip.
Therefore to ensure we don't encounter such latent bugs, we should convert the ints that represent sequence numbers and first seen ordinal numbers in any JSON that might be processed by Javascript to either hex string or CESR strings using core.coring.Number.

A Number instance processes both hex strings, ints, and CESR encoded number. For CESR it automatically resizes with its CESR code and size to match the number value. So for small value sequence numbers only 4 characters (three bytes) is required. For small numbers hex strings are more efficient for different ranges of very large numbers CESR is more efficient. There are four codes that number uses. Short, Long, Big, Huge.

PR #1247, opened 2026-02-26
Convert Seqner to Number for cdel

KeaxD:
for issue #420 

Cdel is called inside complete() from an anchorer object so I tried to look for each instance where anchorer.complete is called and changed from Seqner to Number.huge because it is used in cdel's keyspace

There's also some changes in keeping.py because I messed up my branch again. The changes are the same from my other smids and rmids PR.

### DIFF
--- a/ref/ChangeLog.md
+++ b/ref/ChangeLog.md
@@ -1,5 +1,13 @@
 # Change Log for keripy
 
+## 2.0.0-dev5
+### Backwards breaking interface changes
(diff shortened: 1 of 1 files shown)


KeaxD:
I refractored cdel to be a CesrOnSuber db.

Also I modified the parameters of Anchorer's complete() located in keri/app/delegating.py from Seqner to Number. I will also put an update in the changeLog.md

PR #1260, opened 2026-02-28
Fix scripts multisig rotation race condition

ryan-hansen:
Problem: The multisig-join.sh CLI script was hanging. 

Two causes: 
1 - A timing race: when one member queried the other’s KEL after local rotations, the query could be answered before the witness had applied that member’s rotation, so the group rotation was built with stale keys and failed with “invalid rotation, new key set unable to satisfy prior next signing threshold”. 

2 - When the rotate failed, the companion multisig join had no timeout and waited indefinitely for a group event that never came, so the script’s wait never returned.

Fix: 
1 - Added a --timeout option to kli multisig join so it exits with an error instead of hanging when no group multisig event arrives.

2 - Added retry logic in the group rotation CLI so transient ValidationErrors (e.g. from stale member KEL state) are retried for a short window before failing. 

3 - Hardened the demo script by using --timeout on all join calls and adding short sleeps before cross-queries so witnesses have time to apply rotations. 

4 - Added kering.TimeoutError for join timeouts.

### DIFF
--- a/scripts/demo/basic/multisig-join.sh
+++ b/scripts/demo/basic/multisig-join.sh
@@ -24,6 +24,10 @@ pid=$!
 PID_LIST+=" $pid"
 
 wait $PID_LIST
+if [ $? -ne 0 ]; then
+  echo "Multisig group inception/join failed"
+  exit 1
+fi
 
 kli status --name multisigj1 --alias multisig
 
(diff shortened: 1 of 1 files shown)


ryan-hansen:
Not quite ready yet, apparently.

PR #1284, opened 2026-03-04
feat: migrate imgs/iimgs DBs from raw env.open_db to subing.Suber

### LINKED ISSUE #1163: Remove technical debt. Upgrade all databases in Baser to use Suber subclasses.

jaelliot:
## Summary

Eliminates the last two raw `env.open_db()` calls in business logic from `Baser`, completing the Suber migration tracked in #1163.

### `src/keri/db/basing.py`
- `self.imgs`: `self.env.open_db(key=b'imgs.')` → `subing.Suber(db=self, subkey='imgs.')`
- `self.iimgs`: `self.env.open_db(key=b'iimgs.')` → `subing.Suber(db=self, subkey='iimgs.')`
- `clean()` copy routine updated to use `.sdb` to access the underlying LMDB handle and `Baser.setVal` for writing

### `src/keri/app/organizing.py`
- All `setVal(db=self.imgsdb, ...)`, `getVal(db=self.imgsdb, ...)`, and `delTop(db=self.imgsdb, ...)` calls updated to `self.imgsdb.sdb`
- Minimal-diff approach: preserves existing chunked binary storage key scheme (`pre.content-type`, `pre.0`…`pre.N`, `pre.content-length`) — no semantic change, just lifts the raw handle from the Suber wrapper

## Why `.sdb` instead of Suber methods

The `imgs`/`iimgs` stores use a bespoke multi-key chunked binary format (raw bytes values, non-CESR) that doesn't map to any existing Suber subclass API. Using `.sdb` to get the raw LMDB handle is the correct minimal migration — it removes the last `open_db` calls while keeping the access pattern intact until a future refactor adds a proper chunked-image Suber subclass.

## Testing
- `tests/app/test_organizing.py` — 7/7 ✅  
- `tests/db/test_basing.py` — 14/14 ✅

Closes #1163

### DIFF
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -919,10 +919,10 @@ class Baser(dbing.LMDBer):
                                              klas=(coring.Prefixer, coring.Cigar))
         self.ures = subing.CatCesrIoSetSuber(db=self, subkey='ures.',
                                              klas=(coring.Diger, coring.Prefixer, coring.Cigar))
-        self.vrcs = subing.CatCesrIoSetSuber(db=self, subkey='vrcs.', 
-                                            klas=(coring.Prefixer, coring.Number, coring.Diger, indexing.Siger))
-        self.vres = subing.CatCesrIoSetSuber(db=self, subkey='vres.', 
-                                            klas=(coring.Diger, coring.Prefixer, coring.Number, coring.Diger, indexing.Siger))
+        self.vrcs = subing.CatCesrIoSetSuber(db=self, subkey='vrcs.',
+                        klas=(coring.Prefixer, coring.Number, coring.Diger, indexing.Siger))
+        self.vres = subing.CatCesrIoSetSuber(db=self, subkey='vres.',
+                        klas=(coring.Diger, coring.Prefixer, coring.Number, coring.Diger, indexing.Siger))
         self.pses = subing.OnIoDupSuber(db=self, subkey='pses.')
         self.pwes = subing.OnIoDupSuber(db=self, subkey='pwes.')
         self.pdes = subing.OnIoDupSuber(db=self, subkey='pdes.')
@@ -1182,9 +1182,11 @@ class Baser(dbing.LMDBer):
         # TODO: clean
         self.ccigs = subing.CesrSuber(db=self, subkey='ccigs.', klas=coring.Cigar)
 
-        # Chunked image data for contact information for remote identifiers
-        # TODO: clean
-        self.imgs = subing.Suber(db=self, subkey='imgs.')
+        # Blinded media for contact information for remote identifiers.
+        # CatCesrSuber with TypeMedia format: (Noncer=SAID, Noncer=UUID, Labeler=MIME, Texter=data)
+        self.imgs = subing.CatCesrSuber(db=self, subkey='imgs.',
+                                         klas=(coring.Noncer, coring.Noncer,
+                                               coring.Labeler, coring.Texter))
 
         # Field values for identifier information for local identifiers. Keyed by prefix/field
         # TODO: clean
@@ -1200,9 +1202,11 @@ class Baser(dbing.LMDBer):
         # TODO: clean
         self.icigs = subing.CesrSuber(db=self, subkey='icigs.', klas=coring.Cigar)
 
-        # Chunked image data for identifier information for local identifiers
-        # TODO: clean
-        self.iimgs = subing.Suber(db=self, subkey='iimgs.')
+        # Blinded media for identifier information for local identifiers.
+        # CatCesrSuber with TypeMedia format: (Noncer=SAID, Noncer=UUID, Labeler=MIME, Texter=data)
+        self.iimgs = subing.CatCesrSuber(db=self, subkey='iimgs.',
+                                          klas=(coring.Noncer, coring.Noncer,
+                                                coring.Labeler, coring.Texter))
 
         # Delegation escrow dbs #
         # delegated partial witness escrow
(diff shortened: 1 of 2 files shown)


jaelliot:
Rebase update: this branch has been rebased onto current `upstream/main` and force-pushed.

Latest commit: `9556eaee`

Post-rebase validation:
- `pytest -q tests/db/test_basing.py` (14 passed)

This includes conflict resolution in `basing.py` during rebase while keeping the `CatCesrSuber` TypeMedia migration commit intact.

@kentbull can you take a fresh look and clear the stale changes-requested lock if this aligns with the latest direction?

PR #1295, opened 2026-03-05
fix: use pre= instead of wits= in kli admit telquery (fixes #1160)

### LINKED ISSUE #1160: kli admit fails when issuer of credential has no witness
Steps to reproduce:

1. Create an AID ISSUER without witnesses, for example using Signify-TS and KERIA.
2. Create an AID HOLDER with using keripy, with at least one witness.
3. Mutually resolve OOBIs
4. Create a credential from ISSUER, grant to HOLDER
5. `kli ipex list --poll` from HOLDER
6. `kli ipex admit --said <grant>` from ISSUER

Admit fails with NoneType is not iterable. Note: since the admit.py does not actually wait for the witq to send the message, the process might exit before the error occurs in some circumstances.

However, looking at the code, there is an obvious difference to how KERIA handles this.

### KERIPY

See keripy admit.py:

https://github.com/WebOfTrust/keripy/blob/87bb456e71e42c2e1b898373b7176551658dea9f/src/keri/app/cli/commands/ipex/admit.py#L107-L109

The TEL query will eventually be picked up here:

https://github.com/WebOfTrust/keripy/blob/87bb456e71e42c2e1b898373b7176551658dea9f/src/keri/app/agenting.py#L511-L513

Since `wits` is an empty array and `pre` is `None`, this crashes on `NoneType is not iterable` (deeper down in koming.py).

### KERIA

Notice that KERIA sets the `pre`, instead of the witnesses for the telquery.

https://github.com/WebOfTrust/keria/blob/262b7ce019d1c38fdebfee3df39eb0e080cb2240/src/keria/app/agenting.py#L1353-L1358

```python
            # Lets get the latest KEL and Registry if needed
            self.witq.query(hab=self.agentHab, pre=issr)
            if "ri" in acdc:
                self.witq.telquery(
                    hab=self.agentHab, pre=issr, ri=acdc["ri"], i=acdc["d"]
                )
```

### Possible resolution

Since the WitnessInquisitor resolves the end roles of the query target, we can simply add the issuer prefix to the `telquery` call.

See: https://github.com/WebOfTrust/keripy/blob/87bb456e71e42c2e1b898373b7176551658dea9f/src/keri/app/agenting.py#L515-L536

jaelliot:
## Summary

Fixes #1160 — `kli admit` crashes with `NoneType is not iterable` when the issuer of a credential has no witnesses.

## Root Cause

In `src/keri/cli/commands/ipex/admit.py`, the `telquery` call passed `wits=self.hab.kevers[issr].wits`. When the issuer has no witnesses, this is an empty list `[]`. Downstream in `WitnessInquisitor.msgDo`, `random.choice([])` raises `IndexError`, and passing through `koming.py` with `pre=None` raises `NoneType is not iterable`.

## Fix

Replace `wits=self.hab.kevers[issr].wits` with `pre=issr` in the `telquery` call.

This matches KERIA's behavior (`keria/app/agenting.py:1353-1358`) which passes `pre=issr`, allowing `WitnessInquisitor.msgDo` to resolve endpoints via `hab.endsFor(pre=pre)` — working for issuers with or without witnesses.

## Changes

- **`src/keri/cli/commands/ipex/admit.py`**: 1-line fix — `telquery(src=..., pre=issr, ri=..., i=...)` instead of `telquery(src=..., wits=..., ri=..., i=...)`
- **`tests/app/test_agenting.py`**: Added `test_telquery_uses_pre_not_wits` regression test verifying telquery queues a message with `pre` set and `wits=None` for endpoint resolution

## Testing

- All 4 tests in `test_agenting.py` pass (including new regression test)
- 1 source file changed, 1 test file changed

### DIFF
--- a/src/keri/cli/commands/ipex/admit.py
+++ b/src/keri/cli/commands/ipex/admit.py
@@ -104,7 +104,7 @@ class AdmitDoer(doing.DoDoer):
         # Lets get the latest KEL and Registry if needed
         self.witq.query(src=self.hab.pre, pre=issr)
         if "ri" in acdc:
-            self.witq.telquery(src=self.hab.pre, wits=self.hab.kevers[issr].wits, ri=acdc["ri"], i=acdc["d"])
+            self.witq.telquery(src=self.hab.pre, pre=issr, ri=acdc["ri"], i=acdc["d"])
 
         for label in ("anc", "iss", "acdc"):
             ked = embeds[label]
--- a/tests/app/test_agenting.py
+++ b/tests/app/test_agenting.py
@@ -274,3 +274,37 @@ def test_messenger_prefers_https():
         streamMessengerFrom(hab, pre, https_only, msg=b"test")
         call_url = MockStream.call_args[1]["url"]
         assert call_url == "https://example.com:5643"
+
+
+def test_telquery_uses_pre_not_wits():
+    """Test that WitnessInquisitor.telquery queues a message with `pre` parameter
+    for endpoint resolution instead of `wits`, matching KERIA behavior.
+
+    Regression test for issue #1160: kli admit fails when issuer has no witness
+    because telquery was called with wits=[] (empty list) instead of pre=issr.
+    When wits is an empty list, random.choice([]) raises IndexError.
+    When pre is provided, WitnessInquisitor.msgDo resolves endpoints via
+    hab.endsFor(pre=pre) which works for issuers with or without witnesses.
+    """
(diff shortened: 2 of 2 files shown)


jaelliot:
Status update: this PR is approved and all current checks are green, including Ubuntu, macOS, Windows, coverage, and scripts. Leaving it open for maintainer visibility/merge, but from my side it appears ready for final maintainer action.

PR #1307, opened 2026-03-08
fix: strip prerelease from version comparisons in migrations (#820)

### LINKED ISSUE #820: Semver comparison issues for migrations in same dev release set

jaelliot:
## Summary

Fixes #820 — `semver.compare()` compares alphanumeric prerelease identifiers lexicographically, so `dev4` > `dev10` (because `'4' > '1'`). This breaks version comparison when moving between dev releases within the same version cycle (e.g., `1.2.0-dev4` → `1.2.0-dev10`).

## Root Cause

The `semver` library follows the semver spec which compares alphanumeric prerelease identifiers as strings:

```python
>>> semver.compare("1.2.0-dev4", "1.2.0-dev10")
1  # dev4 is "greater than" dev10!?
```

This causes three problems in `Baser`:

1. **`migrate()`**: Re-runs already-completed migrations when the DB version has a prerelease tag (e.g., DB at `1.2.0-dev4`, migration version `1.2.0` — comparison says migration is "higher")
2. **`current` property**: Could throw false `ConfigurationError` depending on prerelease tag ordering
3. **`complete()`**: Fails to list completed migrations when DB version includes a prerelease tag

## Fix

Added `_strip_prerelease()` helper that normalizes version strings by removing prerelease and build metadata before comparison. Applied to `self.version` (the DB version) in all three comparison sites:

- `migrate()` line 992: skip check for already-run migrations
- `current` line 1064: "DB ahead of library" check  
- `complete()` line 1091: "list completed migrations" filter

This ensures dev releases within the same version cycle (`1.2.0-dev4`, `1.2.0-dev10`, `1.2.0-rc1`) all compare as equal to `1.2.0`.

### `src/keri/db/basing.py`
- Added `_strip_prerelease()` helper function
- Updated 3 `semver.compare()` calls to strip prerelease from DB version before comparison

### `tests/db/test_basing.py`
- Added `test_semver_dev_tag_comparison()` — regression test covering:
  - `_strip_prerelease()` normalizes dev, rc, and build metadata tags
  - All three comparison scenarios (migrate skip, current check, complete listing)
  - Both upgrade (dev4→dev10) and cross-cycle (1.2.0→1.3.0) scenarios

## Testing

- `tests/db/test_basing.py` — **15/15** passed (including new regression test)
- Zero regressions

### DIFF
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -32,6 +32,18 @@ from .. import help
 logger = help.ogler.getLogger()
 
 
+def _strip_prerelease(version_str):
+    """Strip prerelease and build metadata from a semver string.
+
+    Semver compares alphanumeric prerelease identifiers lexicographically,
+    so 'dev4' > 'dev10' (because '4' > '1'). Stripping prerelease ensures
+    dev releases within the same version cycle compare as equal.
+    See: https://github.com/WebOfTrust/keripy/issues/820
+    """
+    ver = semver.VersionInfo.parse(version_str)
+    return str(semver.Version(ver.major, ver.minor, ver.patch))
+
+
 MIGRATIONS = [
     ("0.6.8", ["hab_data_rename"]),
     ("1.0.0", ["add_key_and_reg_state_schemas"]),
@@ -1358,7 +1370,8 @@ class Baser(dbing.LMDBer):
                     f"Skipping migration {version} as higher than the current KERI version {keri.__version__}")
                 continue
             # Skip migrations already run - where version less than (-1) or equal to (0) database version
-            if self.version is not None and semver.compare(version, self.version) != 1:
+            # Strip prerelease from DB version to avoid lexicographic comparison bugs (#820)
+            if self.version is not None and semver.compare(version, _strip_prerelease(self.version)) != 1:
                 continue
 
             # Clear all escrows before first migration to prevent old key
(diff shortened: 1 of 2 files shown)


jaelliot:
Status update: all current checks are green, including Ubuntu, macOS, Windows, coverage, and scripts. Leaving this open for maintainer visibility, but it looks ready for review from my side now that the rebase/conflict cleanup and CI pass are complete.

PR #1341, opened 2026-03-17
fix: remove dead escrow package parser for multicommand 2.0.1 compatibility

### LINKED ISSUE #1332: Multicommand 2.01 is still breaking unit tests Author suggests its a pytest problem and to create a fixture to fix
### Version

2.0.0-dev5

### Environment

_No response_

### Expected behavior

The multicommand author responsed and is not going to fix multi-command. He suggests that its a pytest issue and to create a fixture to fix.

So this needs to be reopened and addressed as it seems problematic to use multicommand as the core of the cli but be stuck on version 1.0 and the author disagreeing that its a mutlicommand problem.

Hi thanks for creating this issue. I'm unable to repro this behaviour.

I've tried the included repro script as well as creating actual packages with nested parsers, both approaches with both versions v2.0.1 and v1.0.0 - in all cases, calling multicommand.create_parser more than once succeeded without error.

I do believe that it's conceivably possible for this kind of exception to happen given that the error shown in the issue does indeed correspond to a [code path that will raise an exception](https://github.com/python/cpython/blame/65d2c51c10425dcfacc0a13810d58c41240d7ff9/Lib/argparse.py#L1257).

Regarding:

multicommand v2.0.1 appears to maintain global state or cache subparser registrations across create_parser() invocations, [...]
Yes, given that constructing argparse parsers in the usual way will almost certainly mutate the associated parser, and multicommand works by searching for parsers at the top of the module scope, in this way multicommand works almost entirely off of global state (the module graph and the top-level parser objects in the modules).

Workaround for pytest

That said, I do wonder how much of an issue this is in practice. It seems like testing would likely be the biggest culprit for tripping this exception, however, if one is using pytest (as indicated in your example) I believe a reasonable workaround would be to pull the parser creation out into a [session-scoped fixture](https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes) and then use this fixture (either directly or via other fixtures) for running whatever tests one needs related to the CLI.

### Actual behavior

Unit tests faile with v2.01 of multicommand

### Steps to reproduce

install multicommand 2.01 and run unit tests

jaelliot:
## Summary

Fixes #1332 — `multicommand==2.0.1` breaks pytest (and any repeated `create_parser()` call) with `ValueError: conflicting subparser`.

## Root Cause

`multicommand` 2.0.1 changed how it handles package-level `__init__.py` files. In 1.0.0, a `parser` object defined in a package `__init__.py` was silently ignored for directory-node packages. In 2.0.1, `_get_or_add_subparsers()` activates and **mutates** these parser objects in place — registering subcommand parsers (e.g., `clear`, `list`) as subparser choices.

`src/keri/cli/commands/escrow/__init__.py` was the only command package in keripy that defined a module-level `parser = argparse.ArgumentParser(...)`. On the first `create_parser()` call, multicommand 2.0.1 registers `clear` and `list` as subparsers on this object. On the **second** call (e.g., a different test function, or any production code that rebuilds the parser), it tries to register them again → `ValueError: conflicting subparser: clear`.

This is **not** pytest-specific — it reproduces in a plain Python script that calls `create_parser()` twice.

## Fix (3 files)

1. **`src/keri/cli/commands/escrow/__init__.py`** — Remove the dead `import argparse` and `parser = ArgumentParser(...)`. This parser was never used for routing (multicommand 1.0.0 ignored it) and was the sole outlier — all other command packages (`delegate`, `oobi`, `contacts`, `multisig`, `witness`, `watcher`) use imports-only `__init__.py` files.

2. **`setup.py`** — Bump `multicommand==1.0.0` → `multicommand==2.0.1`. Keeps the existing exact-pin convention for this dependency.

3. **`tests/app/cli/test_kli_commands.py`** — Add `test_create_parser_can_be_called_multiple_times()` regression test that calls `create_parser(commands)` twice, parsing escrow subcommands each time, asserting both produce working parsers.

## Validation

- 21/21 CLI tests pass under both `multicommand==1.0.0` (without escrow parser) and `multicommand==2.0.1`
- Reproduced the original failure with a standalone script, confirmed the fix eliminates it
- No code anywhere in keripy imports `escrow.parser` — removal is safe
- No other `__init__.py` in the CLI commands tree defines a package-level parser

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,7 +78,7 @@ setup(
                         'multidict>=6.7.0',
                         'ordered-set>=4.1.0',
                         'hio>=0.7.19',
-                        'multicommand==1.0.0',
+                        'multicommand==2.0.1',
                         'jsonschema>=4.26.0',
                         'falcon>=4.2.0',
                         'hjson>=3.1.0',
--- a/src/keri/cli/commands/escrow/__init__.py
+++ b/src/keri/cli/commands/escrow/__init__.py
@@ -5,11 +5,5 @@ keri.app.cli.commands.escrow Package
 
 """
 
-import argparse
-
 from .clear import clear
 from .list import escrows
-
-
-
-parser = argparse.ArgumentParser(description="A collection of escrow operations")
--- a/tests/app/cli/test_kli_commands.py
+++ b/tests/app/cli/test_kli_commands.py
@@ -312,3 +312,13 @@ def test_incept_and_rotate_opts(helpers, capsys):
     doers = args.handler(args)
 
     directing.runController(doers=doers)
+
+
+def test_create_parser_can_be_called_multiple_times():
+    parser = multicommand.create_parser(commands)
+    args = parser.parse_args(["escrow", "list", "--name", "test"])
+    assert args.handler is not None
+
+    parser = multicommand.create_parser(commands)
+    args = parser.parse_args(["escrow", "list", "--name", "test"])
+    assert args.handler is not None


jaelliot:
Closing this per the 2026-03-17 maintainers/dev call.

Kent is taking the command-line test path from here. The direction from the call was that command-line coverage should move to the scripts/bash side instead of staying in the keripy Python tests.

If that move does not happen quickly, the fallback discussed in the call was to mothball/remove the Python KLI tests so they stop running rather than keep carrying the multicommand-specific pytest path.

Referencing Kent here so the agreed follow-up is attached to this closure.

PR #1343, opened 2026-03-17
Multisig delegation Doer-level, Python integration tests

kentbull:
Tests for multisig delegation that test the delegation fixes and new delegator AES seal saving logic.

Adds the following:

## Tests

Tests:
- test_multisig_delegate: tests from 2 delegators to 2 delegates. Asserts on both delegator and delegate key state following successful delegation.
  - Uses all the test doers, context classes, context managers, and helper functions below.

### Test fixtures, Doers, and context managers
tests/app/integration/app_helpers.py: 

#### Test-only Doers:
  - `EscrowDoer`: handles Kevery and Counselor escrows
  - `MultisigInceptLeader`: Used by the **delegatee** initiator of delegated multisig inception for group AIDs with 2+ members.
  - `MultisigInceptFollower`: Used by the **delegatee** non-initiator group members of delegated group AIDs with 2+ members.
  - `MultisigDelegationApprover`: Used by the **delegator** group members to approve delegation requests automatically as they arrive. This completes delegation from the delegator's side.
  - `KeystateQueryDoer`: Used by the **delegatee** group members to discover the AES seal approving delegation from the delegator's KEL. This completes delegation from the delegatee's side. May also be used by the delegator to discover delegatee key state following approval.

#### Context classes:
- `ControllerContext`: basically what Doers the KLI spins up when you need to use a controller, like for `kli incept` or similar.
- `WitnessContext`: Doers needed to run a witness, and the ability to configure ports per witness context.

#### Context Managers
- `openWit`: context manager for a witness. Returns a `WitnessContext`.
- `openCtrlWited`: context manager for a controller that uses a witness as its mailbox and its single backer.

#### Helper Functions:
- HabHelpers
  - `generateOobi`: equivalent of `kli oobi generate` used to generate OOBIs for controllers and witnesses
  - `resolveOobi`: equivalent of `kli oobi resolve` used to resolve OOBIs between components
  - `collectWitnessReceipts`: used from the controller's perspective to wait for witness receipts to arrive for a given event
  - `hasDelegables`: used by the delegator to wait for delegation approval requests to arrive
  - `delegationSeal`: constructor function for a delegation seal
  - `clearSentCue`: used to clean up Postman cues to keep tests deterministic and isolated

### DIFF
--- a/tests/app/integration/app_helpers.py
+++ b/tests/app/integration/app_helpers.py
@@ -1,0 +1,1092 @@
+"""
+tests.app.app_helpers module
+
+Helpers for test setup including context managers for witnesses, controllers,
+and orchestration Doers for multisig and delegation workflows.
+"""
+import json
+from collections import deque
+from contextlib import contextmanager
+from dataclasses import dataclass, field
+from typing import List, Generator, Tuple
+
+from hio.base import Doer, doing, Doist
+from hio.help import decking
+
+from keri import core, kering
+from keri.app import habbing, delegating, grouping, oobiing
+from keri.app.agenting import WitnessReceiptor, Receiptor, WitnessInquisitor
+from keri.app.configing import Configer
+from keri.app.delegating import Anchorer
+from keri.app.forwarding import Poster
+from keri.app.habbing import openHab, HaberyDoer, Habery, Hab, openHby, GroupHab
+from keri.app.indirecting import MailboxDirector, setupWitness
+from keri.app.notifying import Notifier
+from keri.core import Salter, coring, serdering, indexing
+from keri.db import basing, dbing
+from keri.help import helping
+from keri.peer import exchanging
+from keri.peer.exchanging import Exchanger
+
+
+# =============================================================================
+# Data Classes for Structured Returns
+# =============================================================================
+
+@dataclass
+class EscrowDoer(doing.Doer):
+    """
+    Doer that processes escrows for both Habery's Kevery and Counselor.
+    This Doer is just a testing helper to speed up event processing in tests.
+
+    This fills a gap in the standard controller setup where:
+    - MailboxDirector.escrowDo processes mbx.kvy escrows (a separate Kevery for remote events)
+    - Counselor.escrowDo processes counselor escrows but with yield 0.5 delay, not great for tests that should run as fast as possible.
+    - Nothing processes hby.kvy escrows (the Habery's Kevery for local events)
+
+    This doer runs both processEscrows calls on every recur for faster test execution.
+    """
+
+    def __init__(self, hby: Habery, counselor: grouping.Counselor = None, **kwa):
+        super(EscrowDoer, self).__init__(**kwa)
+        self.hby = hby
+        self.counselor = counselor
+
+    def recur(self, tyme):
+        """Process escrows on every recur call for responsive tests."""
+        self.hby.kvy.processEscrows()
+        if self.counselor is not None:
+            self.counselor.processEscrows()
+        return False  # Keep running
+
+
+@dataclass
+class ControllerContext:
+    """Structured context for a KERI controller with all its components."""
+    hby: Habery
+    doers: List[Doer]
+    hbyDoer: HaberyDoer
+    anchorer: Anchorer
+    postman: Poster
+    exc: Exchanger
+    notifier: Notifier
+    mbx: MailboxDirector
+    witReceiptor: WitnessReceiptor
+    receiptor: Receiptor
+    witq: WitnessInquisitor = None
+    counselor: grouping.Counselor = None
+
+
+@dataclass
+class WitnessContext:
+    """Structured context for a KERI witness."""
+    hby: Habery
+    hab: Hab
+    doers: List[Doer]
+    oobi: str
+    pre: str = field(init=False)
+
+    def __post_init__(self):
+        self.pre = self.hab.pre
+
+
+# =============================================================================
+# Context Managers
+# =============================================================================
+
+@contextmanager
+def openWit(name: str = 'wan', tcpPort: int = 6632, httpPort: int = 6642,
+            salt: bytes = b'abcdefg0123456789') -> Generator[WitnessContext, None, None]:
+    """
+    Context manager for a KERI witness along with the Doers needed to run it.
+    Expects the Doers to be run by the caller.
+
+    Returns a WitnessContext with (Habery, Hab, witness Doers, witness controller OOBI URL)
+    """
+    saltQb64 = Salter(raw=salt).qb64
+    # Witness config - use temp=True to avoid filesystem permission issues in tests
+    witCfg = f"""{{
+          "dt": "2025-12-11T11:02:30.302010-07:00",
+          "{name}": {{
+            "dt": "2025-12-11T11:02:30.302010-07:00",
+            "curls": ["tcp://127.0.0.1:{tcpPort}/", "http://127.0.0.1:{httpPort}/"]}}}}"""
+    cf = Configer(name=name, temp=True, reopen=True, clear=False)
+    cf.put(json.loads(witCfg))
+    with (
+            openHab(salt=bytes(saltQb64, 'utf-8'), name=name, transferable=False, temp=True, cf=cf) as (hby, hab)
+    ):
+        oobi = f'http://127.0.0.1:{httpPort}/oobi/{hab.pre}/controller?name={name}&tag=witness'
+        hbyDoer = HaberyDoer(habery=hby)
+        doers: List[Doer] = [hbyDoer]
+        doers.extend(setupWitness(alias=name, hby=hby, tcpPort=tcpPort, httpPort=httpPort))
+        yield WitnessContext(hby=hby, hab=hab, doers=doers, oobi=oobi)
+
+
+@contextmanager
+def openCtrlWited(name: str = 'aceCtlrKS',
+                  salt: bytes = b'aaaaaaa0123456789') -> Generator[ControllerContext, None, None]:
+    """
+    Context manager for setting up a KERI controller that uses a witness as its mailbox and witness.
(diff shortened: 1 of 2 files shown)


SmithSamuelM (approved):
We need to readdress this in the maintainers or implementers call.  I read through the code briefly.  I am not as familiar with the group multisig delegation flow in the KLI so would like Phils concurrence.

tests/app/integration/test_multisig_delegation.py
```diff
@@ -0,0 +1,310 @@
...
+    1. A delegator multisig (dgt) formed by two single-sig participants (dgt1, dgt2)
+    2. Two delegate participants (del1, del2) who create a delegated multisig (del)
+    3. The delegates having OOBId with the delegator multisig
+    4. Delegation approvals from both delegator single-sig AID participants
+    5. Keystate queries from delegates to discover the delegation approval seal
+    6. Generation of an OOBI for the multisig delegate
+    7. Resolution of the multisig delegate OOBI by the delegator
+    8. Verifications by the delegator about delegate state
+    """
+    doist = Doist(limit=0.0, tock=0.03125, real=True)
```
dhh1128, line 29: **[blocker] Unbounded wait loop → CI hang.** `Doist(limit=0.0)` plus bare `while` loops here (and the equivalents in `app_helpers.py`) have no iteration cap and no wall-clock deadline. If the protocol ever fails to converge, this hangs forever instead of failing with a useful message. Please add an iteration/deadline bound so a regression surfaces as a test failure, not a hung CI job.

tests/app/integration/test_multisig_delegation.py
```diff
@@ -0,0 +1,310 @@
...
+from tests.app.integration.app_helpers import (
+    openWit, openCtrlWited, HabHelpers,
+    MultisigInceptLeader, MultisigInceptFollower,
+    MultisigDelegationApprover, KeystateQueryDoer
+)
+
+from keri.core import coring
+
+
+def test_multisig_delegate():
```
dhh1128, line 15: Two testability notes, both fine to defer to a follow-up: (1) only the happy path is exercised — the follower-side conditional branches and the `auto=False` rejection path in `MultisigDelegationApprover`, plus DRT delegation, are currently uncovered; (2) the whole workflow is one ~310-line function, so a failure doesn't tell you which stage broke and no stage can run in isolation. Splitting into staged tests/fixtures would help as this suite grows.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+    def __post_init__(self):
+        self.pre = self.hab.pre
+
+
+# =============================================================================
+# Context Managers
+# =============================================================================
+
+@contextmanager
+def openWit(name: str = 'wan', tcpPort: int = 6632, httpPort: int = 6642,
```
dhh1128, line 98: **[blocker] Hardcoded ports 6632/6642 are flaky under parallel execution.** This repo now runs `pytest-xdist` (`-n auto`); fixed ports collide across workers (or on a duplicate invocation) and fail nondeterministically with "address already in use" — with no protocol diagnostic. Please allocate ephemeral ports (bind to `0` and read back, or a per-worker offset from `PYTEST_XDIST_WORKER`).

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+from keri.help import helping
+from keri.peer import exchanging
+from keri.peer.exchanging import Exchanger
+
+
+# =============================================================================
+# Data Classes for Structured Returns
+# =============================================================================
+
+@dataclass
```
dhh1128, line 36: `@dataclass` on `EscrowDoer` — which has no annotated fields and an explicit `__init__` — generates an `__eq__` that makes *any* two `EscrowDoer` instances compare equal, and a `__repr__` with no field values. It's a behavioral Doer, not a data holder; dropping the decorator avoids the surprising equality semantics.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+        self.rmids = rmids
+        self.group = group
+        self.isith = isith
+        self.nsith = nsith
+        self.toad = toad
+        self.wits = wits
+        self.delpre = delpre
+        self.postman = postman
+        self.counselor = counselor
+        self.witReceiptor = witReceiptor
```
dhh1128, line 341: `self.witReceiptor` is assigned here (and in the other two Doer classes) but never read anywhere — a write-only/dead field. It reads like receipt collection happens via the Doer when it doesn't, which will cost a future reader time. Either use it or drop it.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+        self.ghab: GroupHab = None
+        self.cues = decking.Deck()
+        self.done = False
+        self.pending_sends = []  # Track SAIDs of messages waiting for delivery confirmation
```
dhh1128, line 345: Leader uses snake_case (`pending_sends`, `counselor_started`) while Follower uses camelCase (`pendingSends`, `counselorStarted`) for the same concepts — looks like copy-paste residue. If any method is ever shared/lifted between the two, this silently references the wrong attribute at runtime. Pick one convention.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+            return False
+
+        # Step 3: Start Counselor (once, after sends complete)
+        if not self.counselorStarted:
+            self._startCounselor()
+            return False
+
+        # Step 4: Wait for Counselor to complete
+        if self._isCounselorComplete():
+            print(f"[Follower {self.mhab.pre[:8]}] Multisig inception complete for {self.ghab.pre}", flush=True)
```
dhh1128, line 510: These `print()` calls (≈9 across the helper) emit on every run and tend to bury the actual failure output. The project already has logging infrastructure — prefer that (or drop them).

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+        Start Counselor coordination for delegations that are ready.
+
+        The Counselor collects signatures from all multisig participants
+        and marks the event complete when threshold is met.
+        """
+        for key in list(self.readyForCounselor.keys()):
+            info = self.readyForCounselor[key]
+            prefixer = coring.Prefixer(qb64=self.ghab.pre)
+            number = core.Number(num=info['ixn_sn'])
+            diger = coring.Saider(qb64=info['ixn_said'])
```
dhh1128, line 744: The IXN digest is built as a `Saider` but `Counselor.start()` expects a `Diger`. It works today only because both encode to the same qb64 for a SAID — it'll break the moment `grouping.py` adds a code-family check, and it sets a misleading example for integrators copying this. Construct a `Diger` here.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
...
+        if delegate_info is None:
+            return False
+
+        delegate_pre, delegate_sn, anchor_data = delegate_info
+
+        # Sign and coordinate
+        self._signAndCoordinateAnchor(delegate_pre, delegate_sn, anchor_data)
+        return True
+
+    def _extractDelegateInfoFromAnchor(self, ixn_data: dict) -> Tuple[str, str, list]:
```
dhh1128, line 985: Not blocking — flagging for hygiene. `_extractDelegateInfoFromAnchor` trusts `seal['i']`/`seal['s']` and the embedded KEDs straight off a cloned EXN and pins `db.aess` seals before `processEscrowDelegables`, bypassing normal Kevery validation. That's acceptable in an in-process test scaffold with honest peers, but please keep this pattern out of any real approval path — worth a comment here saying so.

tests/app/integration/app_helpers.py
```diff
@@ -0,0 +1,1092 @@
+"""
+tests.app.app_helpers module
```
dhh1128, line 2: Nit: the module docstring cites `tests.app.app_helpers` but the file lives at `tests.app.integration.app_helpers` (throws off grep), and `indexing` is imported but unused.

dhh1128 (commented):
Thanks for adding Doer-level integration coverage for multisig delegation, @kentbull — this is useful, and the happy path is easy to follow. I did a focused review of the three files. No production code is touched, so everything below is about the tests. I've left inline notes and am summarizing here.

Two of these are blockers, both about CI safety:

- Unbounded loops can hang CI. `Doist(limit=0.0)` plus bare `while` loops with no iteration cap or wall-clock deadline means any protocol regression hangs the test indefinitely instead of failing with a diagnostic (`test_multisig_delegation.py:29`, `app_helpers.py:232`).
- Hardcoded ports 6632/6642 will flake in parallel. They collide under `pytest-xdist` (`-n auto`, now used here) or any concurrent run, giving nondeterministic "address already in use" (`app_helpers.py:98`).

These compound with a third thing I couldn't inline because it isn't in the diff: `tests/app/integration/` isn't excluded from the default CI job, so an unbounded, real-socket test can hang the whole matrix. It's worth gating these behind a marker, or wiring them into the existing Windows-shard exclusion, in this same PR.

A few smaller cleanups are noted inline: a `Saider` where `Counselor.start()` wants a `Diger`; nine `print()` calls that should use the logging infra; a dead `self.witReceiptor` field on all three Doers; snake_case/camelCase divergence between Leader and Follower; a spurious `@dataclass` on the behavioral `EscrowDoer`; a stale docstring path and an unused import.

The rest is optional and fine to defer: only the happy path is exercised — no error, partial-participation, `auto=False`-rejection, or DRT branches — and the single 310-line test function makes stage-level failure hard to localize. I'm just flagging the coverage gap.

One thing I looked at and am not worried about: the delegation-approval helper reads EXN payload fields before pinning DB seals. That would matter in production code, but this is a test-only scaffold against honest in-process peers — I mention it only so the pattern doesn't get copied into real approval logic later.

(These came out of an adversarial multi-persona pass — security, architecture, testability, maintainability — that I went through and vetted.)

PR #1344, opened 2026-03-17
KLI Requester Path Stripping and Receiptor ksn() Method

arilieb:
## This PR Includes
- ksn method for agenting.py Receiptor class which sends a ksn query to the witness and waits for the reply to be processed through the rpy/ksn pipeline into knas/ksns
- Stripping rightmost slash from client.requester.path when used as path parameter in client.request for both `kli mailbox add` and `kli witness authenticate` commands.
  - NOTE: This change enables scripts from new OSS witness-hk repo https://github.com/keri-foundation/witness-hk/pull/1

### DIFF
--- a/src/keri/app/cli/commands/mailbox/add.py
+++ b/src/keri/app/cli/commands/mailbox/add.py
@@ -110,7 +110,7 @@ class AddDoer(doing.DoDoer):
 
         client.request(
             method="POST",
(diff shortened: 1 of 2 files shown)


SmithSamuelM:
@arilieb Are you waiting for Phil to review this?

PR #1352, opened 2026-03-19
IoSet methods

evanja57:
- Full implementation of all IoSet methods in webdber.py, modeled to match the lmdber api

### DIFF
--- a/src/keri/db/webdbing.py
+++ b/src/keri/db/webdbing.py
@@ -141,6 +141,8 @@ def isNonStringIterable(obj):
 
 _RECORDS_KEY = "__records__"
 _META_KEY = "__meta__"
+_META_STORE = "__meta__"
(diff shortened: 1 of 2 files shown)


src/keri/db/webdbing.py
```diff
@@ -815,6 +874,1000 @@ def cntAll(self, db: SubDb) -> int:
...
+        currKey = None
+
+        # Iterate forward through the DB
+        for iokey in items.irange(startKey, None):
+            # Split into (apparent_key, ordinal)
+            try:
+                apparent, ion = unsuffix(iokey, sep=sep)
+            except Exception:
+                # malformed key, skip
+                continue
```
evanja57: dont need to wrap this in a try/except--prolly good to fail loudly if unsuffix errors

Issue #1317, opened 2026-03-11
Bug: witness of delegatee does not store delegator event seal


lenkan:
## TLDR

The witness of delegatee never does db.setAes() to store the authorizer event seal (delegator/issuer) for the `dip` event. Symptoms are

- OOBI response for delegate does not include the SealSourceCouple attachment.
- `kli kevers` shows delegate as **Not Anchored**

## Background

I ran into this while working with a multisig delegated identifier after a rotation. Newly rotated-in members were seeing the group (delegate) identifier as **"Not anchored"**. When I dug into the witness responses, I found that witnesses were returning data without the `SealSourceCouple` attachment that identifies the delegator’s anchor seal. So the delegate’s witness(es) had the receipt/threshold info but were missing the seal that links the delegate to its delegator, which likely drives the incorrect "Not anchored" status. I have created gist as a minimal reproduction of that behaviour (single delegator, single delegate, each with one witness, no multisig/rotation).

## Description

When inspecting delegate kevers via `kli kevers`, the **delegate’s witness** reports the delegator as **"Not Anchored"**, while the **delegator** and **delegate** (controller) both correctly report **"Anchored"** for the same delegated identifier. Witness count, receipts, and threshold are the same in all three places; only the delegate witness’s “anchored” status is wrong.

So the bug is: **delegate witness state (or response) is missing or not using the delegator’s anchor seal (e.g. SealSourceCouple), leading to a false "Not Anchored".**

## Minimal reproduction

1. Clone the gist and run the script:

   ```bash
   git clone https://gist.github.com/c03db9084631695fe8ab522c25a2d431.git delegation-problem
   cd delegation-problem
   ./delegation-problem.sh
   ```

2. At the end, the script runs `kli kevers` for the delegate AID from three contexts: **delegator**, **delegate**, and **delegate witness**.

## Expected vs actual

- **Expected:** All three (delegator, delegate, delegate witness) show the delegator as **✔ Anchored** for the delegated identifier.
- **Actual:** Delegator and delegate show **✔ Anchored**; the **delegate witness** shows **✘ Not Anchored** for the same delegator/delegate pair and receipt data.

## Example output

```
--------------------------------
Delegate kevers from delegator → Delegator:  EAhA4... ✔ Anchored
--------------------------------
--------------------------------
Delegate kevers from delegate  → Delegator:  EAhA4... ✔ Anchored
--------------------------------
--------------------------------
[...]

PR #1402, opened 2026-04-22
fix: resolve missing AES on delegate's witness after delegation

krillero:
Witnesses accept delegated events (dip/drt) before the delegator has anchored them, to avoid a deadlock in the delegation flow. However, the authorizer event seal (AES) was never back-filled once the delegator's anchoring event arrived, leaving the witness permanently showing "Not anchored" for the delegated identifier.

Introduce a pending witness delegation escrow (pwde) that tracks delegated events accepted by a witness without an AES. A new processEscrowWitnessAnchors method, called from processEscrows, periodically checks whether the delegator's KEL now contains the anchoring seal and stores the AES when it does.

Fixes weboftrust/keripy#1317

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -3151,10 +3151,9 @@ class Kever:
             and not self.locallyWitnessed(wits=wits) and seqner and saider):
             couple = seqner.qb64b + saider.qb64b
             self.db.setAes(dgkey, couple)  # authorizer (delegator/issuer) event seal
-
-        #if seqner and saider:
-            #couple = seqner.qb64b + saider.qb64b
-            #self.db.setAes(dgkey, couple)  # authorizer (delegator/issuer) event seal
+        elif (self.delpre and not serder.ilk == Ilks.ixn
+              and self.locallyWitnessed(wits=wits)):
+            self.db.pwde.addOn(keys=serder.pre, on=serder.sn, val=serder.said)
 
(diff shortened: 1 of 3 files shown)


lenkan:
I also tried "repairing" a witness using this escrow. So, after completing the delegation ceremony on the broken version, where the witness then is missing the anchor event seal. I can switch to this branch, and then do

```
kli export --name delegate --alias delegate | curl -X PUT http://localhost:5643
```

The escrow will run on the witness, and the anchoring event seal is added.

kentbull:
After reading through this I like it and believe it is half of the solution. The half we have in the PR is the witness side where it can accept the delegator's witness approval.

What I believe the other half should be would be a seal propagation Doer and an escrow in the witness that would put the local delegate's dip  event in a seal propagation escrow and after the delegate discovers the delegator's approving seal via a keystate query then the seal propagation doer would look up the current witness set for the delegate and push the seal to each witness, effectively doing `kli export | http PUT witness_URL` yet with only the CESR stream slice containing the delegator seal.

edeykholt:
conclusion of 5/5 meeting was to accept this on 1.3 branch if Phil approves. (or a sidecar watcher on 1.3).  Reasonable approach on 1.2, given there isn't a local watcher (GLEIF).

kentbull:
@krillero please reopen this PR to v1.2.14, the new release branch.

lenkan:
I opened https://github.com/WebOfTrust/keripy/pull/1498

PR #1409, opened 2026-05-12
passing bytes to cbor loads

evanja57:
- cbor2 updated to version 6.0.1 (rust rewrite) and loads() now does not accept a bytearray
- 6.0.2 will support bytearrays again ([line 195](https://github.com/agronholm/cbor2/blob/master/rust/lib.rs)) but currently this is breaking tests/core/test_eventing.py::test_direct_mode_cbor_mgp
- so for now we can just pass bytes() to cbor.loads()


evanja57:
new cbor release fixes this

PR #1423, opened 2026-05-25
fix(agenting): route HTTPMessenger response bodies through hab.psr

### LINKED ISSUE #1422: HTTPMessenger lacks Parser wiring (unlike TCPMessenger) — WitnessReceiptor over HTTP silently drops receipts
### Version

`keri==2.0.0.dev6` (and likely all 2.0.0.dev* releases)

### Environment

HTTP witness deployment — any downstream consumer of `keri.app.agenting.WitnessReceiptor` that talks to a witness over HTTP rather than TCP.

### Summary

`HTTPMessenger` in `keri/app/agenting.py` has no `Parser` wired, so HTTP witness response bodies are received but never parsed through `Kevery`. Consumers of `WitnessReceiptor` (which builds `HTTPMessenger` under the hood when the witness has an HTTP endpoint) therefore never see witness receipts persisted into `baser.wigs`, and their busy-wait loops on `len(wigers) == len(wits)` run forever.

The fix is structural and trivial — `TCPMessenger`, the HTTP sibling's counterpart, already does this correctly.

### Expected behavior

`HTTPMessenger.responseDo` should parse received response bodies as CESR and route events (including witness receipts) through `Kevery`, mirroring what `TCPMessenger` does. `WitnessReceiptor`-driven flows should then see `wig` entries land in `baser.wigs` and exit their wait loops.

### Actual behavior

`HTTPMessenger.responseDo` only appends raw response records to `self.sent` (a `decking.Deck`):

```python
# keri/app/agenting.py (around line 893)
def responseDo(self, tymth=None, tock=0.0, **kwa):
    """Doer loop that processes HTTP responses from the client and adds them into `sent` cues."""
    self.wind(tymth)
    self.tock = tock
    _ = (yield self.tock)

    while True:
        while self.client.responses:
            rep = self.client.respond()
            self.sent.append(rep)   # ← responses end here
            yield
        yield
```

`HTTPMessenger.__init__` sets `self.parser = None` and never instantiates a `Parser`. Nothing downstream consumes the response bodies in `self.sent` to feed them to `Kevery`. So:

- `WitnessReceiptor.receiptDo`'s inner `while True: ... if len(wigers) == len(wits): break` loop never exits.
- `baser.wigs[(pre, said)]` remains empty even after the witness returns a well-formed `rct` event + `-CAB<eid><signature>` attachment in the HTTP response body.
- Downstream consumers (e.g., wallets calling `WitnessReceiptor` for inception receipts) hang indefinitely.

Compare to `TCPMessenger.__init__` (lines ~731):

```python
self.parser = parsing.Parser(ims=client.rxbs,
                             framed=True,
                             kvy=self.kevery,
                             version=Vrsn_1_0)
```

```python
def msgDo(self, tymth=None, tock=0.0, **opts):
    """Doer loop that parses inbound TCP messages into the Kevery."""
    yield from self.parser.parsator(local=True)
```

`HTTPMessenger` has neither.

### Steps to reproduce

```python
from urllib.parse import urlparse
from hio.base import doing
from hio.core import http
from keri.db import basing
from keri.app.agenting import streamCESRRequests

# Assumes:
#   - WITNESS = a non-transferable witness AID whose KEL is already resolved
#     into baser.kevers
#   - INCEPTION_CESR = bytes of a signed icp event for some PRE with
#     wits=[WITNESS], toad=1
WITNESS = "BNbRf..."
URL = "https://your.witness.example/"

baser = basing.Baser(name="probe", temp=True, reopen=True)
up = urlparse(URL)
client = http.clienting.Client(scheme=up.scheme, hostname=up.hostname, port=up.port)
clientDoer = http.clienting.ClientDoer(client=client)

streamCESRRequests(client=client, ims=bytearray(INCEPTION_CESR), dest=WITNESS)
doist = doing.Doist(limit=10.0, tock=0.03125, real=True, doers=[clientDoer])
doist.do(doers=[clientDoer])

# Drain
while client.responses:
    rep = client.respond()
    print(rep.status, len(rep.body))    # 200, 281  — the witness *did* send a receipt
```

The receipt body comes back. But if you'd driven this through `WitnessReceiptor` instead, `baser.wigs[(PRE, said)]` would still be empty afterward — `WitnessReceiptor`'s loop has no path to consume the response.

### Proposed fix (minimal)

Add a `Parser` to `HTTPMessenger`, mirroring `TCPMessenger`:

```python
class HTTPMessenger(doing.DoDoer):
    def __init__(self, hab, wit, url, msgs=None, sent=None, doers=None, auth=None, **kwa):
        ...
        self.client = http.clienting.Client(scheme=up.scheme, hostname=up.hostname, port=up.port)
        clientDoer = http.clienting.ClientDoer(client=self.client)

        # NEW: wire a Kevery + Parser like TCPMessenger does
        self.kevery = eventing.Kevery(db=self.hab.db, **kwa)
        self.parser = parsing.Parser(ims=self.client.rxbs,
                                     framed=True,
                                     kvy=self.kevery,
                                     version=Vrsn_1_0)

        doers.extend([clientDoer, doing.doify(self._parseDo)])
        ...

    def _parseDo(self, tymth=None, tock=0.0, **opts):
        yield from self.parser.parsator(local=True)
```

The `responseDo`'s existing behavior (appending to `self.sent` for callers that want to inspect raw responses) can remain — the new parser sits alongside it and routes events to `Kevery` independently.

### Alternative — deprecate `WitnessReceiptor` outright

The `Receiptor` class (in the same file, around line 28) is structurally newer and has the parsing wired correctly. If the keripy team is moving consumers to `Receiptor` as the canonical path, adding a `DeprecationWarning` to `WitnessReceiptor.__init__` with a pointer to `Receiptor` would prevent other downstreams from falling into the same trap. (Downstreams like ours have already migrated; this issue is filed primarily so the trap closes for everyone else.)

### Related

- keri-foundation/locksmith#77 (Locksmith-side fix migrating to `Receiptor`, just filed and PR'd as a workaround that avoids relying on `WitnessReceiptor`/`HTTPMessenger` until this gets fixed upstream)
- WebOfTrust/keripy#814 (separate bug in `Receiptor` itself — unresponsive-witness hang)

### PR filed

**Update (2026-05-24):** opened **WebOfTrust/keripy#1423** which wires `HTTPMessenger.responseDo` to feed received bodies through `self.hab.psr.parseOne(...)`, mirroring the canonical pattern in `Receiptor.receiptDo` (line 116). One subtlety vs. the proposed fix above: hio's `client.rxbs` is the raw socket buffer (HTTP frames), not where parsed response bodies live — those land in `client.responses[].body`. So the PR uses the `Receiptor`-style inline `parseOne` rather than a second `Kevery` + `parsing.Parser(ims=client.rxbs, ...)`. Existing `self.sent.append(rep)` preserved so callers inspecting raw responses are unaffected.

Includes a regression test (`tests/app/test_agenting.py::test_http_messenger_routes_response_through_parser`) confirmed to fail on `main` HEAD and pass after the patch; full `tests/app/` stays green (126/126).

seriouscoderone:
## Summary

Wires `HTTPMessenger.responseDo` to feed received HTTP response bodies through the hab's parser (`self.hab.psr.parseOne`) alongside the existing `self.sent` append. Mirrors the canonical pattern in `Receiptor.receiptDo` and closes the structural gap with `TCPMessenger`, which already parses inbound CESR via its own `Kevery`.

Without this, witness receipts returned in HTTP response bodies are accepted on the wire but never land in `baser.wigs`. Any caller using `WitnessReceiptor` against an HTTP witness then loops indefinitely on `len(wigers) == len(wits)`.

Closes #1422.

## Why

Side-by-side on `main`, `src/keri/app/agenting.py`:

- `TCPMessenger` (line 692): builds a `Kevery` + `parsing.Parser(ims=client.rxbs, framed=True, kvy=...)`, runs `parsator(local=True)` in `msgDo`.
- `Receiptor.receiptDo` (line 116): `hab.psr.parseOne(bytearray(rep.body))` directly on each 200 response. Known-working in production.
- `HTTPMessenger.__init__` (line 856): `self.parser = None`. `responseDo`: only `self.sent.append(rep)`. Nothing routes to `Kevery`.

The fix uses the `Receiptor` pattern (inline `hab.psr.parseOne`) rather than the issue's proposed `parsing.Parser(ims=client.rxbs, ...)` because hio's `client.rxbs` is the raw socket buffer — HTTP-parsed response bodies live in `client.responses[].body`, not `client.rxbs`. Using `hab.psr` also avoids constructing a second `Kevery` that could escrow-desync against the hab's own.

## Changes

`src/keri/app/agenting.py` — `HTTPMessenger.responseDo` now:

```python
while True:
    while self.client.responses:
        rep = self.client.respond()
        self.sent.append(rep)
        if rep.status == 200 and rep.body:
            try:
                self.hab.psr.parseOne(ims=bytearray(rep.body))
            except Exception as ex:
                logger.warning("HTTPMessenger response parse failed: %s", ex)
        yield
    yield
```

- `self.sent.append(rep)` preserved so existing callers inspecting raw responses still get them.
- Parse failure is logged via the module's existing `logger` (`ogler.getLogger()`) and the doer keeps running; matches the resilience posture elsewhere in the file.
- `self.parser = None` is intentionally not removed — callers that wanted to hang a custom parser off it can still do so; nothing in the diff touches that.

`tests/app/test_agenting.py` — adds `test_http_messenger_routes_response_through_parser`:

- Two separate `Habery` instances (witness and controller) so the receipt crosses a real "process" boundary, mirroring production.
- Controller's hby learns the witness's KEL, then incepts with `wits=[wit.pre]`. Witness sees the controller's inception, then produces a real `wit.witness(ctrl.kever.serder)` receipt.
- HTTPMessenger built with a stub `client` exposing only `.responses` and `.respond()` — drives `responseDo` as a generator (two `next()` calls: prime, then process one response).
- Asserts `ctrl_hby.db.wigs[(ctrl.pre, ctrl.kever.serder.saidb)]` has 1 entry and `len(msgr.sent) == 1` (no regression of the raw-response path).

## Verification

- **Regression guard validated.** Test confirmed to fail on `origin/main` HEAD (the bug repros) and pass with this patch.
- `pytest tests/app/test_agenting.py -v` — **6/6 pass** including the pre-existing `test_withness_receiptor` end-to-end TCP test (no TCP regression).
- `pytest tests/app/` — **126/126 pass** in ~134s.

## Not in this PR

- `WitnessReceiptor` deprecation — the issue raises it as an alternative if maintainers are moving consumers to `Receiptor` long-term. Happy to add a follow-up PR with a `DeprecationWarning` and a pointer if you'd like.
- The Locksmith side has already migrated to `Receiptor` (out-of-tree). This PR is filed primarily so the next consumer that walks into the same trap doesn't.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -895,7 +895,21 @@ class HTTPMessenger(doing.DoDoer):
             yield self.tock
 
     def responseDo(self, tymth=None, tock=0.0, **kwa):
-        """Doer loop that processes HTTP responses from the client and adds them into `sent` cues."""
+        """Doer loop that processes HTTP responses from the client.
+
+        Appends each response to ``self.sent`` (so callers inspecting raw
+        responses still work) and, when the response carries a CESR body,
+        routes it through ``self.hab.psr`` so witness receipts land in
+        ``hab.db.wigs``.
+
+        Without this routing, callers that drive an HTTPMessenger via
+        ``WitnessReceiptor`` (witnesses reached over HTTP) never see the
+        receipts the witness actually returned, and their busy-wait loop
+        on ``len(wigers) == len(wits)`` never exits. Mirrors the canonical
+        pattern in ``Receiptor.receiptDo`` and closes the structural gap
+        with ``TCPMessenger`` which already parses inbound CESR via its
+        own ``Kevery``.
+        """
         self.wind(tymth)
         self.tock = tock
         _ = (yield self.tock)
@@ -904,6 +918,11 @@ class HTTPMessenger(doing.DoDoer):
             while self.client.responses:
                 rep = self.client.respond()
                 self.sent.append(rep)
+                if rep.status == 200 and rep.body:
+                    try:
+                        self.hab.psr.parseOne(ims=bytearray(rep.body))
+                    except Exception as ex:
+                        logger.warning("HTTPMessenger response parse failed: %s", ex)
                 yield
             yield
 
--- a/tests/app/test_agenting.py
+++ b/tests/app/test_agenting.py
@@ -237,6 +237,86 @@ def test_witness_inquisitor(mockHelpingNowUTC, seeder):
         doist.exit()
 
 
+def test_http_messenger_routes_response_through_parser():
+    """HTTPMessenger.responseDo routes CESR response bodies through the
+    hab's parser so witness receipts land in db.wigs.
+
+    Regression guard for issue #1422: HTTPMessenger previously only
+    appended raw responses to ``self.sent`` and never fed them through
+    Kevery. Callers driving an HTTPMessenger via WitnessReceiptor would
+    receive the receipt body in ``self.sent`` but ``baser.wigs`` stayed
+    empty — busy-wait loops on ``len(wigers) == len(wits)`` would never
+    exit. This mirrors what TCPMessenger and Receiptor already do.
+    """
+    from types import SimpleNamespace
+    from hio.help import decking
+    from keri.app.agenting import HTTPMessenger
+
+    # Witness and controller live in separate Haberies so the receipt
+    # crosses a real "process" boundary — the same shape as production
+    # where the witness is a remote service.
+    with openHby(name="wit-hby", salt=Salter(raw=b'wit-hby-issue1422-').qb64) as wit_hby, \
+            openHby(name="ctrl-hby", salt=Salter(raw=b'ctrl-hby-issue1422').qb64) as ctrl_hby:
+        wit = wit_hby.makeHab(name="wit", transferable=False, isith="1",
+                              icount=1, ncount=0, nsith="0")
+
+        # Controller's hby learns the witness's KEL via OOBI in real life;
+        # here we just parse the witness's inception into ctrl_hby.
+        ctrl_hby.psr.parse(ims=bytearray(wit.msgOwnInception()))
+        ctrl = ctrl_hby.makeHab(name="ctrl", transferable=True, isith="1",
+                                icount=1, ncount=1, nsith="1",
+                                toad=1, wits=[wit.pre])
+
+        # Witness side must have seen ctrl's inception to build a wig.
+        wit_hby.psr.parse(ims=bytearray(ctrl.msgOwnInception()))
+        rct_bytes = wit.witness(serder=wit_hby.kevers[ctrl.pre].serder)
+
+        # Sanity: receipt was built and ctrl_hby has no wig yet for it.
+        dgkey = (ctrl.pre.encode(), ctrl.kever.serder.saidb)
+        assert len(ctrl_hby.db.wigs.get(keys=dgkey)) == 0, (
+            "ctrl_hby unexpectedly has a wig before the messenger runs"
+        )
+
+        # Build an HTTPMessenger but replace its live client with a
+        # test-double exposing only the responseDo surface we exercise:
+        # .responses (a Deck) and .respond() (pop one).
+        msgr = HTTPMessenger(hab=ctrl, wit=wit.pre, url="http://localhost:9/")
+
+        fake_rep = SimpleNamespace(status=200, body=bytes(rct_bytes))
+        responses = decking.Deck([fake_rep])
+
+        class _StubClient:
+            def __init__(self, deck):
+                self.responses = deck
+            def respond(self):
+                return self.responses.popleft()
+
+        msgr.client = _StubClient(responses)
+
+        # Drive responseDo as a generator. First next() primes past the
+        # initial `_ = (yield self.tock)`. Second next() enters the inner
+        # `while self.client.responses` loop, pops fake_rep, appends to
+        # self.sent, runs hab.psr.parseOne(rep.body), then yields.
+        gen = msgr.responseDo(tymth=lambda: 0.0, tock=0.0)
+        next(gen)
+        next(gen)
+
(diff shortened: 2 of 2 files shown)


pfeairheller (changes requested):
This does not close a "structural gap" with TCPMessenger.  This uses the parser of the Hab that is not in its control.  TCPMessenger creates its own Kevery and Parser and uses those for parsing.  
In addition, this significantly changes the contract between HTTPMessenger and all of its callers with out any identification of the callers and their expectations of the contract.

At a minimum, this needs to create and use its own parser and then further analysis needs to be performed of all uses of this class to see how they will be affected.

seriouscoderone:
Closing — the premise was a misdiagnosis.

`HTTPMessenger` is a **sender** and correctly has no parser: its `responseDo` only counts responses for `idle`, and the witness `/` endpoint returns `204` for key events anyway. Receipts over HTTP arrive via the synchronous `/receipts` endpoint (`agenting.Receiptor`) or the mailbox SSE poll — **not** the event-POST response. `WitnessReceiptor` hangs over HTTP because it assumes the direct-mode (TCP) push-back model, which HTTP/serverless witnesses cannot provide. Wiring a parser into `HTTPMessenger` (whether via `hab.psr` or its own `Kevery`) fixes nothing, because there is nothing to parse.

The correct fix is client-side: use `Receiptor`/mailbox-poll instead of `WitnessReceiptor` over HTTP. Full writeup on #1422. Thanks @pfeairheller for the review.

PR #1433, opened 2026-06-08
adding kram to habbing tests and to parser

evanja57:
## parsing.py
- for rpy, logs, ksn, mbx, tels, tsn, and exn routes
  - if kevery has a kramer, send the message through kvy.processMsg to allow kram to run

## test_habbing_v2.py
- configure and enable KRAM 
### test_postman_endsfor_v2
- make sure kram does not break v2 endpoint reply processing and replies are not bypassing KRAM
### test_cues_v2
- make sure rpy and qry messages from hby.processCues work through KRAM and makes sure replayed queries are detected

### DIFF
--- a/src/keri/core/parsing.py
+++ b/src/keri/core/parsing.py
@@ -855,6 +855,7 @@ class Parser:
                 if group counter extract and discard but keep track of count
                 so if error while processing attachments then only need to flush
                 attachment count not full stream.
+                dispatch msg (route to kvy.processMsg if kvy has a kramer, otherwise direct)
 
 
         """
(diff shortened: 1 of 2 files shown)


src/keri/core/parsing.py
```diff
@@ -1205,7 +1206,13 @@ def msgParsator(self, ims=None, framed=True, piped=False,
                                                  f"to reply msg = {serder.pretty()}.")
 
                 try:
-                    rvy.processReply(**exts)
+                    if kvy is not None and getattr(kvy, "kramer", None) is not None:
```
ryan-hansen: Let's get rid of the getattr and just check kvy.kramer directly:
`if kvy is not None and kvy.kramer is not None:`

kramer is already added to kvy in __init__, defaulted to None, so should be safe.  Same thing for all four if blocks.

ryan-hansen (changes requested):

PR #1436, opened 2026-06-12
eo delegated inception approval

m00sey:


pfeairheller (approved):

kentbull:
We talked about this in the maintainer call today. Sam said that since Phil approves then he approves.

kentbull (approved):

dhh1128 (commented):
Not requesting changes — m00sey's already approved this, and the shape is right: a coordinated group rotation is the only path an EstablishmentOnly delegator has to approve a delegated inception, and this unblocks a vLEI-root-with-GARs topology that couldn't approve one at all before.

One thing worth a follow-up, maybe not in this PR. The new group-rotation branch in `delegate/confirm.py` (L139-148) has no automated coverage — `setup-root-gars.sh` is a shell demo, not a test. What makes me raise it is the failure mode when its unstated precondition isn't met. The branch runs `BaseHab.rotate`, whose `satisfy()` firewall (habbing.py:1235-1241) fails closed when a member hasn't locally pre-rotated and synced first, and it does that by raising a `ValidationError` that `confirmDo` doesn't catch — so a member who forgot to `kli rotate` locally gets an uncaught crash whose message never names the actual remedy (each member rotates locally, the delegator queries to sync, then confirm). A test over both the pre-rotated accept path and the not-pre-rotated reject path, plus a line in the PR or CLI help spelling out the pre-rotation step, would turn an opaque crash into something an operator can act on.

Smaller, and inherent rather than a defect: because approval now rides on a rotation, each one spends a fresh coordinated pre-rotation across all members, with no batching of several pending approvals into a single establishment event. That's just how EstablishmentOnly works, so it's fine — but a governance root delegating to many entities ends up with a KEL where routine approvals and key-hygiene rotations look alike, which an auditor has to disentangle by reading anchored seals. A sentence noting the tradeoff would keep it from being a surprise later.

There's no PR description, so I'm reading intent off the diff; correct me where I've got it wrong.

Issue #1175, opened 2026-02-10
weboftrust keri docker image


RangoW:
### Version

1.2.7

### Expected behavior

```
$ docker run --pull=never -it  weboftrust/keri:1.2.6 version
Library version: 1.2.6
```

### Actual behavior

```
$ docker run --pull=never -it  weboftrust/keri:1.2.7 version
/keripy/src/keri/kering.py:47: SyntaxWarning: invalid escape sequence '\.'
  VEREX2 = b'(?P<proto2>[A-Z]{4})(?P<major2>[0-9A-Za-z_-])(?P<minor2>[0-9A-Za-z_-]{2})(?P<kind2>[A-Z]{4})(?P<size2>[0-9A-Za-z_-]{4})\.'
Traceback (most recent call last):
  File "/keripy/venv/lib/python3.12/site-packages/lmdb/__init__.py", line 42, in <module>
    from lmdb.cpython import *
ModuleNotFoundError: No module named 'lmdb.cpython'
During handling of the above exception, another exception occurred:
....
Exception: This CFFI feature requires setuptools on Python >= 3.12. The setuptools module is missing or non-functional.
```

### Steps to reproduce

just run docker images

PR #1438, opened 2026-06-14
fix(docker): align image with Python 3.14 requirement

jaelliot:
Align keripy Dockerfile with the existing Python `>=3.14.0` requirement and ensure `setuptools` is available for `lmdb`/CFFI initialization.

## Context

`WebOfTrust/keripy#1175` reports that `docker run weboftrust/keri:1.2.7 version` fails with an `lmdb.cpython` import error followed by a setuptools/CFFI error on Python >=3.12.

The Dockerfile was still pinned to a Python 3.13 Alpine image even though keripy currently requires Python >=3.14.0. Python 3.14 virtual environments also do not provide setuptools by default, so the Docker build needs to bootstrap it explicitly before installing dependencies.

## Changes

- Update `images/keripy.dockerfile` to use a Python 3.14 Alpine base image
- Install `setuptools` during Docker build alongside the pip upgrade
- Leave KERI source/protocol logic unchanged

## Validation

- `docker build -f images/keripy.dockerfile -t keripy-issue-1175-fixed .`
- `docker run --rm keripy-issue-1175-fixed version`
- `docker run --rm --entrypoint python keripy-issue-1175-fixed -c "import setuptools; import lmdb; print('setuptools ok'); print('lmdb ok')"`
- Pre-push test suite: 570 passed, 3 skipped

References #1175.

### DIFF
--- a/images/keripy.dockerfile
+++ b/images/keripy.dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=python:3.13.2-alpine3.20
+ARG BASE=python:3.14.6-alpine3.24
 
 FROM ${BASE} AS builder
 
@@ -23,7 +23,7 @@ RUN python -m venv venv
 
 ENV PATH=/keripy/venv/bin:${PATH}
 
-RUN pip install --upgrade pip
+RUN pip install --upgrade pip setuptools
 RUN mkdir /keripy/src
 
 COPY requirements.txt setup.py ./


dhh1128 (commented):
This looks right to me, and thanks for chasing down #1175. I confirmed the mismatch you describe: `setup.py` already has `python_requires='>=3.14.0'` and CI runs on `3.14`, so the Dockerfile still sitting on `python:3.13.2` was genuinely out of step — good catch. The `python:3.14.6-alpine3.24` base tag checks out as a valid published image, and bootstrapping `setuptools` explicitly makes sense to me given that 3.12+ virtual environments stopped shipping it by default, which lines up with the lmdb/CFFI import error in the issue. I'm happy with this as-is. One tiny, entirely optional thought: since the whole point is keeping the image aligned with our version floor, a one-line comment near the `ARG BASE=` pin noting "keep in sync with `python_requires` in setup.py" might save a future reader the same archaeology — but that's a nicety, not a blocker.

dhh1128 (approved):
Approving — this is a correct, well-scoped fix. It brings the image back in line with the `>=3.14.0` floor already declared in setup.py (and the version CI runs on), the `python:3.14.6-alpine3.24` base tag is valid and current, and the explicit setuptools bootstrap is the right remedy for the lmdb/CFFI build failure now that 3.12+ venvs stopped shipping setuptools by default. The keep-in-sync comment I mentioned is optional. Thanks for chasing down #1175.

jaelliot:
Note: CI failures (KWA NameError in test_verifier_e1e_identity_edge) are pre-existing on `main` (commit 9161a70). Confirmed by checking the same commit's CI status — all three platforms fail identically. Not caused by this PR.

dhh1128:
I'm agreeing with @jaelliot that the CI failures came from elsewhere. They were my fault. They have been fixed now, but I don't have permission to request a re-run of CI.

Issue #1000, opened 2025-05-07
Turn logging on by default with PYTHONUNBUFFERED = True


kentbull:
### Feature request description/rationale

This will help newcomers avoid wasting tons of time. We should default PYTHONUNBUFFERED to True.

kentbull:
I am referring specifically to the weboftrust/keri, keria, witness, and other Docker images. They should all default to PYTHONUNBUFFERED as True. If people need better performance by turning this off then they can manually set it to fals.e

kentbull:
Yes, that's right, just for the docker images. I'll submit a PR.

Issue #703, opened 2024-03-06
Document config file labels


SmithSamuelM:
The keripy config file has well known labels for config data but these are not documented anywhere:

Suggest label definition in the doc strings of keri.app.configing
and documented labels with examples in a markdown doc in scripts/keri/cf

reserved labels:
dt
curls
iurls
durls
wurls

Hab name labels: "nel"
 
Not used in config file?
murls

```
{'dt': '2021-01-01T00:00:00.000000+00:00',
                                   'nel': {
                                       'dt': '2021-01-01T00:00:00.000000+00:00',
                                       'curls': ['tcp://localhost:5621/'],
                                   },
                                   'iurls': ['tcp://localhost:5620/?role=peer&name=tam']}

```

Habery. reconfigure

```python
def reconfigure(self):
        """Apply configuration from config file managed by .cf. to this Habery
        Process any oobis or endpoints

        conf
        {
          dt: "isodatetime",
          curls: ["tcp://localhost:5620/"],
          iurls: ["tcp://localhost:5621/?name=eve"],
        }

        Config file is meant to be read only at init not changed by app at
        run time. Any dynamic app changes must go in database not config file
        that way we don't have to worry about multiple writers of cf.
        Use config file to preload database not as a database. Config file may
        have named sections for Habery or individual Habs as needed.

        """
        conf = self.cf.get()
        if "dt" in conf:  # datetime of config file
            dt = help.fromIso8601(conf["dt"])  # raises error if not convert
            if "iurls" in conf:  # process OOBI URLs
                for oobi in conf["iurls"]:
                    obr = basing.OobiRecord(date=help.toIso8601(dt))
                    self.db.oobis.put(keys=(oobi,), val=obr)
            if "durls" in conf:  # process OOBI URLs
                for oobi in conf["durls"]:
                    obr = basing.OobiRecord(date=help.toIso8601(dt))
                    self.db.oobis.put(keys=(oobi,), val=obr)
            if "wurls" in conf:  # well known OOBI URLs for MFA
                for oobi in conf["wurls"]:
                    obr = basing.OobiRecord(date=help.toIso8601(dt))
                    self.db.woobi.put(keys=(oobi,), val=obr)
```

BaseHab.reconfigure

```python

    def reconfigure(self):
        """Apply configuration from config file managed by .cf. to this Hab.
        Assumes that .pre and signing keys have been setup in order to create
        own endpoint auth when provided in .cf.

        conf
        {
          dt: "isodatetime",
[...]

PR #1439, opened 2026-06-15
docs: document configuration file labels

jaelliot:
Documents keripy configuration file structure requested in issue #703.

## Scope

- Documentation only.
- Adds a configuration reference for top-level Habery labels (`dt`, `iurls`, `durls`, and `wurls`).
- Documents per-habitat sections keyed by habitat name, with nested `dt` and `curls`.
- Clarifies that names such as `tam` and `nel` are example habitat names, not reserved labels.
- Uses `Habery.reconfigure()`, `Hab.reconfigure()`, and `test_habery_reconfigure_v1()` as source material.
- No Python, Docker, dependency, or CI changes.
- Leaves parked issue #1000 Docker work untouched.

## Validation

- `git diff --check`
- Documentation review against the current reconfiguration code and tests

References #703.

### DIFF
--- a/docs/config-labels.md
+++ b/docs/config-labels.md
@@ -1,0 +1,61 @@
+# Configuration File Labels
+
+## Overview
+
+KERI habitats store configuration in a JSON (or HJSON, MsgPack, or CBOR) file
+managed by `src/keri/app/configing.py`.  This document describes the labels
+(keys) used in that configuration file.
+
+The labels below are drawn from the existing `Configer` docstring examples and
+the hab configuration logic in `src/keri/app/habbing.py`.
+
+## Example configuration
+
+```json
+{
+    "dt": "2021-01-01T00:00:00.000000+00:00",
+    "nel": {
+        "dt": "2021-01-01T00:00:00.000000+00:00",
+        "curls": [
+            "tcp://localhost:5621/"
+        ]
+    },
+    "iurls": [
+        "tcp://localhost:5620/?role=peer&name=tam"
+    ],
+    "durls": [
+        "http://127.0.0.1:7723/oobi/EBNaNu-M9P5cgrnfl2Fvymy4E_jvxxyjb70PRtiANlJy",
+        "http://127.0.0.1:7723/oobi/EMhvwOlyEJ9kN4PrwCpr9Jsv7TxPhiYveZ0oP3lJzdEi"
+    ],
+    "wurls": [
+        "http://127.0.0.1:5644/.well-known/keri/oobi/EBNaNu-M9P5cgrnfl2Fvymy4E_jvxxyjb70PRtiANlJy?name=Root"
+    ]
+}
+```
+
+## Label reference
+
+| Label   | Meaning                          | Type / shape     | Example                                          | Notes |
+|---------|----------------------------------|------------------|--------------------------------------------------|-------|
+| `dt`    | Datetime                         | ISO 8601 string  | `"2021-01-01T00:00:00.000000+00:00"`            | Timestamp of the configuration file. Must be parseable by `fromIso8601`. |
+| `nel`   | Network endpoint list            | Object           | `{"dt": "...", "curls": ["tcp://..."]}`          | Contains the controller's own datetime and controller URLs. |
+| `curls` | Controller URLs                  | Array of strings | `["tcp://localhost:5621/"]`                      | TCP endpoints where this controller accepts connections. Used by witnesses and other peers to reach the controller. Nested inside `nel`. |
+| `iurls` | Introduction OOBI URLs           | Array of strings | `["tcp://localhost:5620/?role=peer&name=tam"]`   | Out-of-band introduction URLs. Written to `db.oobis` so the habitat can discover other identifiers. |
+| `durls` | Delegation OOBI URLs             | Array of strings | `["http://127.0.0.1:7723/oobi/EB..."]`           | Delegation OOBI URLs. Written to `db.oobis`. Used when the identifier has delegated authority relationships. |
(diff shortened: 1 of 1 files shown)


docs/config-labels.md
```diff
@@ -0,0 +1,61 @@
...
+    ]
+}
+```
+
+## Label reference
+
+| Label   | Meaning                          | Type / shape     | Example                                          | Notes |
+|---------|----------------------------------|------------------|--------------------------------------------------|-------|
+| `dt`    | Datetime                         | ISO 8601 string  | `"2021-01-01T00:00:00.000000+00:00"`            | Timestamp of the configuration file. Must be parseable by `fromIso8601`. |
+| `nel`   | Network endpoint list            | Object           | `{"dt": "...", "curls": ["tcp://..."]}`          | Contains the controller's own datetime and controller URLs. |
```
ryan-hansen: This is not correct.  "nel" is not network endpoint list, it's just keyed based on the hab’s name (e.g. "tam", "nel"). It's an object that contains that hab’s dt and curls for controller endpoint setup.  You can see an example of how it's used in habbing.py about line 1230.

dhh1128 (commented):
Thanks for putting this together — a reference for the config labels is something we've needed for a while, and issue #703 has been open a long time. Most of it matches the code well: `iurls`/`durls` going into `db.oobis`, `wurls` into `db.woobi`, `dt` parsed by `fromIso8601`, and the format and tail-dir notes all agree with `configing.py`.

I have one question about `nel` and `curls`, and I might be missing something. As far as I can tell, `nel` appears only in the docstring examples (in `configing.py` and the two `reconfigure` docstrings in `habbing.py`) — I couldn't find code that reads a `nel` key. And the place that consumes `curls` (`Hab.reconfigure`) reads it from the top level of the per-name config, `conf[self.name]["curls"]`, not from inside `nel`. So I worry the "Nested inside `nel`" note, and describing `nel` as a processed "Network endpoint list", could lead a reader to write a config the code doesn't honor. Would it make sense to mark `nel` as an example-only grouping that isn't parsed, or drop it and show `curls` at the level the code reads it?

Relatedly, the single example JSON seems to blend two config shapes — the Habery-level file with top-level `iurls`/`durls`/`wurls`, and the per-identifier Hab-level file (keyed by hab name) with `curls`. I found that a little confusing to trace; would splitting them into two small examples, or a sentence noting they come from different files, make it clearer? I could be wrong on the details, so push back if I've misread the flow.

evanja57:
I'll second everything that @dhh1128 has said in his comment, and add that I think `nel` is just an example habery name that was used in the test file (`test_habbing_v1.py:545` in `test_habery_reconfigure_v1()`). 
In the test we do assert:
```Python
        assert nelHby.cf.get() == {'dt': '2021-01-01T00:00:00.000000+00:00',
                                   'nel': {
                                       'dt': '2021-01-01T00:00:00.000000+00:00',
                                       'curls': ['tcp://localhost:5621/'],
                                   },
                                   'iurls': ['tcp://localhost:5620/?role=peer&name=tam']}
```
But here nel is just the name of the hab. There is another instance in that test where it is `tam` instead of `nel`, which again is just the name of the hab. A more clear example shape would be something like
```Json
{
    "dt": "...",
    "<hab-name>": {
        "dt": "...",
        "curls": [...]
    },
    "iurls": [...],
}
```

Also, I agree that the per hab vs habery level label reading and reconfiguring split is a bit confusing to trace and should probably be documented separately. Something perhaps with an example shape like:
```Json
{
  "dt": "2026-07-30T00:00:00.000000+00:00",
  "iurls": ["..."],
  "durls": ["..."],
  "wurls": ["..."],

  "<hab-name>": {
    "dt": "2026-07-30T00:00:00.000000+00:00",
    "curls": ["http://localhost:5621/"]
  }
}
```
And then explain how 
`Habery.reconfigure()` reads `dt`, `iurls`, `durls`, and `wurls`.
vs
`hab.reconfigure()` reads `dt` and `curls`

ryan-hansen (changes requested):
Man, I made this comment a long time ago.  Forgot to fully submit the review.

docs/config-labels.md
jaelliot: Thanks for catching this, Ryan. You are correct — `nel` is not a reserved label or acronym; it is an example habitat name, the same way `tam` is.

The correcting commit (`435da77`) rewrites the documentation so that:
- The habitat-name key is documented as dynamic.
- `nel` and `tam` are explicitly described as example habitat names, not reserved configuration labels.
- Habery-level fields (`dt`, `iurls`, `durls`, `wurls`) and habitat-level fields (`dt`, `curls`) are documented separately.
- The JSON example uses `"<hab-name>"` as a placeholder.

Sources verified: `src/keri/app/habbing.py` (`Habery.reconfigure()`, `Hab.reconfigure()`), `src/keri/app/configing.py`, and `tests/app/test_habbing_v1.py`.

The PR description has also been updated to reflect the corrected documentation structure.

jaelliot:
Thanks for catching this, Ryan. You are correct — `nel` is not a reserved label or acronym; it is an example habitat name, just like `tam`.

The correcting commit `435da77` rewrites the documentation so that:
- The habitat-name key is documented as dynamic.
- `nel` and `tam` are explicitly described as example habitat names, not reserved configuration labels.
- Habery-level fields (`dt`, `iurls`, `durls`, `wurls`) and habitat-level fields (`dt`, `curls`) are documented in separate tables.
- The JSON example uses `"<hab-name>"` as a placeholder.

Sources verified: `src/keri/app/habbing.py` (`Habery.reconfigure()`, `Hab.reconfigure()`), `src/keri/app/configing.py`, and `tests/app/test_habbing_v1.py`.

The PR description has also been updated to reflect the corrected documentation structure.

ryan-hansen (approved):

PR #1440, opened 2026-06-16
Multisig catchup

kentbull:
Adds multisig catchup by expanding `kli vc export` with new all registry, all credential flags and `kli import` to process everything exported by `kli vc export`. 

Also adds:
- pulls in`kli vc registry rename` from `1.1.x` - useful after importing a registry
- backports `kli vc schema import` from `main` - this is already on `main`, backporting it here to simplify export/import tests.
- `scripts/workflows`
  - `/multisig-export-ingest.sh` - a test demonstrating export, import, schema import, and registry rename
  - `/schema/gleif-vlei-credential-schema.json` - the schema file imported in the above script
  - `/data/credential-data.json` - the credential data used for issuing credentials exported and imported in the above script
- `--registry-name` option to `kli multisig join` so the command is fully non-interactive (doesn't require input) to support simple scripting.

And there was a bug I fixed that somehow did not land on 1.2.x yet with the `kli vc registry incept` command where `registry_kwa` had to be defined to prevent Hio's `temp` arg from blowing up Registry.make.


kentbull:
The stress test script exercising this `kli vc export` and the `kli import` code is in the [qvi-software repo PR here](https://github.com/GLEIF-IT/qvi-software/pull/52).

kentbull:
Closed as merged as a bunch of smaller PRs yesterday.

PR #1441, opened 2026-06-17
Default v2 with v1 tests

ryan-hansen:
This is the complete V1 test PR.  I'm opening this as a draft for review but I'm going back over the core code changes to make sure they are all correct and necessary.  

UPDATE: Scripts tests not ready yet.  Work in progress.


tests/app/test_oobiing.py
```diff
@@ -23,11 +23,14 @@
 from keri.peer import Exchanger
 from keri.recording import OobiRecord
 
+V2 = Vrsn_2_0
```
evanja57, line 26: const = const probably just makes things more confusing

ryan-hansen: Good catch.  Got a little snow blind on that one I guess.

src/keri/app/oobiing.py
evanja57: self.parser = Parser(framed=True, kvy=kvy, rvy=rvy, version=Vrsn_2_0)

ryan-hansen: Done

kentbull:
How exciting! Glad to see so much work on v2. Keep it up.

src/keri/app/habbing.py
evanja57: hab.exchange will fail on v2 because of specialExchange only supporting v1. We probably need to use eventing.exchange when version = 2.

evanja57:
Also in serdering.py we might want to consider bumping the pvrsn and kind lines 354-360:
```Python
    Protocol = None  # class based message protocol, None means any in Protocols is ok
    Proto = Protocols.keri  # default message protocol type for makify on base Serder
    PVrsn = Vrsn_1_0  # default protocol version
    GVrsn = Vrsn_2_0  # default CESR genus version
    Kind = Kinds.json  # default serialization kind
    Genus = GenDex.KERI  # default CESR genus code
    MUCodes = Counter.MUCodes # message universal code tables from Counter
```
That way callers get default v2 instead of default v1 when calling without explicitly passing those args

src/keri/app/httping.py
evanja57, line 95: I think cesr requests will fail because it is not json starting with '{'. Same with line 188 in this file checking for 0x7b which is '{'

SmithSamuelM:
Yes the reason specialExchange can only be v1 is that it depends on embedding other messages inside the exchange. This embedding requires that embedded messages share the same serialization of the embedding message. Which breaks down for v2 with 4 serialization types.  (really on works if everything is json). So the fix in version two is not to embed but to attach nested substreams that have the otherwise embedded message. I am working on adding this capability to the parser and then to exchange. So for now, full ipex exchanges with embedded messages can only work in v1. So those tests that use the ipex messages that use specialExchange must be hardcoded to v1

PR #1444, opened 2026-06-19
Xdist test parallel

evanja57:
In this PR:

- Add pytest xdist to CI and run Ubuntu and macOS with pytest -n auto
- Keep Windows as two serial jobs, core and non core, because Windows xdist was slower and larger Python projects shard Windows when needed [DVC tests](https://github.com/iterative/dvc/blob/main/.github/workflows/tests.yaml#L57-L131), [conda lock tests](https://github.com/conda/conda-lock/blob/main/.github/workflows/test.yml#L33-L81)
- Move the flake8 checks into one Ubuntu lint job so lint runs once in parallel with tests instead of running on every test job
- Use GitHub matrix include to for each OS test command and add the two Windows jobs without adding separate workflow jobs
- Replace fixed localhost ports with dynamically allocated ports, then pass generated witness URLs into seedWitEnds so parallel workers do not reuse hardcoded endpoints
- Add shared xdist port reservations under the pytest base temp parent, using atomic lock file creation so worker processes don't reserve the same fixture issued port [xdist shared fixtures](https://pytest-xdist.readthedocs.io/en/stable/how-to.html#making-session-scoped-fixtures-execute-only-once), [xdist worker model](https://pytest-xdist.readthedocs.io/en/stable/how-it-works.html)
- Replace the path deletions in test_habbing_v1 and test_habbing_v2 with unique Habery names and bases, so the tests don't delete shared global state during parallel runs
  
  
Observed speedup:

- I ran these a bunch on my fork for timing estimates as I was developing
  - Windows: 16m15 avg before -> 5 min after  
  - Ubuntu: 2m11 avg before -> 1m23 after
  - macOS: 4m20 avg before -> 2m39 after

### DIFF
--- a/.github/workflows/python-app-ci.yml
+++ b/.github/workflows/python-app-ci.yml
@@ -11,6 +11,28 @@ on:
   workflow_dispatch:
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python 3.12.6+
+        uses: actions/setup-python@v5
+        with:
+          python-version: '>=3.12.6'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install flake8
+          if (Test-Path requirements.txt) { pip install -r requirements.txt }
+        shell: pwsh
+      - name: Lint changes
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --ignore=E7,F841,E301,E302,E303 --max-complexity=10 --max-line-length=127 --statistics
+        shell: pwsh
+
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 35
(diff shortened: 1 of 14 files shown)


tests/conftest.py
```diff
@@ -123,9 +124,70 @@ def seeder():
     return DbSeed
 
 
+def _unused_tcp_port():
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        return sock.getsockname()[1]
```
ryan-hansen, line 130: There may be a binding gap problem here.  Probably pretty low likelihood we would run into it, but possible, because as soon as the context manager exits, the socket is closed, so the port is no longer actually bound until the server starts.  So, in theory, another worker could steal the port in the gap.  A simple work-around would be to add some retry/exception handling where the port is actually used (or tried) and if it throws an "already in use" error, just try it again with a different port.  That's probably over-simplified, and maybe overkill unless we start seeing port binding failures, but maybe worth taking a look at.

.github/workflows/python-app-ci.yml
```diff
@@ -40,19 +75,12 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install flake8 pytest
+          pip install pytest pytest-xdist
```
ryan-hansen, line 168: We're adding x-dist to the CI pipeline but not to the actual project dependencies, right?  So xdist won't work locally until you manually install it.  Not sure if we want to add it to the "extras_require" block in setup.py, but maybe at minimum add something to the README about installing it and running tests in parallel.

ryan-hansen (dismissed):

PR #1447, opened 2026-06-23
feat: consume and discard KERIACDCGenusVersion

Sotatek-Patrick-Vu:
For forwards compatibility in KERIA's parser with genusified streams

### DIFF
--- a/src/keri/core/parsing.py
+++ b/src/keri/core/parsing.py
@@ -705,9 +705,17 @@ class Parser:
 
         cold = sniff(ims)  # check for spurious counters at front of stream
         if cold in (Colds.txt, Colds.bny):  # not message error out to flush stream
-            # replace with pipelining here once CESR message format supported.
-            raise kering.ColdStartError("Expecting message counter tritet={}"
-                                        "".format(cold))
+            # strip KERIACDCGenusVersion counter if present (genusify=True); else error
+            ctr = yield from self._extractor(ims=ims, klas=Counter, cold=cold)
+            if ctr.code != CtrDex_1_0.KERIACDCGenusVersion:
+                raise kering.UnexpectedCountCodeError("Expected KERIACDCGenusVersion "
+                                                      "counter got code={} tritet={}"
+                                                      "".format(ctr.code, cold))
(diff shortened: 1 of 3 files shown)


src/keri/core/parsing.py
```diff
@@ -705,9 +705,17 @@ def msgParsator(self, ims=None, framed=True, pipeline=False,
 
         cold = sniff(ims)  # check for spurious counters at front of stream
         if cold in (Colds.txt, Colds.bny):  # not message error out to flush stream
-            # replace with pipelining here once CESR message format supported.
-            raise kering.ColdStartError("Expecting message counter tritet={}"
-                                        "".format(cold))
+            # strip KERIACDCGenusVersion counter if present (genusify=True); else error
+            ctr = yield from self._extractor(ims=ims, klas=Counter, cold=cold)
+            if ctr.code != CtrDex_1_0.KERIACDCGenusVersion:
+                raise kering.UnexpectedCountCodeError("Expected KERIACDCGenusVersion "
```
kentbull, line 711: This should default the version to v1.0 rather than throw an error because existing components that send the old v1.0 streams should not throw an error in the parser. The parser should assume they are v1.0 streams and parse them.

kentbull (changes requested):

src/keri/core/parsing.py
iFergal: Existing v1.0 streams start with `Colds.msg` so they do not hit this if block. So this PR essentially keeps the old behaviour, but _if_ there if the `KERIACDCGenusVersion` counter is prepended it will just consume it

kentbull: I see the generic exception block in each of the places that call `msgParsator`, so it looks like we're good to merge.

kentbull (approved):

src/keri/core/parsing.py
kentbull: As in, the exception raised here, after consuming the `KERIACDCGenusVersion` count code, would be caught by the outer parsing call site exception handlers, so this would work.

PR #1452, opened 2026-06-23
fixing empty config file logic

evanja57:
- in runWitness, configFile is defaulted to ""
- ("" is not None) is true because "" is a string object, meaning with no config file arg it will by default create an empty config file
  - "" is falsy, and "any string" is truthy
- I was annoyed by the empty keri/cf/main/.json that was getting created every time I ran tests


src/keri/cli/commands/witness/start.py
ryan-hansen: Should the signature also change to configFile=None rather than ""?

evanja57: yeah good call; just pushed that

kentbull (approved):

PR #1470, opened 2026-06-29
feat: kli vc registry rename

kentbull:
Adds support for renaming a registry. Can look up a registry by SAID or name.

This is a part of a sequence of PRs preparing for the export and import behavior.

### DIFF
--- a/src/keri/app/cli/commands/vc/registry/rename.py
+++ b/src/keri/app/cli/commands/vc/registry/rename.py
@@ -1,0 +1,125 @@
+# -*- encoding: utf-8 -*-
+"""
+KERI
+keri.kli.commands.vc.registry.rename module
+
+"""
+import argparse
+
+from hio.base import doing
+
+from keri.app.cli.common import existing
+from keri.vdr import credentialing, viring
+
+parser = argparse.ArgumentParser(description='Rename a local credential registry')
+parser.set_defaults(handler=lambda args: handler(args),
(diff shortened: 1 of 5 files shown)


dhh1128 (commented):
This looks solid, and I appreciate how thoroughly `test_rename.py` walks the branches — missing, already-exists, idempotent-same-registry, and rename-by-SAID are all covered, which made it easy to convince myself the logic is right. I traced how `credentialing.loadRegistries()` (`credentialing.py:62`) rebuilds each Registry's name straight from the `reger.regs` key, so re-keying the record is the whole job — there's no second index to fall out of sync.

A few small things stood out, none blocking:

- In `renameRegistry()` (`rename.py`) the failure paths `return -1` (registry not found, name already exists, put failed), but since we're inside `Doer.recur`, I think those return values get discarded and the process still exits 0. Sibling commands like `incept.py` return `-1` from the handler itself to signal a nonzero exit for scripting. Was an exit code on failure intended here, or is printing the message enough for now? I might be missing how `recur`'s return is used.
- When both `--registry-name` and `--registry-said` are given, `_registryRecord()` silently lets the SAID win and ignores the name (the tests rely on this). That precedence seems reasonable — would a one-line note in the `--registry-said` help text help, so a user who passes both isn't surprised?
- Tiny nit: `rename.py` is missing its trailing newline at EOF.

Nice piece of work.

PR #1472, opened 2026-06-30
feat: consume and discard KERIACDCGenusVersion

Sotatek-Patrick-Vu:
For forwards compatibility in KERIA's parser with genusified streams

### DIFF
--- a/src/keri/core/counting.py
+++ b/src/keri/core/counting.py
@@ -71,7 +71,7 @@ class CounterCodex_1_0(MapDom):
     BigGenericGroup: str = '-0W'  # Big Generic Material Quadlets
     ESSRPayloadGroup: str = '-Z'  # ESSR Payload Group Quadlets
     BigESSRPayloadGroup: str = '-0Z'  # Big ESSR Payload Group Quadlets
-    KERIACDCGenusVersion: str = '--AAA'  # KERI ACDC Protocol Stack CESR Version
+    KERIACDCGenusVersion: str = '-_AAA'  # KERI ACDC Protocol Stack CESR Version
 
 
     def __iter__(self):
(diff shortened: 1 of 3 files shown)


dhh1128 (commented):
I think the mechanics here are right. I checked that `-_AAA` is `Cizage(hs=5, ss=3, fs=8)`, and that the genus/version counter carries its payload in the 3-char soft field with no counted material following it — so extracting the 8 characters and then either `pass`-ing (`parsing.py`, attachment loop) or re-sniffing (`parsing.py`, cold-start branch) consumes exactly the counter and nothing more. I also confirmed the `--AAA` → `-_AAA` move matches the corrected code table rather than inventing a value.

There are two things I'd like to understand before I'm fully comfortable, since this touches stream-parsing semantics:

- The new logic only checks `ctr.code == KERIACDCGenusVersion` and then discards, without looking at the version in the soft part (`ctr.count`). So if a peer ever prepends a genus/version for a different CESR version, it looks like we'd silently accept it and keep parsing the following material with our current tables. I take it the intent is just to tolerate the genus counter that newer producers prepend to what is still a v1 stream — in which case this is fine — but is ignoring the announced version what we want, or should an unexpected version be rejected rather than assumed compatible?
- I don't see a test that feeds a genusified stream through `Parser`/`msgParsator`. The added tests cover the `counting.py` code-point tables, but the two new branches in `parsing.py` (the cold-start strip and the attachment-loop `pass`) look untested. Given how strict we usually are about a failing test first, would it make sense to add a round-trip: a message prefixed with `-_AAAAAA` parses cleanly, and a leading non-genus counter still raises `UnexpectedCountCodeError`? That would also pin the byte-consumption I reasoned about above, so a future refactor can't quietly break it.

There's a minor point too, no change needed: now that the genus code is `-_`, the `Hards.update([('--', 5)])` entry looks orphaned in the 1.x tables. It's not harmful as far as I can tell — just possibly dead config worth a glance.

Thanks for doing this — the forwards-compat direction makes sense.

PR #1477, opened 2026-07-03
Log kli witness start failures and prevent startup stalls (#238)

### LINKED ISSUE #238: Ensure `kli witness start` logs errors correctly

dhh1128:
## Summary

`kli witness start` could stall during startup and did not reliably log error
conditions (#238). This PR makes the command log every startup/runtime failure
visibly, eliminates the interactive-prompt stall, and fixes a few related
logging-correctness defects. Fixes #238.

## Changes

**Error logging & startup stalls** (`src/keri/cli/commands/witness/start.py`)
- Startup/runtime failures (e.g. an HTTP/TCP port already in use) are now logged
  at **CRITICAL with a traceback**, so they surface even at the default
  `--loglevel CRITICAL`, and the `Habery` is closed in a `finally` block so a
  failed start no longer leaves a stale LMDB lock behind.
- An encrypted keystore started with **no passcode on a non-TTY** now fails fast
  with a logged `AuthError` instead of blocking on an interactive `getpass`
  prompt — the literal "stalled in startup" symptom for witnesses run under
  systemd/docker/scripts.
- `--loglevel` is normalized with `.upper()`, so a lowercase level such as
  `debug` is honored instead of silently becoming an invalid level string.

**`--logfile` → `--logdir`**
- `--logfile` is deprecated in favor of the accurately-named `--logdir`. The
  logging backend derives the log file name from `--name` and the subdirectory
  from its own prefix, so only the directory is meaningful; the old code
  mis-assigned the whole file path as the log *head directory*. `--logfile` is
  retained as an alias (its directory portion is used), and a deprecation notice
  is printed to stderr reporting the actual resolved log path.

## Tests

Regression tests added to `tests/app/cli/test_kli_commands.py` (co-located with
the existing witness-start test), covering: loglevel normalization,
`--logdir`/`--logfile` handling and directory extraction, the stderr deprecation
notice, CRITICAL-level failure logging + `Habery` cleanup, and the non-TTY
fail-fast guard. Each assertion was mutation-checked — reverting the
corresponding fix turns the test red on the targeted assertion.

## Notes for reviewers

- The default `--loglevel` is unchanged (still `CRITICAL`); fatal failures are
  logged at CRITICAL specifically so they remain visible without making the
  witness verbose by default.
- The non-TTY guard is placed at the witness (daemon) layer. `setupHby` — shared
  by many *interactive* `kli` commands — is the underlying `getpass` choke point,
  but making it non-interactive globally is out of scope here and would change
  behavior for commands that legitimately prompt.

---

This code was developed with claude (using the Opus 4.8 model, mostly). However, I (Daniel the human) have read, reviewed, and worked the code, and I stand behind it and take responsibility for it.

### DIFF
--- a/ref/ChangeLog.md
+++ b/ref/ChangeLog.md
@@ -2,6 +2,17 @@
 
 ## 2.0.0-dev6
 
+*** kli witness start now logs startup/runtime failures at CRITICAL (with traceback) so
+they are visible even at the default --loglevel, and closes the Habery on failure so a
+failed start leaves no stale LMDB lock. An encrypted keystore started with no passcode on
+a non-TTY now fails fast with a logged AuthError instead of stalling on an interactive
+getpass prompt. --loglevel is normalized so a lowercase level (e.g. debug) is honored.
+See WebOfTrust/keripy#238.
+
+*** kli witness start: --logfile is deprecated in favor of --logdir. The witness writes
+its log file under the given directory; --logfile is retained as an alias whose directory
+portion is used (it previously mis-set the log head directory to the file path itself).
+
 *** peer.exchange has been replaced with a combination of either core.eventing.exchange
 or peer.specialExhange. peer.specialExchange supports the embeds and diger parameters
 and returns a tuple of (exn, atc) exn Serder instance and atc attachments bytearray.
--- a/src/keri/cli/commands/witness/start.py
+++ b/src/keri/cli/commands/witness/start.py
@@ -7,6 +7,8 @@ Witness command line interface
 """
 import argparse
 import logging
+import os
+import sys
 
 from hio.help import ogler
 
@@ -16,6 +18,9 @@ from ...common import Parsery, setupHby
 
 from ....app import (Habery, HaberyDoer, Keeper, Configer,
                      runController, setupWitness)
+from ....kering import AuthError
+
+logger = ogler.getLogger()
 
 
 d = "Runs KERI witness controller.\n"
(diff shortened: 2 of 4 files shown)


kentbull:
Fixes https://github.com/WebOfTrust/keripy/issues/238

dhh1128:
Heads-up on the failing `test (ubuntu-latest)` check — it is **not** a regression from this PR.

The sole failure is `tests/app/cli/test_kli_commands.py::test_standalone_kli_commands` → `DatabaseError: Database migrations must be run. DB version None; current 2.0.0-dev6` (599 passed, 1 failed). It's a pre-existing test-isolation issue:

- It reproduces independent of this PR's diff, and **only** on the `ubuntu-latest` job, which runs `pytest tests/ … -n auto` (pytest-xdist). The macOS and sharded Windows jobs, which don't use `-n auto`, all pass — as does this test in isolation.
- Root cause: ~30 tests across the suite create a *non-transient* keystore literally named `test` at `/usr/local/var/keri/ks/test`. Under `-n auto` these are distributed across workers that race on that shared directory, so `test_standalone_kli_commands` intermittently opens a store another worker is mid-initialising (version `None`) and the migration guard raises.
- This became visible when #1444 enabled `-n auto` for the ubuntu job.

Suggested fix (test infra, out of scope for this witness-logging PR): give the shared-`test`-store tests unique/temp store names, or pin them to one worker via `@pytest.mark.xdist_group(...)`. Happy to open a separate issue/PR for that if useful.

(For the record, an earlier push of this branch did briefly add 4 `TypeError` failures in `test_kraming`/`test_witness`; those were a real leak — a new test drove the CLI with `--loglevel debug` and left the shared `ogler.level` at DEBUG, surfacing latent malformed `%`-format log statements on other workers. That's fixed in 1ad7fa44 by restoring `ogler.level`.)

dhh1128:
Resolved in 178b2fa3: `test_standalone_kli_commands` was the only test using the generic non-temp keystore name `test` at `/usr/local/var/keri/ks/test`; giving it a unique name isolates its store from concurrent xdist workers. `test (ubuntu-latest)` is now green. This follows the same per-worker isolation philosophy #1444 applied to port allocation. The broader hardening (unique/temp stores or a per-worker keri root for any other fixed-name non-temp stores) still belongs in test infra if you want defense-in-depth, but this PR's CI is now clean.

kentbull:
This looks good to me. Please open a similar PR to v1.2.14 (and probably 1.3.5) so we can get this in the next release. I will do it for you if you would rather.

kentbull (approved):

dhh1128:
Opened the two backports as you requested:

- **1.3 line** → #1486 (base `v1.3.5`)
- **1.2 line** → #1487 (base `v1.2.14` — looks like `v1.2.13` was renamed to `v1.2.14`, so I targeted that)

Both are the same fix adapted to the 1.x CLI layout (`keri.app.cli.*`, absolute imports, `help.ogler`). I split each into individually-droppable commits: on 1.3 there's an extra first commit restoring the boot-Keeper `ks.close()` (#1367) that 1.3.5 is missing — it's actually a prerequisite there, since the witness can't reach a successful Habery re-open without it. The `--logfile`→`--logdir` rename is its own commit on both in case you'd rather not change the CLI surface in a patch release. Details in each PR description.

dhh1128:
@kentbull — reviving this one. You approved it on 2026-07-10, but GitHub currently shows no active review decision (the approval appears to have been dismissed by the later commit that fixed the xdist keystore-name collision). Could you re-approve so it can merge?

Status of the three-line rollout you asked for:

- **1.2 line — #1487: merged** 2026-07-14
- **1.3 line — #1486: approved, still open**
- **`main` — this PR: still open**

So the fix is currently live only on the oldest line. CI here is green across all platforms including the sharded Windows jobs. Fixes #238.

kentbull (approved):

PR #1481, opened 2026-07-08
Enable KRAM for v2 query workflows

KeaxD:
- Routes v2 `qry`, `rpy`, `exn`, `xip`, `pro`, and `bar` messages through `Kevery.processMsg`
- Adds `Kevery(enableKram=True)` to create a `Kramer` from config
- Enables KRAM in witness `Reactant` using the habitat config
- Uses the habitat parser version for TCP witness messengers instead of hardcoded v1
- Fixes v2 DB replay attachment encoding in `cloneEvtMsg()`
- Adds v2 test coverage for parser/KRAM routing, Habery reply flows, and `WitnessInquisitor`

### DIFF
--- a/src/keri/app/directing.py
+++ b/src/keri/app/directing.py
@@ -490,11 +490,11 @@ class Reactant(doing.DoDoer):
         #  needs unique kevery with ims per remoter connnection
         rvy = Revery(db=hab.db)
         self.kevery = Kevery(db=self.hab.db,
-                                      cf=self.hab.cf,
-                                      enableKram=True,
-                                      lax=False,
-                                      local=False,
-                                      rvy=rvy)
+                             cf=self.hab.cf,
+                             enableKram=True,
+                             lax=False,
+                             local=False,
+                             rvy=rvy)
 
         if self.verifier is not None:
             self.tevery = Tevery(reger=self.verifier.reger,
--- a/src/keri/app/directing.py
+++ b/src/keri/app/directing.py
@@ -490,6 +490,8 @@ class Reactant(doing.DoDoer):
         #  needs unique kevery with ims per remoter connnection
         rvy = Revery(db=hab.db)
         self.kevery = Kevery(db=self.hab.db,
+                                      cf=self.hab.cf,
+                                      enableKram=True,
                                       lax=False,
                                       local=False,
                                       rvy=rvy)
--- a/src/keri/core/parsing.py
+++ b/src/keri/core/parsing.py
@@ -1406,12 +1406,11 @@ class Parser:
                 Ilks.qry, Ilks.rpy, Ilks.exn, Ilks.xip, Ilks.pro, Ilks.bar
             ):
                 kwa = dict(exts)
-                kwa.pop('serder', None)
                 kwa['rvy'] = rvy
                 kwa['exc'] = exc
                 kwa['tvy'] = tvy
                 try:
-                    kvy.processMsg(serder=serder, kwa=kwa)
+                    kvy.processMsg(kwa=kwa)
                 except AttributeError as ex:
                     raise ValidationError(f"Error while processing msg in Kevery"
                                                 f"= {serder.pretty()}.") from ex
--- a/src/keri/core/parsing.py
+++ b/src/keri/core/parsing.py
@@ -12,7 +12,7 @@ from base64 import urlsafe_b64encode as encodeB64
 
 from hio.help import ogler
 
-from ..kering import (Colds, sniff, Vrsn_2_0, Version, Ilks,
+from ..kering import (Colds, sniff, Vrsn_2_0, Ilks,
                       UnexpectedCountCodeError, MaterialError, ValidationError,
                       QueryNotFoundError, ExtractionError, ShortageError,
                       ColdStartError, InvalidVersionError,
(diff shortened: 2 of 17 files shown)


KeaxD:
test_join_group_hab_persists_group_name_on_reload in test_habbing_v2.py was causing problem because of db dir path

src/keri/app/directing.py
```diff
@@ -490,9 +490,11 @@ def __init__(self, hab, remoter, verifier=None, exchanger=None, doers=None, **kw
...
         rvy = Revery(db=hab.db)
         self.kevery = Kevery(db=self.hab.db,
-                                      lax=False,
-                                      local=False,
-                                      rvy=rvy)
+                             cf=self.hab.cf,
+                             enableKram=True,
+                             lax=False,
+                             local=False,
+                             rvy=rvy)
```
ryan-hansen, line 497: So you're enabling KRAM for the TCP path (Reactant), but I don't see any KRAM for HTTP, setupWitness sets up its own kevery, so KRAM would need to be enabled there too.

src/keri/core/parsing.py
```diff
@@ -1402,6 +1402,21 @@ def msgProcess(self, exts, kvy, tvy, exc, rvy, vry):
         if isinstance(serder, SerderKERI):
             ilk = serder.ilk  # dispatch abased on ilk
 
+            if kvy is not None and serder.pvrsn.major >= Vrsn_2_0.major and ilk in (
+                Ilks.qry, Ilks.rpy, Ilks.exn, Ilks.xip, Ilks.pro, Ilks.bar
+            ):
+                kwa = dict(exts)
+                kwa.pop('serder', None)
```
ryan-hansen: It feels like there should be a cleaner way to do this than copying exts to kwa and then popping serder off.  exts already includes serder, which is where the serder variable is coming from in this scope.  Maybe instead of serder = exts['serder'], do something like this:
```
kwa = dict(exts)
serder = kwa.pop('serder') 
```
Then inside the version check do this:
```
kwa.update(rvy=rvy, exc=exc, tvy=tvy) 
kvy.processMsg(serder=serder, kwa=kwa)
```
Not the biggest deal.  Just looks kinda weird and a bit confusing at first why we're assigning serder from exts, then copying exts to kwa, then popping serder off kwa.

ryan-hansen (changes requested):
I'm still reviewing, but here are a couple of notes to start with.

src/keri/core/parsing.py
KeaxD: Since kwa is only used in the version block, it seemed clearer to me to keep it there rather than create it at a higher level.

SmithSamuelM:
Orginally Serder was a top level parameter from the Parser.  When I split the parser from the processor, I put serder as one of the parsing artifacts in exts. This was to avoid the naming conventions errors that had crept into parsing artifacts. The artifacts are now typed and defined in the dataclass MsgParseDom.   

But message processing was not changed since in many cases does not use all the fields in MsgParseDom and it takes advantage of the python ** packing/unpacking for method/function parameters.  So MsgProcess gets a dict version of MsgParseDom and serder is one of those parameters.

Calls inside of MsgProcess usually either just take an unpacked parameter list the explicitly declares those parameters that are used and collects the rest in  a packed **kwa.  The later is to prevent errors for unused parameters and the former is to explicity make parameters required..  Now downstream of this some functions then put serder backinto a dict kwa like to pass down.  The solution in this case is to examine why they put it back into a dict instead of using it as a top level parameter in that function call.  As a top level unpacked parameter it becomes "required" which why one does it that way and packed parameters are optional or potentially optional.

This is where the problem is, not at the top level but at the lower level.    This is also a Pythonic thing.  Parameter packing and unpacking and required version optional  via * and ** is a super power but its easy to misuse it or defeat it..

This tutorial does a good job of explaining the 5 types of parameters and how and why to use them.

https://levelup.gitconnected.com/5-types-of-arguments-in-python-function-definition-e0e2a2cafd29


I did not have the time to refactor Parser.msgProcess to clean up the call signatures. I left it as is.  Since the parser doesn't know which fields in ext to msgProcess are required and which ones are not, it just passes in exts.  Before exts was a dict, so the call to msgProcess converts exts the dataclass into exts a dict. So zero change to the ensuing code.     Now if one were to refactor the code in msgProcess and discover the serder is a required parameter then one could pull serder out before calling msgProcess. But its the same effort to pull it out after calling.    Some of the downstream usages want to tunnel sets parameters and so a common pythonic approach to tunneling parameters sets is to use ** pack/unpack to do it.  The packing operator is optimized so the performance hit is minimal. Its more about intent. Are you tunneling parameters through nested method calls? If so then it gets cumbersome to expose those parameters so tunneling them by packing is less cumbersome.  It does hide the required/optional stuff so its a tradeoff.   Tunneling obfuscates the semantics of the parameters so if the set is small and the nesting is not two deep, one should consider not tunneling. Whereas if the set is very large and nested multiple levels then tunneling might net out better. In the case of exts the set is very large and is nested in the parse sub-methods, so it would be cumbersome to not tunnel. Indeed, to enable dispatch I created a method dispatcher which only works well if the parameters are tunneled. So exts in parser. But once msgProcess gets the exts dict, it now becomes more use case specific whether to continue to tunnel or not.  

When I use the term tunnel I mean to pack parameters into a dictionary and pass down the dictionary. The call siganture of the called (pass down) function can choose to unpack or keep it packed and pass it down another layer. If it keeps it packed then its effectively tunneling the parameters through itself. It is not unpacking them.


FYI an anti pattern for packing is to be less verbose when multiple function calls use the same parameter values (not tunneling but to save on copy/paste). I noticed this in some of the v1/v2 unit tests where instead of duplicating explict sets of parameters, a shortcut was used to create a dictionary of those parameters and then pass in the dictionary everywhere. The number of parameters was small a handful at most.  This obfuscated the parameter semantics, so one has to go look up the packed dictionary, and it's not tunneling lower. So the net effect is harder to read code to save some copy/paste of short parameter lists. Don't know if that was an AI thing. But its an anti-pattern. I let it go at the time. But since I am discussing it here I thought I would bring it up now.

ryan-hansen (approved):

PR #1487, opened 2026-07-10
Log kli witness start failures and prevent startup stalls (#238) — v1.2 backport

### LINKED ISSUE #238: Ensure `kli witness start` logs errors correctly

dhh1128:
## Summary

Backport of #1477 to the **1.2 line** (base `v1.2.14`). `kli witness start` could
stall during startup and did not reliably log error conditions (#238). This PR
makes the command log every startup/runtime failure visibly, eliminates the
interactive-prompt stall, and deprecates the mis-named `--logfile` in favor of the
accurate `--logdir`. Fixes #238.

The change is split into two commits so the CLI rename can be dropped
independently:

1. **Log failures + prevent non-TTY stalls (#238).** Startup/runtime failures are
   logged at CRITICAL with a traceback (so they surface at the default
   `--loglevel`), the `Habery` is closed in a `finally` so a failed start leaves
   no stale LMDB lock, and an encrypted keystore started with no passcode on a
   non-TTY fails fast with a logged `AuthError` instead of blocking on an
   interactive `getpass`. `--loglevel` is normalized with `.upper()` so a
   lowercase level (e.g. `debug`) is honored.
2. **Deprecate `--logfile` in favor of `--logdir` (#238).** The logging backend
   derives the file name from `--name` and the subdirectory from its own prefix,
   so only the directory is meaningful; the old code mis-assigned the whole file
   path as the log head directory. `--logfile` is retained as a deprecated alias
   whose directory portion is used, with a stderr deprecation notice. Droppable
   if you'd rather not change the CLI surface in a patch release.

Unlike the 1.3 backport, no boot-Keeper `ks.close()` commit is needed — the 1.2
line already has that fix.

## Tests

Regression + unit tests added to `tests/app/cli/test_kli_commands.py`, covering
CRITICAL failure logging + `Habery` cleanup, the non-TTY fast-fail (both the
no-passcode and wrong-passcode paths), `--loglevel` normalization, and
`--logdir`/`--logfile` directory handling + the deprecation notice. Each
assertion was mutation-checked — reverting the corresponding fix turns the
targeted test red.

## Notes

- **Version:** left `__version__` untouched; happy to bump to `1.2.14` or leave
  the release cut to you — your call.
- Pre-existing, out of scope: `runWitness`'s `aids = cf.get("aids", default=None)`
  would raise on this branch if a `--config-file` is actually loaded, because
  1.2.13's `Configer.get()` takes no `default=` kwarg. This PR doesn't touch that
  behavior (it only moves the line inside the new `try`, so it would now be logged
  at CRITICAL rather than surfacing bare). Flagging in case you want a follow-up.

## Relationship to #1477

Same fix as the approved #1477 (base `main`, 2.0), adapted to the 1.2 CLI layout
(`keri.app.cli.*`, absolute imports, `help.ogler`, the `aids` setup path). Opened
per your request on #1477.

---

This code was developed with Claude (using the Opus 4.8 model, mostly). However,
I (Daniel the human) have read, reviewed, and worked the code, and I stand behind
it and take responsibility for it.

### DIFF
--- a/src/keri/app/cli/commands/witness/start.py
+++ b/src/keri/app/cli/commands/witness/start.py
@@ -7,11 +7,15 @@ Witness command line interface
 """
 import argparse
 import logging
+import sys
 
 from keri import __version__
 from keri import help
+from keri import kering
 from keri.app import directing, indirecting, habbing, keeping, configing
 from keri.app.cli.common import existing
+
+logger = help.ogler.getLogger()
 
 d = "Runs KERI witness controller.\n"
 d += "Example:\nwitness -H 5631 -t 5632\n"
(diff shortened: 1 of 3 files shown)


kentbull (approved):

kentbull:
I reviewed both this PR and the one to `main`. They are ready to go.

PR #1492, opened 2026-07-13
Added V2 support for /multisig/

KeaxD:
- Added V2 support for /multisig/icp by routing explicit V2 embeds through exchange() with nested substreams instead of V1 specialExchange.
- Kept legacy behavior intact: default /multisig/icp wrapping still uses V1 specialExchange
- Updated BaseHab.exchange() to convert embedded inner message streams into V2 nested substreams.
- Updated MultisigNotificationHandler to forward V2 nested payloads into Multiplexor.
- Updated Multiplexor.add() to support both:
      - V1 embedded payloads from ked["e"]
      - V2 multi-embed nested payloads from nests


KeaxD:
- Added V2 support for core multisig exchange wrappers:
  - `/multisig/ixn`
  - `/multisig/rpy`
- Added a signed V2 embed manifest in outer `a["embeds"]` to preserve:
  - embed label -> child SAID mapping
  - aggregate manifest SAID via `a["embeds"]["d"]`
- Preserved V2 nested payloads through the parser/KRAM/exchanger path into multisig handlers
- Added V2 wrapper-shape tests for:
  - `multisigInceptExn`
  - `multisigRotateExn`
  - `multisigInteractExn`
  - `multisigRpyExn`
- Added V2 KRAM-path handler tests for:
  - `/multisig/icp`
  - `/multisig/rot`
  - `/multisig/ixn`
  - `/multisig/rpy`
- Cleaned up comments and docstrings to match the new V2 nested-substream workflow.

src/keri/app/grouping.py
```diff
@@ -300,18 +301,6 @@ def loadHandlers(exc, mux):
...
-
-    `specialExchange` still builds the legacy embedded `/multisig/*` EXN
-    shape, so keep that body at v1 while allowing the surrounding stream
-    framing to follow the explicit caller version or the global default.
-    """
-    gvrsn = version if version is not None else Version
-    kind = kind if kind is not None else Kinds.json
-    return dict(version=Vrsn_1_0, kind=kind), gvrsn
-
-
```
ryan-hansen, line 317: Be careful removing this without making sure some version is passed.  The multisig functions (e.g. multisigRegistryInceptExn) set the default version and kind to None, which gets passed to specialExchange.  So if the callers don't explicitly set the version and kind, they get set to None, not to the default.

src/keri/app/grouping.py
```diff
@@ -440,13 +487,13 @@ def multisigRegistryInceptExn(ghab, usage, vcp, anc, version=None, kind=None):
         anc=anc
     )
 
-    kwa, gvrsn = _exnVersion(version=version, kind=kind)
     exn, end = specialExchange(sender=ghab.mhab.pre,
                                route="/multisig/vcp",
                                attributes={'gid': ghab.pre, 'usage': usage},
                                embeds=embeds,
-                               **kwa)
```
ryan-hansen, line 451: Group multisig on V2 should stop using specialExchange embeds and use nested substreams via hab.exchange() instead. The PR does that for icp/rot/ixn/rpy, but vcp/iss/rev/exn still always call specialExchange, so this also applies to multisigIssueExn, multisigRevokeExn, and multisigExn.

ryan-hansen (changes requested):

src/keri/app/grouping.py
KeaxD: vcp/iss/rev/exn are for ACDC/IPEX, according to the issue created at Keri Foundation:

"Exchange messages used for IPEX/ACDC can wait until V2 ACDC registry support exists", did we add that support?

src/keri/app/grouping.py
KeaxD: version and kind are already hardcoded in specialExchange so not passing anything would be fine

src/keri/app/grouping.py
ryan-hansen: Oh right...my bad.

src/keri/app/grouping.py
ryan-hansen: That's only true if the arguments are not passed at all.  But in the case of multisigRegistryInceptExn (for example), its defaults for version and kind are None.  So inside those multisig functions, when it calls specialExchange (and endorse), it will call them with version, kind, and gvrsn all set to None, overriding the defaults.  _exnVersion used to normalize that, but with that function removed, that will no longer happen. 

Does that make sense?

KeaxD: oh yeah I see, so we could just not pass version and kind and let special exchange handle those. At least until we come back and implement the v2 path, then we could add it back like the other handles

dhh1128 (changes requested):
A second set of eyes on top of the review already on this PR, not a rehash of it. Moving /multisig coordination off v1 e-embeds onto real CESR v2 nested substreams is the right target, and gating v1 specialExchange behind an explicit version check is the right shape.

The thing I think has to be settled first is the default version, which flips to v2 and so contradicts the description's claim that the default stays on v1. In `multisigInceptExn` / `Rotate` / `Interact` / `RpyExn` (src/keri/app/grouping.py) the branch is `if version and version.major == Vrsn_1_0.major ... else Version`; the global `Version` (src/keri/kering.py:24) is major 2, and `cli/commands/multisig/incept.py:169` passes `version=None`, so `None` falls into the `else` and emits the v2 manifest + nested-substream exn. The test rename to `..._explicit_v1` (now needing `version=Vrsn_1_0` to get the old shape) points the same way. Was a global v2 default intended here, or should these routes pass v1 explicitly until the rest is ready? Most of what follows only matters once that's decided.

Two more that seem as serious.

The v2 shape is a cross-implementation contract with no spec I could find. The SAIDified `a.embeds` manifest built in `BaseHab.exchange()` (habbing.py), the nests forwarded in `peer/exchanging.py` (`if route.startswith('/multisig') and incomingNests`), and `Multiplexor.add()` keying on `ked['a']['embeds']` together define an application-layer shape above the CESR version string that keripy, KERIA, signify-ts, and SignifyPy all have to reproduce identically — and multisig groups mix implementations routinely. With the default flipped, a keripy member on v2 and a signify-ts member on v1 can't complete a joint inception. I can't see signify-ts or KERIA from here, so I'm asking: is there a shared spec for this shape, or a defined behavior for a reader that doesn't understand it?

The signed manifest isn't bound to the substreams it commits to. The outer exn signs `serder.raw` (including the `a.embeds` manifest of inner SAIDs), but the nests travel as CESR attachments outside `serder.raw`, and the v2 branch of `Multiplexor.add()` (grouping.py ~756-768, ~836-842) parses them without checking `nest.serder.said == manifest[label]`. A relay or group member can keep the signed exn and manifest and substitute other validly group-signed nested events; aggregation stays keyed on `esaid = manifest['d']` and nothing rejects the swap. This doesn't put a forged event into any KEL, since inner signatures still gate Kevery, so it's coordination confusion rather than an authority bypass — but a `said`-in-manifest check plus a swap test in `Multiplexor.add()` would close it.

Smaller things, in rough order:

- The new v2 tests (tests/app/test_grouping.py, tests/peer/test_exchanging.py) only cover the accept path; a mismatched or swapped nest is the case that would have caught the default flip.
- `Multiplexor.add()` keys aggregation on the group-event SAID for v1 (`embed['d']`) but on the manifest SAID for v2 (`manifest['d']`). In a mixed-version group the two halves key differently, never reach threshold, and stall with no signal.
- The manifest and nests machinery handles N embeds, but every migrated route (icp/rot/ixn/rpy) carries exactly one, and the multi-embed routes (vcp/iss/rev) stayed on v1; the `len(nests)==1` branch already resolves the SAID without the manifest. Is the manifest carrying its weight yet?
- `serializeParsedSubstream` (habbing.py) handles roughly eleven attachment families but the exercised path only uses indexed sigers. The unexercised branches are where CESR self-framing tends to mis-frame silently across implementations; scoping it to the families that have an emitter, or a per-family round-trip test, would help.
- Leaving the credential routes (vcp/iss/rev) on v1 means a v2 vLEI issuer still hits v1 specialExchange on the path it needs most. That reads like a deliberate KEL-lifecycle-first slice; a line in the description on the sequencing would save the next reader the trace.

One thing I'm unsure about underneath all of this: the parallel v2 path assumes a v2 event can't ride inside the v1 `e` embed. I couldn't find where that's shown, and e-embedding looked orthogonal to the inner event's version to me. If a v2 event does round-trip through the v1 embed with signatures intact, a second permanent wire representation across four implementations is a lot to maintain; if it genuinely breaks, a short note on what breaks would anchor the rationale.

KeaxD:
@dhh1128 The first point is mostly a clarification of intent: the default to V2 on the migrated multisig routes is intentional. `version=None` is meant to fall through to global `Version`, and the legacy V1 `specialExchange(...)` path is only selected when `version=Vrsn_1_0` is passed explicitly. So the `...explicit_v1` test rename was deliberate. If the branch description still implies “default stays on v1,” that description is stale and I’ll update it.

On `a["embeds"]`, this was added intentionally for the V2 envelope and not just as extra redundancy. In V1 the child payloads lived in body field `e`. In V2 the child message bytes move into `nests`, so I it made sense to me that we needed a signed place in the outer exn to preserve the labeled embed mapping. `a["embeds"]` carries `label -> child SAID`, and `a["embeds"]["d"]` is the aggregate manifest SAID for the whole labeled set. That preserves named embeds like `icp`, `rot`, `ixn`, `rpy`, etc., and gives receive-side multiplexing one stable digest analogous to V1 `e["d"]`. 
Although this is solely a design choice that I added, so I am unsure about how it would fit in the overall architecture. After reading your review, I think it should be reverted, since this PR primarly focuses on /multisig/icp, /multisig/rot, /multisig/ixn, /multisig/rpy, those routes only carry one child so the route itself already tells you the label. Then for multi-child, we could keep the manifest

Credential routes remaining on V1 is deliberate; there is still some work that was planned before changing those routes. Ryan also had the same concern in his earlier comment on this thread.

KeaxD:
I narrowed this PR to the current single-child V2 multisig routes only: `/multisig/icp`, `/multisig/rot`, `/multisig/ixn`, and `/multisig/rpy`.

Those routes now use the simpler V2 shape of signed child SAID in `a["d"]` plus one nested substream, and `Multiplexor.add()` now validates that binding on receipt. 

For aggregation, legacy V1 messages still key by `e["d"]`, and the new single-child V2 messages synthesize that same legacy key so V1 and V2 proposals for the same payload land in the same bucket.

I also removed the generic manifest-based V2 embed path from `BaseHab.exchange()`. Multi-child V1 routes are unchanged, and broader join/CLI workflow changes were left out for a separate PR.

KeaxD:
@dhh1128  So after thinking about the aggregation key for v1 and v2, I think it would be fair to think that it is fine for v2 to have a different aggregation key from v1. Currently the v2 aggregation key is derived from v1 which would allow for v1 and v2 to have the same key. But a mixed version exchangeg doesn't make sense, if a participant in a multisig is v1, v2 participants should simply emit v1 messages, if they are all v2, they should all emit v2. There is no need to support mixed version exchange as far as I understand.

SmithSamuelM:
Its only with v2 exchange messages that we plan to anchor exchanges. We really don't want to be anchoring v1 exchanges.  It is incumbent on the community to upgrade their production branches to supportk v2 for IPEX exchanges. Anchoring v1 exchanges would just create a mess where now exchnages in v1 had to live forever. We want to avoid that.

KeaxD:
I changed multisig wrapper versioning into explicit `pvrsn` and `gvrsn` parameters, so the outer message/body version and the attachment genus can be controlled independently instead of being implicitly tied together.

On the receive side, `Multiplexor` now keys wrapper handling off the exn body version (`pvrsn`) and preserves the intended route-specific behavior:
- V1 wrappers continue to use the legacy embedded `e["d"]` proposal identity
- current single-child V2 wrappers use the signed child SAID from `a["d"]`
- approved replay of legacy embedded children now chooses the parser from the child attachment framing, so V1 embedded flows still work when the child attachments are V2-framed

For v1 replay path in Multiplexor, the parser version is defined inside multiplexor not at initialization, for every embed we check the version.

A v2 parser would work for:
- v1 child body + v2 attachments
- v2 child body + v2 attachments
but it has trouble doing v1 child body + v1 attachments

dhh1128 (commented):
Most of what I raised last time is answered — the `a.embeds` manifest is gone, the single child
binds through `a.d`, `Multiplexor.add()` checks that binding, and there are reject-path tests now.
Thanks for the rework. I also went and checked the thing I said I couldn't see last time, and it
cuts in your favor on the main design question: the v2 spec fixes the `exn` top-level fields at
`[v, t, d, i, ri, x, p, dt, r, q, a]` and says "No other top-level fields are allowed (MUST NOT
appear)" (`kswg-keri-specification` spec-body.md:1197). So dropping `e` isn't a preference, it's
required, and my round-1 question about whether a v2 event could just ride inside the v1 embed is
answered — it can't, not conformantly.

What follows is new ground, a level down in the CESR plumbing.

I think what you had before commit `ab63a550` was actually the stronger half of a two-part design;
with that commit, I think you've kept the lesser part. I also think the genus sniff may break some
production callers.

The pattern repeated at `src/keri/app/grouping.py:359, 444, 528, 722, 971` decides which code table
to parse a child's attachments with by trying `Counter(qb64b=atc, version=Vrsn_1_0)` and treating a
success as "v1". As far as I can tell that isn't a discriminator. Every v1 counter code point also
exists in the v2 table, and nineteen of them mean something different there at identical `Cizage` —
`-C` is `NonTransReceiptCouples` in v1 and `AttachmentGroup` in v2, `-B` is `WitnessIdxSigs` vs
`BodyWithAttachmentGroup`, and so on. A successful v1 parse tells you nothing. It reads even
`-_AAACAA`, the genus code that encodes the answer, as v1.

Here is where I worry that this might cause problems. `msgOwnEvent()` and `msgOwnInception()` default to `framed=False`, and
with `framed=False` the child's attachments open with `-CAX`, the v2 `AttachmentGroup`. The sniff
picks v1, `Parser(version=Vrsn_1_0).parse()` returns `None`, and the builder raises
`ValueError("V2 multisig payloads must be framed CESR streams")`. The stream is a perfectly good
framed CESR stream — `Parser(version=Vrsn_2_0)` on the same bytes returns one message — so the
error misdescribes what went wrong. I ran this against your branch head to be sure I wasn't
misreading:

```
framed=True   atc=b'-KAWAABAk07SlJEo'  -> sniff V2  -> OK
framed=False  atc=b'-CAX-KAWAABAk07S'  -> sniff V1  -> ValueError
```

`cli/commands/multisig/incept.py:165` and `join.py:243`, `:449`, `:602` all rely on that
`framed=False` default, and `incept.py:172` passes `version=self.inits.get("version")`, which is
`None` for a v2 config and so takes the new branch. `cli/commands/ends/add.py:113` reaches
`multisigRpyExn` the same way. Every v2 test in the PR passes `framed=True` explicitly, which is
why CI stays green.

I don't think this is a tuning problem, because the spec defines the genus version so that trial
parsing can't recover it. CESR says a genus/version code is `-_GGGVVV`, where `GGG` is the genus and
`VVV` is "the Version of that protocol genus", and it fixes what a version bump means: "A Major
change occurs when a code's meaning changes", while a minor change is a code being added
(CESR spec-body.md, Protocol genus/version codes). Genus version 1 to 2 is a major bump, so code
meanings are *supposed* to have changed underneath the same characters. A successful parse against
the 1 table is therefore evidence of nothing, and that's by construction rather than by accident.

The other half is that the two version numbers aren't the same kind of thing. The v2 version string
is `PPPPMmmGggKKKKBBBB.`, where `Mmm` is "the version of the KERI protocol used by the associated
message" and `Ggg` is "the major and minor version numbers of the Version of the CESR genus table
used in the message" (KERI spec-body.md, Version string field; CESR spec-body.md, Version 2.XX
string field format). One versions the field maps, the other versions the code tables, and they move
on separate schedules. The legacy v1 string, `PPPPvvKKKKllllll_`, has no `Ggg` field at all — so a v1
body structurally cannot state its genus version, while a v2 body always can:

```
body pvrsn=1, gvrsn asked=1 -> serder.gvrsn=None  atc=b'-AABAAAs'   (v1 tables)
body pvrsn=1, gvrsn asked=2 -> serder.gvrsn=None  atc=b'-KAWAAAs'   (v2 tables)
body pvrsn=2, gvrsn asked=1 -> serder.gvrsn=(2,0) atc=b'-KAWAABA'   (clamped up)
```

That makes `serder.gvrsn if serder.gvrsn else serder.pvrsn` right in its first half and, I think,
substituting the wrong quantity in its second. For a v1 body the answer isn't in the body; falling
back to the protocol version is a keripy convention (`messagize`'s `gvrsn = max(svrsn, gvrsn)`,
`core/eventing.py:1492`), not something the spec supports. Rows one and two above differ only in
their attachments, and your own `test_multisig_incept_v1_body_supports_v2_attachments` covers row
two. This looks like the same soft spot as the v1-body/v1-attachment gap you mentioned in the thread.

The spec does give a place to put the answer for exactly that case. `-A`, `-B` and `-C` are the three
universal count codes that permit a genus/version override: "Should the first Group Code embedded in
each of these groups be a genus/version code, then the parser MUST switch code tables" (CESR
spec-body.md, Universal Code table genus/version codes that allow genus/version override). You're
already emitting the child in a `-B` group, so a v1-genus child could carry `-_AAABAA` as the first
element inside it and be unambiguous to any conforming parser. Nothing on these paths sets
`genusify`, so today no genus code is emitted at any level and a receiver has to already know.

I'd flag that keripy's vocabulary doesn't help here: `kering.py:24` defines one `Version` as "Default
KERI Protocol, CESR Genus Version" and comments `Vrsn_1_0` as "Protocol/Genus Version 1 Specific", so
a single constant does both jobs. That's harmless while KERI 2.0 and CESR genus 2.0 coincide, and
this PR follows the house convention rather than inventing anything. It stops being harmless at the
v1 body, which is the case the four builders have to handle. Which is also why I wondered about
`ab63a550`: `57caf7a4` introduced separate parameters whose docstrings name the two axes exactly
("optional explicit wrapper protocol version", "optional explicit attachment genus version")
alongside the inference, and a day later the explicit half was dropped and the inferred half kept.
The single `version` that replaced them now feeds `version=` on the body, `gvrsn=` on the endorse,
and `gvrsn=` on `serializeParsedSubstream`, so a caller who does know the child's genus version has
no way to say so. Was the revert about call-site churn, or did the split cause a problem I'm not
seeing?

Separately from all that, `ShortageError`, `UnexpectedOpCodeError` and `KeyError` all escape the
`except UnexpectedCodeError`, and the copy at `:971` sniffs bytes recovered from a remote peer's exn.

Here are two other points to ponder.

The binding check you added covers four routes, and it looks to me like the `else` falls open. At `grouping.py:898-907`,
`signed` comes from `payload.get("d")`; the `raise` fires only when `route in ("/multisig/icp",
"/multisig/rot", "/multisig/ixn", "/multisig/rpy")`, and the SAID comparison is guarded by
`if signed is not None`. A v2 exn on `/multisig/vcp`, `/iss`, `/rev` or `/exn` with no `d` in `a`
therefore skips both checks and falls through to `esaid = nserder.said`, a value the sender chose.
`loadHandlers` registers all eight routes and `exchanging.py:156` forwards nests for anything under
`/multisig`, so the branch is reachable; the `["", "multisig", *_]` case only asks that `gid` be a
group we're in. Since the producers for those four routes are pinned to `version=Vrsn_1_0`
(`:594, 633, 671, 782`), nothing honest ever emits v2 there — which argues for the v2 branch simply
refusing a route it has no producer for, rather than for extending the tuple.

And `nserder.said` isn't always a verified digest. `Serder.said` at `serdering.py:2087` returns the
raw `_sad['d']` when the ilk has no `saids` entry, which is the case for `rct` (`serdering.py:453`).
The nest travels as unsigned attachment bytes, so a relay can drop the real child and substitute a
`rct` whose `d` field is literally the value in `a.d`; `len(nests)` is still 1 and the equality still
holds. Asserting the nested ilk is saidive, or verifying with `Diger(qb64=signed).verify(ser=nserder.raw)`
instead of comparing strings, would close it. Strip and append-a-second both already fail closed, so
this looks like the only remaining gap in that check.

Smaller things, in rough order:

- I don't think either nesting mechanism in the tree is right on its own, which is an argument for
  merging them rather than for preferring #1517. `_normalizeNestedStream` (`acdc/ipexing.py:108`)
  preserves the child's bytes and reads the leading counter against a named set of `Codens`, which
  is better than guessing — but splice a fully-v1 child into a v2 `-B` group and a v2 parser reads
  the v1 `-A` as `GenericGroup` and returns nothing, silently. Your parse-and-reserialize handles
  that v1 child correctly and breaks on the v2 `-C` shape instead. A shared helper that decides genus
  from the body, preserves bytes when the child is already v2, and transcodes only when it's v1 would
  cover both. `app/habbing.py` seems like the wrong home for it either way — there's no `Hab`
  dependency, and it makes `app/grouping.py` import the heaviest module in the tree. Next to
  `messagize` in `core/eventing.py` might fit better.
- `a["d"]` for the child's SAID reads oddly next to the convention that `d` inside a block is the
  SAID of that block — `exchange()` does exactly that to `e` at `exchanging.py:424`. signify-ts has
  already committed to the other reading: `MultisigIcpExchange` types the child as
  `e: { icp: Icp | Dip; d?: string }` (`signify-ts/src/keri/app/exchanging.ts:40-43`), so over there
  `d` is the optional block SAID and `icp` names the child. `a["icp"]` / `a["rot"]` / `a["ixn"]` /
  `a["rpy"]` would match that and #1517's `a.acdc` / `a.iss` / `a.anc`, and makes their migration
  mechanical. For `/multisig/icp` the current field is also degenerate — I get `a["d"] == a["gid"]`,
  since an icp's SAID is the AID. It's a cheap change now and a much less cheap one once something
  else implements the shape.
- Nests aren't persisted. `Exchanger.logEvent` (`exchanging.py:276`) stores the exn, sigs and `epath`
  but has nowhere to put nests, and `escrowPSEvent(serder, tsgs, pathed)` drops them too. A v2
  `/multisig` exn that escrows partially-signed and is released later replays without nests and hits
  your own `count=0, expected 1`. That may be what keri-foundation #62 is already tracking from #1517.
- `serializeParsedSubstream` covers every field of `MsgParseDom` today, so nothing is dropped right
  now — but it's an allowlist snapshot of a dataclass that grows by minor bump, and the fail direction
  is silent. Two things I'd look at while it's open: `frcs` is rejected even though `messagize`
  handles FirstSeen bonds, so a child replayed from a witness can't be wrapped at all; and the call
  in `Multiplexor.add` (`:986`) happens after the `meids`/`maids` writes at `:946-947`, so a rejected
  family leaves the buckets half-updated.
- The `nests` plumbing in `BaseHab.exchange` looks dead: `nests = None` at `habbing.py:1696` is never
  reassigned, there's no `nests` parameter on the method, so `eframed` at `:1709` always equals
  `framed`. The description still says this method converts embeds into v2 substreams, and also that
  the default `/multisig/icp` wrapping stays on v1 specialExchange and that multi-embed payloads are
  supported — none of which matches HEAD, and the last two you've already corrected in comments.
- `Saider`, `Saids` and `Serder` were added to the `habbing.py` import at `:25` but aren't used.
- `test_multisig_incept_default_framing_uses_default_version_with_legacy_special_exn` now passes
  `version=Vrsn_1_0` explicitly, so the name promises more than the body checks — and the default-flip
  guard the old version of that test provided is gone.
- The `if route.startswith("/multisig")` at `exchanging.py:155` puts an app route into the generic
  exchange layer. The side effect is that #1517's own IPEX routes lose their nests at that boundary
  with no log, which will matter as soon as #62 adds validation there.

One that's pre-existing and probably its own issue, though this PR is rewriting the function:
`approved = any([True for sub in submitters if sub.qb64 in self.hby.kevers])` at `grouping.py:953`
reads to me like it wants `hby.habs`. `submitters` always contains the sender, and `processEvent`
has already required the sender to be in `kevers`, so I think `approved` is true on the first remote
proposal and the notify-for-approval branch below it never runs. The member checks at `:915` and
`:924` look like the same substitution. I might be misreading the intent — is `kevers` deliberate
there?

Last thing, and I raise it only because it changes with the default rather than with the shape.
KERIA at `src/keria/app/grouping.py:183` does `esaid = exn.ked['e']['d']` unconditionally, and
`Vrsn_2_0` doesn't appear anywhere in its `src/`. With the default flipped, that's a `KeyError` on
the first v2 proposal rather than a graceful "I don't understand this". I take Sam's point that the
community has to move to v2 for anchored exchanges, so I'm not arguing the direction — but keeping
these four routes explicit-opt-in until KERIA and signify-ts can read the shape would maybe
be desirable? Should we raise corresponding PRs there?

KeaxD:
- Adds durable storage for V2 nested EXN child substreams via `enst`, and wires it through exchange logging, partial-signature escrow, replay, DB clone/copy, and `serializeMessage`.
- Updates multisig V2 handling so `/multisig/icp`, `/rot`, `/ixn`, and `/rpy` require exactly one nested child, bind it by SAID in `a.d`, validate the child ilk/digest, and replay stored approved children so multisig signatures merge correctly.
- Keeps CLI wrappers V1-only by explicitly passing `version=Vrsn_1_0` for CLI `multisig*Exn(...)` calls. The CLI gets only the minimal local-approval seeding needed for script compatibility.
- Adds IPEX V2 nested artifact validation so offer/grant bodies must match the nested ACDC/TEL/KEL artifacts they claim to carry.
- Change the shape of the a field so it alligns with IPEX

ryan-hansen (approved):

PR #1497, opened 2026-07-14
Fix replay version handling and related tests.

ryan-hansen:
- baser clone functions version set to global default
- explicit version passed in where necessary in tests and other usages

All tests are now passing again.

#NOTE: This does NOT yet fix all v1 hard-coded sites (there's at least one more in Baser.clean), or fix the common args issue (separate PR to come).

### DIFF
--- a/src/keri/cli/commands/import_.py
+++ b/src/keri/cli/commands/import_.py
@@ -11,7 +11,6 @@ from hio.base import doing
 
 from ..common import Parsery, setupHby
 
-from ...kering import Vrsn_1_0, Vrsn_2_0
 from ...app import HaberyDoer
 from ...core import Parser
 
@@ -65,8 +64,7 @@ class ImportDoer(doing.DoDoer):
 
         with open(self.file, 'rb') as f:
             ims = f.read()
-            Parser(kvy=self.hby.kvy, rvy=self.hby.rvy, local=False,
-                   version=Vrsn_1_0).parse(ims=ims)
+            Parser(kvy=self.hby.kvy, rvy=self.hby.rvy, local=False).parse(ims=ims)
             self.hby.kvy.processEscrows()
 
         self.exit()
--- a/src/keri/cli/commands/import_.py
+++ b/src/keri/cli/commands/import_.py
@@ -11,6 +11,7 @@ from hio.base import doing
 
 from ..common import Parsery, setupHby
 
+from ...kering import Vrsn_1_0
 from ...app import HaberyDoer
 from ...core import Parser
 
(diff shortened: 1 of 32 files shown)


src/keri/db/basing.py
```diff
@@ -1890,7 +1897,7 @@ def cloneEvtMsg(self, pre, fn, dig, gvrsn=Vrsn_1_0, *, version=None):
         #return msg
 
 
-    def cloneDelegation(self, kever, gvrsn=Vrsn_1_0, *, version=None):
+    def cloneDelegation(self, kever, gvrsn=Version, *, version=None):
```
evanja57: I think that we should get the gvrsn from the kever instead of defaulting to the global version. Something like:
```Python
def cloneDelegation(self, kever, gvrsn=None, *, version=None):
    if version is not None:
        gvrsn = version
    if gvrsn is None:
        gvrsn = kever.serder.gvrsn or kever.serder.pvrsn
```
That way v1 callers don't accidentally try and get delegation attachments as v2 because of the global default.

src/keri/cli/commands/ipex/grant.py
```diff
@@ -127,7 +126,7 @@ def grantDo(self, tymth, tock=0.0, **kwa):
         msg = bytearray(exn.raw)
         msg.extend(atc)
 
-        parsing.Parser(version=Vrsn_1_0).parseOne(ims=bytes(msg), exc=self.exc)
+        parsing.Parser().parseOne(ims=bytes(msg), exc=self.exc)
```
evanja57: I think that because we are implementing acdc v2 with keri v2 in the acdc package, this old ipex code can stay hardcoded to v1. Also this eventually goes through specialExchange so it needs to stay v1. Same goes for the other ipex/ files.

src/keri/cli/commands/import_.py
```diff
@@ -65,8 +64,7 @@ def exportDo(self, tymth, tock=0.0, **kwa):
 
         with open(self.file, 'rb') as f:
             ims = f.read()
-            Parser(kvy=self.hby.kvy, rvy=self.hby.rvy, local=False,
-                   version=Vrsn_1_0).parse(ims=ims)
+            Parser(kvy=self.hby.kvy, rvy=self.hby.rvy, local=False).parse(ims=ims)
```
evanja57: if we are removing version from parser here we should add a version parameter to cli so that v1 consumers aren't automatically forced to v2

src/keri/db/basing.py
ryan-hansen: Good call.  Done.

src/keri/cli/commands/ipex/grant.py
ryan-hansen: Over-zealous search & replace.  Good catch.

src/keri/cli/commands/import_.py
ryan-hansen: Same search & replace.

PR #1505, opened 2026-07-15
Worked example of contractually-protected disclosure

dhh1128:
## What this adds

One new test file, `tests/acdc/test_clc_disclosure.py`: a worked, working example of **contractually-protected disclosure** — Chain-Link Confidentiality (CLC) realized as a credential rather than an out-of-band agreement. It complements `tests/acdc/test_examples.py` (registry lifecycle, graduated/selective disclosure, blindable registries) by combining several ACDC v2 features the existing tests do not yet exercise together — including **both** graduated-disclosure mechanisms, each on the credential it fits.

## The scenario

Alice, a Utah resident, holds two credentials under one holder AID: an **attributive** SEDI digital-identity credential (issuee plus individually-disclosable name, date-of-birth, state-endorsed photo, and residence blocks) and an **aggregative** age credential (boolean flags for the thresholds 13, 16, 18, 21, 55, 65). To enter a dance club that serves alcohol, she issues her own bespoke presentation ACDC to the club: it edges (I2I) to her two source credentials and carries the CLC terms — a Purpose clause, an anti-assimilation clause, and a statutory safe-harbor clause — in its Rules section, so the club is bound to those terms before it receives her photo and age proof.

## What it demonstrates

Told as five `pytest` functions (the file also runs standalone as a script):

- **Phase 1** — the two source credentials, each with the disclosure mechanism that fits it. sedi-id is **attributive**: the issuee is `a.i` and each attribute is an individually partially-disclosable nested block, so the holder reveals the **photo** block and withholds DOB/residence/name by compaction (`keri.core.Compactor`), every mix recomputing to the same committed section SAID. age is **aggregative**: an array of boolean flags where all thresholds are present, so revealing **over-21** discloses nothing about the others — the club cannot tell whether she is over 55 or 65. Both disclosures are tamper-evident.
- **Phase 2** — the **bespoke presentation ACDC**: I2I edges binding both sources to a single holder, a schema that pins the edge operator to `const "I2I"` (so the same-holder binding is schema-enforced), and the CLC terms in the Rules section. The issuee is resolved uniformly via `SerderACDC.iseaid` whether the far node is attributive (`a.i`) or aggregative (`A[1].i`). The bespoke references its sources; it does not restate the photo or the age proof.
- **Phase 3** — the **gated IPEX exchange** (apply → offer → agree → grant → admit). The club's `apply` is field-level (per-schema attribute paths, including the joining issuee); the `offer` carries no PII; the `agree` is signed; and a disclosure gate opens **only** for a valid, offer-binding, signed agree — a forged signature, a stray agree, or a signed spurn all keep it shut. The grant carries two disclosures (the identity photo, partial; the age over-21 flag, selective), and on receipt the club runs a real verification: self-verify the artifact, match the SAID it agreed to, and schema-validate.
- **Phase 4** — key-state accountability (a captured establishing event still verifies the agree after the club rotates its key) and "terms follow the data" (weakening a clause changes the SAID, and the club's same verification rejects it).
- **Cross-kind** — the Phase 1–3 invariants hold across JSON, CESR, CBOR, and MessagePack.

Every ACDC validates against a purpose-authored JSON Schema (Draft 2020-12), the source credentials are bound to real registries via `regcept`, and every pinned SAID is derived, not pasted.

## Resolved design decisions

Worked through with Sam's guidance in the thread below:

- **Section per credential.** The core SEDI identity is modeled **attributively** — a fixed, well-labeled field set gains nothing from the aggregate (unlabeled-position) form and loses clean label-based pathing and partial disclosure. The **aggregative** form is used where it earns its keep: the age-threshold flags, where hiding which thresholds are asserted is the whole point (a poor man's sparse Merkle tree, as mdoc models age proofs). This example therefore demonstrates both mechanisms on the credential each suits. (This reverses the original draft's aggregate-identity choice, per Sam's first-principles reconsideration in-thread.)
- **I2I to either section.** Supported; the issuee is resolved via `SerderACDC.iseaid` (attributive `a.i` or aggregative `A[1].i`, merged in #1506). `keri.vdr.verifying.verifyChain` is V1-era and predates aggregate sections, so a v2-native verifier uses `.iseaid`.
- **No presentation registry** for the one-time bespoke presentation — appropriate here. A high-stakes flow would populate `rd` to force verification against an issuee-controlled presentation registry (a follow-up example).
- **Data-structure-level IPEX** — the right altitude. `keri.vc.protocoling` is V1 and does not speak v2 messaging; the v2-native `keri.acdc.ipexing` is still a stub, so the exchange is built directly from `keri.core.exchange`. A companion example will re-run this scenario through that tooling once it lands.
- **Graduated-disclosure negotiation** (the `apply`'s field request) is illustrative here — a static, one-shot narrowing, not full back-and-forth. Its exact encoding and placement are being designed in [discussion #1512](https://github.com/WebOfTrust/keripy/discussions/1512).

## Scope and honest boundaries

Deliberately Habery-free, at the data-structure level:

- The live TEL/registry runtime (`keri.vdr`) is out of scope, so **revocation** of the source credentials is documented as the verifier's required 4th check — with the hooks asserted (each I2I edge names a registry-bound far node) — but not executed here. `test_examples.py` covers the registry lifecycle.
- The agree is signed but **not KEL-anchored** — a deliberate privacy choice (no correlatable log of every patron). Phase 4 records the honest cost: a captured establishing key makes the signature *checkable*, not provably *current at signing*, so this is durable verifiability under cooperative timing, **not** non-repudiation robust to key compromise. Establishing currency-at-signing requires ordering the signature against the KEL (anchoring, or a witnessed receipt); a high-stakes CLC flow would anchor — to a *blinded* presentation registry, to keep correlation-resistance.

## How to run

```
pytest tests/acdc/test_clc_disclosure.py
```

It also runs standalone as a script, and passes under `pytest -n auto`.

---

*This worked example was developed with AI assistance (Claude, Anthropic).*

### DIFF
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -79,12 +79,15 @@ NONCES = [Noncer(raw=raw).qb64 for raw in RAWS]
 def _saidify_schema(mad, kind=Kinds.json):
     """Compute a JSON Schema's SAID and return (said, schema-with-$id).
 
-    Mirrors how keri.acdc.messaging builds its default schemas: run the schema
-    map through a Mapper that self-addresses the '$id' field. The '$id' must be
-    the FIRST field (its position is part of the serialization the SAID digests),
-    so callers author the schema with "$id": "" up front. The returned schema
-    carries its SAID in '$id'; an edge that points at a credential of this schema
-    references that same SAID in its 's' field.
+    keri exposes public builders only for the *default* ACDC schemas
+    (acm/acg/actSchemaDefault), not a general "self-address an arbitrary schema"
+    helper, so this mirrors the pattern those builders use internally: run the
+    schema map through a Mapper that self-addresses the '$id' field. The '$id'
+    must be the FIRST field (its position is part of the serialization the SAID
+    digests), so callers author the schema with "$id": "" up front. The returned
+    schema carries its SAID in '$id'; an edge that points at a credential of this
+    schema references that same SAID in its 's' field. Mapper deep-copies its
+    input, so the caller's schema map is never mutated.
     """
     mapper = Mapper(mad=mad, makify=True, strict=False, saids={"$id": 'E'},
                     saidive=True, kind=kind)
@@ -305,6 +308,15 @@ SAFE_HARBOR_TEXT = (
 # Registry inception timestamps (the State and endorser each stand up a registry).
 REG_STATE_STAMP = "2026-01-05T12:00:00.000000+00:00"
 REG_ENDORSER_STAMP = "2026-01-06T12:00:00.000000+00:00"
+
+# A note on the pinned SAID values asserted throughout this file (credential,
+# registry, and exn SAIDs): every one is DERIVED here, not authored -- the tests
+# construct the object and pin the SAID it computes. They are stable because the
+# inputs are fixed (the salt-derived actor AIDs, the NONCES, these timestamps, and
+# the schema maps above). Change any of those and the dependent SAIDs change too;
+# regenerate them by running this file's __main__ and reading the failure diffs, or
+# print the .said of the object in question. This is the same "derive, don't paste"
+# convention as tests/acdc/test_examples.py.
 
 # Fixed timestamps for the IPEX exn messages. exchange() defaults the 'dt' field
 # to the current time, which would make the messages (and their SAIDs) vary run to
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -447,7 +447,7 @@ def _photo_disclosure(aggor):
     return disclosed
 
 
-def _bespoke_rules():
+def _rules_in_bespoke():
     """The bespoke ACDC's Rules section: Purpose + Assimilation + SafeHarbor.
 
     Mirrors the spec's bespoke-ACDC rule shape -- a top-level block plus one named
@@ -505,7 +505,7 @@ def _bespoke_presentation(sedi, over21, kind, compactify=False, rule=None):
                      over21=True)
     return acdcmap(israid=ALICE, uuid=NONCES[9], schema=schema, attribute=attribute,
                    edge=_bespoke_edges(sedi, over21),
-                   rule=rule if rule is not None else _bespoke_rules(),
+                   rule=rule if rule is not None else _rules_in_bespoke(),
                    kind=kind, compactify=compactify)
 
 
@@ -789,7 +789,7 @@ def test_gated_ipex_exchange_JSON():
                      attributes=dict(acdc=bespoke.said,
                                      credentials=[sedi.said, over21.said],
                                      governance=GOV_PROVISION_SAID,
-                                     terms=_bespoke_rules()),
+                                     terms=_rules_in_bespoke()),
                      stamp=OFFER_STAMP, kind=kind)
     assert offer.sad['p'] == apply.said                 # answers the apply
     assert bespoke.said.encode() in offer.raw           # commits to the bespoke by SAID
@@ -906,7 +906,7 @@ def test_accountability_and_terms_follow_data_JSON():
 
     # Offer and the club's signed agree, as in Phase 3.
     offer = exchange(sender=ALICE, receiver=CLUB, route="/ipex/offer",
-                     attributes=dict(acdc=bespoke.said, terms=_bespoke_rules()),
+                     attributes=dict(acdc=bespoke.said, terms=_rules_in_bespoke()),
                      stamp=OFFER_STAMP, kind=kind)
     agree = exchange(sender=CLUB, receiver=ALICE, route="/ipex/agree",
                      prior=offer.said, stamp=AGREE_STAMP, kind=kind)
@@ -936,7 +936,7 @@ def test_accountability_and_terms_follow_data_JSON():
                                                             ser=agree.raw)
 
     # --- Terms follow the data: the CLC terms are bound into the bespoke SAID. ---
-    tamperedRules = _bespoke_rules()
+    tamperedRules = _rules_in_bespoke()
     tamperedRules['Assimilation']['l'] = "Verifier may do anything it likes."
     weakened = _bespoke_presentation(sedi, over21, kind, rule=tamperedRules)
     assert weakened.said != bespoke.said              # cannot weaken terms silently
@@ -982,7 +982,7 @@ def test_clc_serialization_kinds(kind):
     # Gated exchange: the offer binds nothing PII, the agree binds the offer SAID,
     # and the club's signed agree (assembled via messagize) verifies.
     offer = exchange(sender=ALICE, receiver=CLUB, route="/ipex/offer",
-                     attributes=dict(acdc=bespoke.said, terms=_bespoke_rules()),
+                     attributes=dict(acdc=bespoke.said, terms=_rules_in_bespoke()),
                      stamp=OFFER_STAMP, kind=kind)
     agree = exchange(sender=CLUB, receiver=ALICE, route="/ipex/agree",
                      prior=offer.said, stamp=AGREE_STAMP, kind=kind)
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -924,6 +924,22 @@ def test_accountability_and_terms_follow_data_JSON():
     Alice's wallet must capture the club's key state -- its establishing event -- at
     acceptance time, alongside the signed agree. This test proves why.
 
+    What this does and does NOT establish. Capturing the establishing event makes the
+    signature CHECKABLE against the club's first key; it does not prove that key was
+    the club's CURRENT key when it signed. An undated signature by the (now retired)
+    first key made AFTER the rotation verifies against the captured inception
+    identically -- the scheme cannot distinguish a genuine pre-rotation acceptance
+    from a post-rotation forgery by a compromised retired key, nor stop the club from
+    repudiating a real acceptance. (Pre-rotation protects the club's ongoing IDENTIFIER
+    control -- a thief with the old key cannot continue the KEL -- but not the
+    forgeability of an undated message signed by that key.) So this is durable
+    VERIFIABILITY under cooperative timing, not non-repudiation robust to key
+    compromise. Establishing currency-at-signing requires ordering the signature
+    against the KEL -- anchoring the agree as a KEL transaction seal, or a witnessed
+    receipt -- which this example forgoes for privacy. A high-stakes CLC flow would
+    anchor instead (to a blinded presentation registry, to keep correlation-resistance),
+    sitting at the opposite, accountability-maximizing end of the same tradespace.
+
     It also shows terms follow the data: the CLC terms live in the bespoke ACDC's
     Rules section, which is committed by the ACDC's SAID. Any change to a clause
     yields a different SAID, so a verifier cannot strip or weaken the terms while
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -423,13 +423,10 @@ def _source_credentials(kind):
     aggor = Aggor(ael=_sedi_ael(), makify=True, kind=kind)
     sedi = acdcagg(israid=STATE, uuid=NONCES[10], regid=regState.said,
                    schema=sediSchema, aggregate=aggor.ael, kind=kind)
-    # DESIGN DECISION (pending Sam, plan-of-record §9.1): over-21 is a pre-derived
-    # boolean credential issued by an endorser to Alice (the plan's default), NOT a
-    # zero-knowledge predicate proof. An over-21 -> sedi-id edge could only be I2I if
-    # over-21's issuer equaled sedi-id's issuee (both Alice) -- i.e. if Alice
-    # self-issued over-21; under endorser issuance issuer(over-21) = ENDORSER != ALICE,
-    # so the edge is omitted rather than mislabeled I2I. The load-bearing I2I edges
-    # are the bespoke ACDC's edges to both source creds (Phase 2), where Alice issues.
+    # over-21 is a standalone boolean credential the endorser issues to Alice,
+    # bound to her by its own a.i. It carries no edge to sedi-id and needs none:
+    # the two source creds are independent, joined only at presentation by the
+    # bespoke ACDC's I2I edges (Phase 2), where Alice is issuer and the issuee of both.
     over21 = acdcmap(israid=ENDORSER, uuid=NONCES[11], regid=regEndorser.said,
                      schema=over21Schema,
                      attribute=dict(d='', u=NONCES[14], i=ALICE, over21=True), kind=kind)
@@ -495,10 +492,10 @@ def _bespoke_presentation(sedi, over21, kind, compactify=False, rule=None):
     "no correlatable log" intent. Returns the SerderACDC, most-compact when
     compactify=True (same SAID either way).
 
-    DESIGN DECISION (pending Sam, plan-of-record §9.2): the state-endorsed photo is
-    NOT copied into these attributes; it stays the single source of truth in
-    sedi-id, reached via the I2I edge and disclosed by selective disclosure of just
-    that element (see _photo_disclosure). So no photo bytes are duplicated here.
+    The state-endorsed photo is NOT copied into these attributes; it stays the
+    single source of truth in sedi-id, reached via the I2I edge and disclosed by
+    selective disclosure of just that element (see _photo_disclosure). So no photo
+    bytes are duplicated here.
     """
     _, schema = _saidify_schema(dict(BESPOKE_SCHEMA_MAD), kind=kind)
     attribute = dict(d='', u=NONCES[8], i=CLUB,
@@ -525,9 +522,8 @@ def test_source_credentials_and_selective_disclosure_JSON():
         and leaves the rest as bare SAIDs, and the verifier recomputes the
         aggregate digest (AGID) to confirm the disclosure is authentic.
 
-      * over-21 -- a derived boolean age credential issued to Alice, following the
-        plan's default of a pre-derived boolean (not a zero-knowledge predicate),
-        so the club never needs the birthdate to confirm the age claim.
+      * over-21 -- a pre-derived boolean age credential issued to Alice, so the
+        club never needs the birthdate to confirm the age claim.
 
     The load-bearing selective-disclosure claim -- prove-age-without-birthdate --
     is asserted here: a disclosure revealing the issuee binding and the state-
@@ -575,9 +571,17 @@ def test_source_credentials_and_selective_disclosure_JSON():
     assert disclosed[SEDI_ISSUEE]['i'] == ALICE
     assert isinstance(disclosed[SEDI_PHOTO], dict)          # photo revealed
     assert disclosed[SEDI_PHOTO]['photo'] == "<state-endorsed-photo-bytes>"
-    assert isinstance(disclosed[SEDI_DOB], str)             # birthdate withheld (SAID)
-    assert isinstance(disclosed[SEDI_RESIDENCE], str)       # residence withheld
-    assert isinstance(disclosed[SEDI_NAME], str)            # name withheld
+    assert isinstance(disclosed[SEDI_DOB], str)             # birthdate withheld...
+    assert isinstance(disclosed[SEDI_RESIDENCE], str)       # residence withheld...
+    assert isinstance(disclosed[SEDI_NAME], str)            # name withheld...
+    # ...and each withheld element is provably its committed block SAID, not the
+    # value. This is the assertion that actually proves non-disclosure: dob,
+    # residence, and name are themselves strings, so the isinstance checks above
+    # establish only the bare (compact) form -- they cannot on their own tell a
+    # block SAID from the plaintext. Equality to the aggregate's committed 'd' can.
+    assert disclosed[SEDI_DOB]       == aggor.ael[SEDI_DOB]['d']
+    assert disclosed[SEDI_RESIDENCE] == aggor.ael[SEDI_RESIDENCE]['d']
+    assert disclosed[SEDI_NAME]      == aggor.ael[SEDI_NAME]['d']
 
     # Privacy invariant: the withheld birthdate value never appears on the wire --
     # dob/residence/name travel only as bare block SAIDs. Those SAIDs are
@@ -703,8 +707,8 @@ def test_gated_ipex_exchange_JSON():
          attribute values. (Alice's AID is still visible -- she is the offer's
          sender -- so "no PII" means no attribute values, not sender anonymity.
          Note too that the offer commits to the real, private bespoke SAID rather
-         than the spec's metadata-ACDC variant, a deliberate scope reduction per
-         plan-of-record §6; the SAIDs leak no PII because the ACDCs are private, but
+         than the spec's metadata-ACDC variant, a deliberate scope reduction; the
+         SAIDs leak no PII because the ACDCs are private, but
          the metadata variant is the spec's canonical pre-agreement artifact and
          additionally decorrelates across presentations.)
       2. The agree binds the OFFER's SAID (its 'p' field), not the credential
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -65,6 +65,7 @@ from keri import Kinds, Ilks
 from keri.core import (Salter, Noncer, Aggor, Mapper, Diger, Verfer,
                        exchange, messagize, incept, rotate)
 from keri.core.coring import MtrDex
+from keri.core.serdering import SerderACDC
 from keri.acdc import regcept, acdcmap, acdcagg
 
 
@@ -869,6 +870,28 @@ def test_gated_ipex_exchange_JSON():
     revealed = {i for i, el in enumerate(grant.sad['a']['photo']) if isinstance(el, dict)}
     assert revealed == {SEDI_ISSUEE, SEDI_PHOTO}        # the /A/i and /A/photo paths
 
+    # The club VERIFIES the granted credential before trusting it -- a real operation,
+    # not a SAID-equality trick. Reconstructing it self-verifies the artifact
+    # (SerderACDC recomputes the SAID over the content and raises if it does not match
+    # the committed 'd'); the club then confirms this SAID is exactly the one it agreed
+    # to in the offer (so the terms it accepted are the terms bound into the data it
+    # received -- terms follow the data), and that it validates against the bespoke
+    # schema (which enforces I2I and the required CLC clauses).
+    granted = SerderACDC(sad=grant.sad['a']['acdc'], verify=True)   # self-verifies or raises
+    assert granted.said == bespoke.said                            # == the agreed SAID
+    assert_acdc_schema_valid(granted, schema=bespoke.sad['s'])      # schema-valid (compact form)
+    # A complete verifier performs a 4th check the presentation itself cannot carry:
+    # revocation. Walking the bespoke's I2I edges to each source credential, it checks
+    # that credential's registry (named by its 'rd') for a revocation event before
+    # trusting the assertion -- a revoked source cred must be rejected even when the
+    # presentation is otherwise perfect. The hooks are present here (each edge names a
+    # far node, and each far node is bound to a real registry), but running the live
+    # TEL/registry check (keri.vdr's Tevery/Reger over a Habery) is out of scope for
+    # this data-structure-level example; test_examples.py exercises the registry
+    # lifecycle, and a future automation companion will fold it in.
+    assert bespoke.sad['e']['identity']['n'] == sedi.said and sedi.sad['rd']
+    assert bespoke.sad['e']['age']['n'] == over21.said and over21.sad['rd']
+
     # 5. admit (club -> Alice): acknowledges receipt, closing the exchange.
     admit = exchange(sender=CLUB, receiver=ALICE, route="/ipex/admit",
                      prior=grant.said, stamp=ADMIT_STAMP, kind=kind)
@@ -936,10 +959,15 @@ def test_accountability_and_terms_follow_data_JSON():
                                                             ser=agree.raw)
 
     # --- Terms follow the data: the CLC terms are bound into the bespoke SAID. ---
+    # If Alice had granted WEAKENED terms, the club's grant-time verification (Phase 3)
+    # would reject them: changing a clause changes the SAID, so the delivered credential
+    # is no longer the one the club agreed to and fails the "granted == agreed" check.
+    # (Altering the committed rules SAID in place instead fails SerderACDC self-
+    # verification outright.)
     tamperedRules = _rules_in_bespoke()
     tamperedRules['Assimilation']['l'] = "Verifier may do anything it likes."
     weakened = _bespoke_presentation(sedi, over21, kind, rule=tamperedRules)
-    assert weakened.said != bespoke.said              # cannot weaken terms silently
+    assert weakened.said != bespoke.said              # != the agreed SAID -> rejected
 
 
 @pytest.mark.parametrize("kind", [Kinds.json, Kinds.cesr, Kinds.cbor, Kinds.mgpk])
--- a/tests/acdc/test_clc_disclosure.py
+++ b/tests/acdc/test_clc_disclosure.py
@@ -32,8 +32,9 @@ Every ACDC validates against a real, purpose-authored JSON Schema (Draft 2020-12
 from its first commit -- not the generic default -- and the source credentials
 are bound to real registries created via regcept, matching the worked examples in
 tests/acdc/test_examples.py. Actor AIDs are derived once here from a fixed salt so
-the example is reproducible; they are transferable Ed25519 basic AIDs, which is
-all the data-structure-level examples in this file need.
+the example is reproducible; each is a self-addressing ('E') transferable AID, the
+SAID of an inception event that commits to the actor's initial key and a pre-rotated
+next key.
 
 A note on why later phases model the IPEX exchange and the club key rotation at
 the data-structure level: it is a deliberate scope choice, NOT a gap in keripy.
@@ -58,20 +59,36 @@ from jsonschema.exceptions import ValidationError
 from keri import Kinds, Ilks
 from keri.core import (Salter, Noncer, Aggor, Mapper, Diger, Verfer,
                        exchange, messagize, incept, rotate)
+from keri.core.coring import MtrDex
 from keri.acdc import regcept, acdcmap, acdcagg
 
 
 # --- Reproducible example actors (see module docstring). ---
-# Four transferable Ed25519 basic AIDs derived from one fixed salt. The 'D'
-# prefix marks a transferable public key used directly as an AID; that is
-# sufficient for the data-structure-level modeling here.
-# Five signers: four for the actors, plus _SIGNERS[4] as the club's dedicated
-# rotation (next) key in Phase 4 -- not shared with any actor. HD derivation is
-# deterministic by index, so taking five instead of four leaves the first four
-# (and every SAID that depends on them) unchanged.
-_SIGNERS = Salter(raw=b'clcworkexamplsal').signers(count=5, transferable=True,
+# Each actor is a self-addressing ('E') transferable AID: its prefix is the SAID of
+# an inception event that commits to the actor's current signing key and a digest of
+# its pre-rotated next key. This is the identifier form a real transferable
+# participant uses. A bare public key ('D' basic prefix) is a spec-legal corner that
+# real deployments skip: self-addressing binds the identifier to its inception, and
+# thus to the pre-rotation, rather than leaving which-KEL-is-authoritative to the
+# controller's witnesses and each validator's watchers to establish and confirm.
+#
+# Eight signers derived from one fixed salt: _SIGNERS[0..3] are the four actors'
+# current signing keys (State, endorser, Alice, club) and _SIGNERS[4..7] are their
+# matching pre-rotated next keys. The club rotates to its next key (_SIGNERS[7]) in
+# Phase 4.
+_SIGNERS = Salter(raw=b'clcworkexamplsal').signers(count=8, transferable=True,
                                                    temp=True)
-STATE, ENDORSER, ALICE, CLUB = (signer.verfer.qb64 for signer in _SIGNERS[:4])
+
+
+def _actor_aid(cur, nxt):
+    """Self-addressing (E) AID: the SAID of an inception committing to cur + next(nxt)."""
+    return incept(keys=[cur.verfer.qb64],
+                  ndigs=[Diger(ser=nxt.verfer.qb64b).qb64],
+                  code=MtrDex.Blake3_256).pre
+
+
+STATE, ENDORSER, ALICE, CLUB = (
+    _actor_aid(_SIGNERS[i], _SIGNERS[i + 4]) for i in range(4))
 
 # Per-example blinding nonces, derived (not pasted) the same way the sibling
 # examples derive theirs, but from a distinct raw prefix so the two files do not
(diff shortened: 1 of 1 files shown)


SmithSamuelM:
@dhh1128 

I will answer in chunks since there is a lot here.

> Modeling the SEDI credential as an aggregate (acg) section. I used the aggregate section for the identity credential so that Alice can disclose an arbitrary subset of attributes, revealing the photo while withholding the birthdate. That felt closer to how a real mobile-driver's-license-style credential would behave than the nested-block partial-disclosure pattern would be. I followed the "Composed Schema for selectively disclosable Aggregate section" in the specification, using an array with items and anyOf rather than fixed positions. Is the aggregate section the right choice here, or would you rather this example use the attribute section?

I believe for the core SEDI identity credential the appropriate choice is to use the selectively disclosable Aggregate section.  The state of Utah has not decided but my suggestion is that it be selectively disclosable given most of the use cases seem to fit that model.  In many cases the only disclosed field value would be the issuee AID field.

> Pointing an I2I edge at an aggregate credential. This is the choice I am least confident about, and I would appreciate a careful answer. Because the aggregate credential has no top-level attribute-section issuee, its issuee is carried inside an aggregate element. My bespoke ACDC's I2I edge therefore refers to a credential whose issuee is an element, and I assert the same-holder binding by comparing the bespoke issuer to that element's issuee. When I read keri.vdr.verifying.verifyChain, it looked to me as though it resolves the issuee only from the top-level attribute section, which would mean it cannot enforce I2I for an aggregate far node at all. Have I understood that correctly? Is I2I to an aggregate-issuee credential something the specification is meant to support? If it is, is this a case that a future v2-native verifier should handle, and if it is not, should I avoid aiming an I2I edge at an aggregate credential entirely?

The issuee in an ACDC with Agg section (selectively disclosable) is the "i" field in the second (index 1) attribute block.
This could/should be better clarified in the spec.

I just pushed at change to the SerderACDC class so that its .iseaid property also returns this field value and not just the attribute section field value or None if neither is provided.  The .issuee property returns a Prefixer Matter primitive instance whose .qb64 property is the issuee aid.  I added these asserts to the test_selective_disclosure_aggregate_JSON()

```
    assert acdc.iseaid == BOB
    assert acdc.issuee.qb64 == BOB
```

keri.vdr.verifying.verifyChain is out of date as it was written before the V2 ACDC spec with agg sections.

The operator I2I is meant to work for either att or agg ACDCs when there is an "i" field.  The spec should be clarified for to precisely describe where the 'i' field in an Agg section is found when present.  

The "i" field for Issuee is optional. When not present, it means the ACDC is issued to whom it may concern (untargeted).
In  that case the I2I operator would fail on an edge since there is no Issuee.
This is where the term holder is confusing because anyone could "hold" and present an untargeted ACDC. Unlike w3c credentials where "holder" binding is variable with weak and strong binding. IN an ACDC either you have cryptographically verifiable binding via an "i" field with the issuee aid or you have no binding. Its boolean.

To restate. 
An attributive ACDC has an "a" section and if its "a" section has an "i" field at its top level then the value of that "i" field is the issuee AID.
An aggregative ACDC has an "A" section and if the second block in its "A" section has an "i" field then the value of the "i" field is the issuee AID.
An ACDC can only be either attributive or aggregative not both (mutually exclusive)

> Leaving the bespoke presentation out of any registry. I did not give the bespoke ACDC an rd field, partly because the GoodFood example has none and partly because a one-time presentation that is signed but not anchored seemed to fit the goal of not keeping a correlatable public log of every patron. Is that a fair reading, or should the bespoke ACDC be registered after all?

I think not having a presentation registry is appropriate for the example. 
However, there is some discussion about how to manage presentation registries. Should they be optional on a citizen by citizen basis. Or even a bulk issuance set member basis.  A presentation registry enables the Issuee to detect impersonation fraud of a compromise of its signing keys, so high stakes presentations want to to use a presentation registry, which requires that the "rd" field be populated at the time of issuance to force the verifier to verify the presentation against the presentation registry controlled by the issuee.    

This leads me to believe that we need to find a high stakes use case where the Issuee would want to enforce a presentation registry. Maybe a drug prescription presentation where the prescription is either an ACDC or the cargo of a bespoke ACDC.


> What the offer commits to. In the IPEX offer, I have Alice commit to the SAID of the real (private) bespoke ACDC rather than to a separate metadata ACDC. My understanding is that the specification's Chain-Link Confidentiality flow describes a metadata ACDC, with empty attributes, as the pre-agreement artifact, and that the metadata variant would additionally decorrelate across presentations. I skipped it to keep the example smaller, and I am not sure that was the right trade-off. Should the offer carry a metadata ACDC instead?

I think this is good for this example.  I will discuss in more detail later as the tradespace here has not yet been definitively explored.

> Modeling the IPEX exchange at the data-structure level. I built the apply, offer, agree, grant, and admit messages directly with keri.core.exchange, and I signed the agree with keri.core.messagize, rather than driving the keri.vc.protocoling handlers. I did that because those handlers require a live Habery and operate on the v1 credential objects, and because the v2-native keri.acdc.ipexing module is still a placeholder in the reorganization that is underway. Is it acceptable for a worked example to present the exchange at the data-structure level, or would you prefer that I wait until the v2 IPEX tooling exists and show the flow through that? I tried to point the reader at the real handlers in the docstring so that no one concludes the tooling is missing, but I suspect there is a cleaner way to frame it, and I would welcome your suggestion.

We are doing a fresh implementation in acdc/ipexing. The current tooling in vc.protocologin does not support V2 messaging.  So just doing the bare bones on credential objects is the right level of detail for the example.  As we discussed, doing the same example as a baseline but a different layers of automation is a good way to prove out the automation as its build out. So an example as done at the data structure level is good. And then we want to rework the example with a derived example but using more automation when that becomes available. Having layered examples also helps implementers in other languages as they want to see the bare bones. The real handlers don't exist yet so pointing the reader to vc.protocoling would be confusing at this stage. When acdc.ipexing is fleshed out then pointing to a companion version of the example that uses the tooling would be the most informative.

SmithSamuelM:
@dhh1128 

> What the offer commits to. In the IPEX offer, I have Alice commit to the SAID of the real (private) bespoke ACDC rather than to a separate metadata ACDC. My understanding is that the specification's Chain-Link Confidentiality flow describes a metadata ACDC, with empty attributes, as the pre-agreement artifact, and that the metadata variant would additionally decorrelate across presentations. I skipped it to keep the example smaller, and I am not sure that was the right trade-off. Should the offer carry a metadata ACDC instead?


Exploring the tradespace a little more.  These are my thoughts open for discussion.

There are several mechanisms that could be used to negotiate a graduate disclosure. These include: attribute path lists, aggregate block label lists, metadata ACDCs, Schema, and Decompositions of Schema.   

The problem with negotiating with schema docompositions is that attributes appear disjoint in a schema so pathing to indicate an attribute gets complicated quickly and the pathing logic has to understand JSON schema syntax at a deep level.  Whereas in a metadata ACDC the attribute paths are not disjoint so indicating an attribute at any nesting level can be handled with a path string.   

So it feels like to me that negotiation should be using Composed Schema as type. So only the SAID of the schema is provided and assume the full schema is cached or attached. Then use pathing of attributes to request the level of partial disclsure and metadata ACDCs for terms negotiation prior to disclosure.

Desired attributes are requested as a list of paths of each attribute or block. If the path is a block then the whole block is meant to be disclosed. If the path is a field in a block only the fields in that block at that field level needs to be disclosed. Any fields whose value is a nested block in that block does not need to be expanded and disclosed.  

With nested partial disclosure in an ATT section in order to disclose any field in a block all the simple fields in that block must also be disclosed but non-simple fields i.e. fields whose value is a nested block need only disclose the SAID of the block. So specifiying any simple field via a path would get you all the simple fields in the block holding the specified leaf of the path.

Whereas specifying a block (directory) would be short hand for disclosing the full tree attached all the way down.

For ATT section ACDCs the pathing gets simple since its an array and only the unique field label needs to be specified to designate a given block.

For example
```
'a':
{
    'd': "Ebslde",
    'u': "Abldle"
    'i': "EDdecd",
    "grades":
    {
         "d": "Efghi",
         "u":  "Abdefs",    
         "gpa":  3.7,
        "transcript":
        {
             "d": "EBDE",
             "u":  "ACEF",    
               "math": 3.5,
              "english":3.8",
        }
     }
}
```

}

SAID of composed schema as type defines the ACDC in the apple: EABCDSS...
Attribute path list defines what attributes are being asked for: ['a/i', 'a/grades/']
So apply is asking for the top level of the "a" section via 'a/i'   or '/a/i' and for the full grades block and any nested subblocks of grades via 'a/grades/' or '/a/grades'
Or because grades is nested down from the top level of 'a'.  A path list of simply ["/a/grades/"

Where as a path list of  merely ['/a/grades/'] subsumes the other path list.

To get gpa (and the issuee aid) without the grades detail  would be  simply ['/a/grades/gpa']


In response an offer the chained to the apply could either reply with the same schema and path list, or the rule could be that an offer chained to the same apply that leaves the path list None implies it is offering the same path list.
and then the offer includes a metadata acdc that exposes the rule section and the att section said but not the pathed attributes.
The agree then chains to the offer which indicates agreement
The grant then provides an uncompacted version of the ACDC with the appropriate level of disclosure.


To generalize this to apply that want granduated disclosure of multiple ACDCs in a chain, I suggest a dict
where the item label is the SAID of the ACDC in the chain and the value of the item is the path list of disclosed attributes.

The path lis is not limited to attribute and aggregate sections paths could include edge and rule section paths.

For example two chained acdcs one with attribute section and one with aggregate section

{
      "EABCDED": ["/a/gpa"],
      "EBQRSTR" : ["/A/i","/A/photo"]
}

The path dict requests the gpa field in the ACDC with schema "EABCDED" which means that anything on the path to the "gpa" field must also be disclosed and also requests the selectively disclosable   Issuee and photo blocks of the schema "EBQRSTR"


We cannot use ACDC SAIDs in an apply in general because the actuall ACDC said is assumed to not be know by the applicant. But for defined ACDCs as would be the case for any ACDCs governed by an EGF the  SAIDS of the ACDC schemas would be known. And so the request is for an unidentified ACDC but  with a given schema SAID and the paths indicated by the path list.

When the ACDC desclosure includes a bespoke ACDC the applicant won't necessarily now the SAID of its schema in the application but the inital offer can provide it via a bespoke metadata acdc, and subsequent applies can negotiate on that bespoke ACDC schema said.


The paths shown use the "/' delimeter but for CESR compatibility we could use the CESR Pathing BASE64 compatible path delimeter "-".     Or if we  do a little work to name the path list  field normatively we could have the tooling auto convert from "/" delimited paths to more compact "-" delimited paths when using CESR serialization of the exchange message.
 Or we allow either delimeter and just inject and extract the path as a tuple for reasoning so we are delimiter independent at the reasoning level..

SmithSamuelM:
I created a markdown doc of this discussion to memorialize it better.
Also I created some diagrams about anchoring IPEX exchange transactions that might be helpful (added to document)

https://github.com/SmithSamuelM/Papers/blob/master/whitepapers/ACDC_IPEX_Negotiation.md
https://github.com/SmithSamuelM/Papers/blob/master/whitepapers/ACDC_IPEX_Negotiation.pdf

dhh1128:
Thanks, Sam — this is exactly the guidance I was hoping for, and the `SerderACDC.iseaid` change resolves my biggest doubt. I think I've adopted everything where you were gave clear direction, and I have some thoughts on the negotiation tradespace where you left things open.

## Adopted (pushed to the branch)

- **Aggregate section for SEDI** — kept as-is. Good to have it confirmed. I updated your reasoning into the docstring.
- **I2I to an aggregate issuee** — this was my least-confident call, so thank you for the definitive answer *and* the code. I merged your `iseaid`/`iseaidb`/`issuee` change onto the branch and rewrote the same-holder check to resolve through `.iseaid` for both far nodes — `bespoke.sad['i'] == sedi.iseaid` (aggregative) and `== over21.iseaid` (attributive), so they read identically. I also added `assert sedi.iseaid == ALICE` / `sedi.issuee.qb64 == ALICE` (and the attributive pair on over-21), mirroring the asserts you put on `test_selective_disclosure_aggregate_JSON`. My example already happened to place the issuee at aggregate index 1, so it lines up with `.sad["A"][1]["i"]` — I've annotated that so the convention is explicit rather than accidental.
- **`verifyChain`** — reframed. I'd written it as an open "can this even be enforced" question; it's now stated as what you clarified: `verifyChain` predates V2 agg sections and resolves the issuee only from a top-level attribute `i`, so it's simply out of date for an `acg` far node — a staleness limitation of that function, not a spec one — and the V2-correct resolution is `.ise
- **No presentation registry / no `rd`** — kept. I left a forward-pointer noting that high-stakes presentations would populate `rd` to force verification against an issuee-controlled presentation registry, so the nightclub's no-`rd` choice reads as a deliberate privacy decision rather than an omission.
- **Data-structure-level IPEX** — kept, and I took your point about `vc.protocoling`. I've stopped presenting it as "the real handlers a production agent would call" — that's V1 and doesn't speak V2 messaging, so pointing readers there is misleading at this stage. The docstrings now say the exchange is deliberately bare-bones (which also serves other-language implementers), that `keri.acdc.ipexing` is the v2-native home still being built, and that a companion example will re-run this same scenario through that tooling once it 's available. I like the "layered examples, one scenario, escalating automation" framing as an explicit strategy.

## The exploratory conversation (item 4)

A few reactions to your sketch:

**I read your proposal as *not* path-lists-vs-metadata-ACDCs, but a division of labor**: path lists say *which attributes*, and the metadata ACDC carries the *terms* (rule section + attribute-section SAID exposed, values withheld) to be agreed before disclosure. If that's what you intend, it also answers my original item-4 question cleanly — it was never really either/or.

The strongest thing I can add: **I think the path list should reuse `keri.core.coring.Pather` rather than a fresh `/`-mini-language.** `Pather` already gives you all three things your closing paragraph asks for, right? — you author with `/`, it exposes a `parts` **tuple** for delimiter-independent reasoning, and it serializes compactly with the CESR-safe `-` separator when the parts are Base64-safe (with `""` as the zeroth part for absolute paths, which matches your `/a/i` form exactly). If I'm right, then "should we use `/` or `-`?" stops being a question: the reasoning model is the tuple, and `/` vs `-` is just the string vs. CESR rendering of the same `Pather`. That also means the negotiation paths and the CESR SAD paths we already use for signing/disclosure share one grammar and one implementation.

One subtlety: the path doesn't literally denote the disclosed set — **it seeds a closure, and the closure is section-type-dependent.** Your `grades/gpa` example is right that asking for `gpa` forces the simple siblings of its block and collapses nested blocks to their SAIDs. But that expansion rule differs by section:

- **ATT** — disclosing any simple field forces *all* simple siblings in that block (they share the block SAID and blinding), and nested blocks collapse to their SAID. So `/a/grades/gpa` = gpa + its simple siblings + the issuee on the path.
- **AGG** — elements are independently blinded, so `/A/photo` discloses *exactly* that element, no forced siblings.

Am I right?

It looks like your multi-ACDC example already mixes both (`/a/gpa` alongside `/A/i`, `/A/photo`), so the grammar is syntactically uniform but the disclosure closure branches on section type. I think that closure wants to be a single normative function — call it `disclosureSet(schema, paths)` — that both discloser and disclosee compute identically, so the request path is just the seed and there's no room for two implementations to disagree about what was actually agreed.

**Possible gaps**

1. **The multi-ACDC dict is keyed by *schema* SAID** (`{"EABCDED": [...], "EBQRSTR": [...]}`). That can't disambiguate a chain that contains two credentials of the *same* schema — two same-type creds collapse to one key. Since you already note the applicant can't know the actual ACDC SAIDs, keying by the **edge/node label** in the requester's own apply-schema (which the applicant *does* control) might be the more robust handle for chained disclosure.

2. **"An offer with a null path list implies the same list as the apply"** worries me for a security-relevant negotiation. CLC is about *binding* terms, so I'd rather the agreed disclosure scope be explicit and integrity-protected than inferred by an echo convention two implementations might read differently. This is where your metadata-ACDC idea might be most helpful: let the offer's metadata ACDC carry the agreed path list (in its attributes/rules), and have the agree commit to *its* SAID. Then "what was agreed" is a hashable artifact, not a convention.

**The bespoke-schema bootstrap is nice and directly relevant here** — the applicant can't know a holder-authored bespoke schema SAID up front, so the initial offer supplies it via a bespoke metadata ACDC and subsequent applies negotiate against it. I like that resolution, and I assume it's the mechanism the disclosure example would eventually grow into.

Not sure how much of this you'd like me to include in this PR. You said the current offer-commits-to-the-real-SAID approach is fine for the example, and I've kept it so far. But if the negotiation model above is roughly where you want to go, I'd be glad to improve this example, or to build the *next* layered example around it — `Pather`-based path lists, a normative `disclosureSet()`, and the metadata-ACDC terms artifact.

Two things the diagrams make me want to check with you before the follow-up:

1. **Is the whole apply→admit run one exchange transaction, or two?** The diagrams show all five under a single Boxer, which would make `apply` the `xip` inception and the rest `exn` continuations. But negotiation (apply/offer/agree) and disclosure (grant/admit) feel like they could be separate transactions — the agree closes the terms, the grant opens the disclosure. I assume normal case is one transaction, but did you intend for two txns to also be conceivable? It decides whether the automated example builds one anchored transaction or two, and where the `xip` boundary falls.

2. **Anchoring vs. correlation-resistance isn't the binary I implied.** My current docstring treats the un-anchored agree as *the* privacy move ("not KEL-anchored, so the club keeps no correlatable log"). But diagram 4 shows the reconciliation I was missing: anchor the exchange (and the presentation) to a *blinded* state registry, and you get non-repudiable accountability **and** correlation-resistance at once. So the nightclub's un-anchored choice is one point on a spectrum, not the only private option — and I should probably soften that framing in the example. Happy to.

That also gives the presentation-registry follow-up a concrete shape: yes, the high-stakes case (the prescription idea) is worth its own worked example, and diagram 4 is essentially its architecture — the discloser anchors the grant into a blinded presentation registry (BAB → `granted`), so the issuee can detect impersonation after a key compromise while the registry stays correlation-resistant. Showing that side by side with the nightclub's no-`rd` presentation would make the "when do you want a presentation registry" decision concrete. Do you want me to take that on, too?

SmithSamuelM:
@dhh1128 

> Pointing an I2I edge at an aggregate credential. This is the choice I am least confident about, and I would appreciate a careful answer. Because the aggregate credential has no top-level attribute-section issuee, its issuee is carried inside an aggregate element. My bespoke ACDC's I2I edge therefore refers to a credential whose issuee is an element, and I assert the same-holder binding by comparing the bespoke issuer to that element's issuee. When I read keri.vdr.verifying.verifyChain, it looked to me as though it resolves the issuee only from the top-level attribute section, which would mean it cannot enforce I2I for an aggregate far node at all. Have I understood that correctly? Is I2I to an aggregate-issuee credential something the specification is meant to support? If it is, is this a case that a future v2-native verifier should handle, and if it is not, should I avoid aiming an I2I edge at an aggregate credential entirely?

So a little asynchrony here in thought.  I have not digested your latest set of lengthy commentary.  But I wanted to capture my second thoughts on this question first.

After more thought, I think my answer to the SEDI iidentity credential as an aggregate credential (array of attributes instead of nested field map).  I realized I was falling into the trap of letting prior usage dictate instead of first principles.

So I now think that the SEDI core identity core identity credential should be an attribute credentials. Let me explain why.

From a data correlation perspective (the first principle). The one feature that an aggrative credential provides over a attributative credential is that the attribute blocks at the top level are not labeled with meaningful labels. So the position of a block within the array can be managed so as not to leak information where such information could be leaked via relative position.  In the case of the core SEDI identity credential, there are no fields that benefit from such arrangment.  The field labels will be well known and all the fields are a defined set so there is no “advantage” to using the aggrative form and there is a definite disadvantage to using the aggrative form. The disadvantage is that reasoning about the fields to be disclosed now does not use fully labeled paths but either must a derived path.  If one says that the array index is in the path, then one defeats the one property of the array that one wanted to preserve. That means pathing for apply messages means that under the hood one has to convert an abstraction to a path without being given the index. One has to create a reverse lookup table that maps the unique label embedded in each attribute block into an index so that a path can look up the array index.

While this is no terribly complex, since there is no advantage to using an array for this credential, more complexity for the most used credential, the core SEDI identity credential feels like a poor decision.

So, given you have already done an example with an aggregation. I think it’s useful example still. But it would be great to do another example one where the core SEDI credential was instead an attributive one,  each attribute besides the issuer aid, gets a nested labeled partially disclosable block with a said and uuid in its block. 


This brings up the question of what is a valid use case for the aggrative credential that benefits from not labeling the attribute blocks where relative position in the array could leak information.  The use case that I can think of is an array of Boolean flags for age thresholds.   Each age threshold has an element so that the number of elements does not leak anything about which age thresholds are present. All are present but each is true or false.  

And to go further, one could add empty elements in the array and randomize the relative positions in the array of the age thresholds so that a disclosure of given array of age threshold does not leak information about which elements in the array correspond to which ages. But the later one is probably overkill.  

An aggrative array in this case is a poor mans sparse merkle tree.

So in hindsight I got the answer backwards. A better example would be to use an attributive ACDC for the core SEDI identity credential and use an aggrative credential for the proof of age credential only includes several Boolean flags for a set of ages like 13, 16, 18, 21. For the most important minimum legal ages

The plan is to eventually use a sparse merkle tree for this but using the Boolean array replicates how mdoc does so its at least as good as mdoc

Anyway appreciate your feedback thoughts as too the inversion of the example.

SmithSamuelM:
@dhh1128 

One of the reason to do these examples is to force exploring the tradespace in a concrete way. So learning something and then changing based on the learning is making progress and at a stage where its lower risk.

> Aggregate section for SEDI — kept as-is. Good to have it confirmed. I updated your reasoning into the docstring.

So now I am admitting I got it wrong. Your oriiginal trepidation was well placed.


## pathing

I agree that using the existing Pather makes the most sense.  This does impose some subtle constraints on the serialization.  Currently pather is used in attachments to reason about the message to which the path is attached. The pathed material group code in the attachment tells the parser to interpret the encoded string as a path.  

For generic exchange messages field values that are strings do not know to use pather serialization unless we make them special fields.  So where we convert to a Pather is relevant. There are some choices and so more discussion before hardening is warrented.

Another related option not mentioned so far is which block in an exchanges should requested disclosure paths in an Apply go in.  There is a query block and an attribute block. By design there are two blocks so that the exchange (as well as the other message types query and reply) can more easily model ReST endpoint logic to minimize the impedance mismatch for building KERI APIs.  If one thinks of an http verb with a URL with path and query string and verb message body, the modeling is as follows.  The route field in the exchange hold the path from the URL. The query block holds the query string as field map, and the attribute block holds the verb message body.     

So we can put disclosure request information in the apply in three places, route, query block, attribute block

The route prefix is /apply but one could overload that and use /apply/schemasaid. 
One could define a field in the query block to put a dictionary of paths
One could define a field in the apply block to put a dictionary of paths.

Becaue the vLEI did not employ graduated disclosure there was no need to go to any lengths to define this stuff, but now we are discussing the syntax for negotiotinhg over graduated disclosure and wanted to expose these subtilties

## Reasoning with graphs vs reasoning with schema

> The multi-ACDC dict is keyed by schema SAID ({"EABCDED": [...], "EBQRSTR": [...]}). That can't disambiguate a chain that contains two credentials of the same schema — two same-type creds collapse to one key. Since you already note the applicant can't know the actual ACDC SAIDs, keying by the edge/node label in the requester's own apply-schema (which the applicant does control) might be the more robust handle for chained disclosure.

You are right the simple example does not account for a chain with repeated schema.  So we need to be more thoughtful in how we define it.

This opens up a bigger question, and you reference to the own “apply-schema” brings in a related issue.  This is an area where I have spent some significant effort.  The W3C community wants to reason using knowledge graphs, A schema is a type of knowledge graph,  a chained graph of ACDCs is also a type of knowledge graph,  ACDCs can be modeled with labeled property graphs which are also a type of knowledge graph.  

The core insight for me is not should we have the ability to reason about or with knowledge graphs but what type of reasoning are we doing and should we constrain the type of reasoning in a layering of concerns so as to better address the concerns at each layer.  Knowledge graphs as widely used in the world do not as a first order design principle have any security, authenticity, verifiability, confidentially, privacy etc.  These latter concepts are bolt ons to the standard knowledge graph community and I found the W3C approach of attempting to bolt on security to RDF graph expansions a pathological mess (excuse my strong language). What knowledge graphs are meant to do is reason about semantics not reason about authenticity or correlatability. These are different types of reasoning.

One of the prime motivations for ACDCs was to invert that layering. Build mechanisms for security and confidentiality and correlation resistance in layers where these happen before the semantic reasoning happens.

So disclosure is reasoning about exposure. What parts to expose when. Not what those parts mean.   So to separate concerns my intuition is we should minimize the role schema play in this reasoning. This is accomplished with schema is type, i.e.  the SAID of the schema, and the schema validation validates type.  But using bespoke schema as the way to indicate exposure is using schema for the wrong reason. And it’s more complicated to boot.

But reasoning about graphs is general when those graphs are expandable or Collapsable or compactifiable is reasoning about the graph structure not the semantics of the information contained in the graph.  In this case it seems to me we can reason about graph structure using a graph. The simplest expression for that graph structure we can think of not the most complex expression (JSON schema) for that graph structure we have available.

So if we think of an ordered dictionary as a graph, we can model a graph a ACDCs with a dictionary (associative array or field map (we already model the guts of each acdc with a dictionary) so we can extend that by modeling the full graph of ACDCs with an expanded dictionary. This allows use to use a pathing language that is as simple as we can make it.

It has to include a way to path along a chain of ACDCs not merely into an ACDC.  As you point out with the over simple example I provided of a path list.  But we could use an ordered path list to represent a branch in an ACDC graph. The ordering resolves the schema SAID ambiguity.

We could use the edge labels in the path to explicity trace which edge leads to which ACDC schema SAID and then path down into that ACDC to the disclosure.

Yes the insight is correct a request to disclose that uses a path to a leaf implies a closure of all that must be disclosed in order to disclose the leaf but no more. So compacted sibling branches stay compacted. But then we need a compact way of requesting full expansion so I used the trailing “/“ to indicate. The opposite expand all below.

SmithSamuelM:
@dhh1128 

Because this pull request discussion is broadening into a more general discussion on the pathing negotiantion syntax for  graduated disclsosure I created a discussion here and copy pasted some of the parts about that topic specifically.

I expanded some of my comments there instead of here.

https://github.com/WebOfTrust/keripy/discussions/1512

SmithSamuelM:
@dhh1128 

> Is the whole apply→admit run one exchange transaction, or two? The diagrams show all five under a single Boxer, which would make apply the xip inception and the rest exn continuations. But negotiation (apply/offer/agree) and disclosure (grant/admit) feel like they could be separate transactions — the agree closes the terms, the grant opens the disclosure. I assume normal case is one transaction, but did you intend for two txns to also be conceivable? It decides whether the automated example builds one anchored transaction or two, and where the xip boundary falls.

Both can work. We have to decide if a spurn ends the transaction and one starts over with a new xip for the apply. I think this is cleaner. But technically one could include the spurn in the transaction and chain a new apply there and just keep going. I can think of use cases where the actualy negotional back and forth wants to be memorialized not the linear final result. But for most use cases the final linear result is best.

>Anchoring vs. correlation-resistance isn't the binary I implied. My current docstring treats the un-anchored agree as the privacy move ("not KEL-anchored, so the club keeps no correlatable log"). But diagram 4 shows the reconciliation I was missing: anchor the exchange (and the presentation) to a blinded state registry, and you get non-repudiable accountability and correlation-resistance at once. So the nightclub's un-anchored choice is one point on a spectrum, not the only private option — and I should probably soften that framing in the example. Happy to.

Exactly!  this is one of the main reasons for blinded state registries.  (not simply perfect forward privacy)

SmithSamuelM:
@dhh1128 

>That also gives the presentation-registry follow-up a concrete shape: yes, the high-stakes case (the prescription idea) is worth its own worked example, and diagram 4 is essentially its architecture — the discloser anchors the grant into a blinded presentation registry (BAB → granted), so the issuee can detect impersonation after a key compromise while the registry stays correlation-resistant. Showing that side by side with the nightclub's no-rd presentation would make the "when do you want a presentation registry" decision concrete. Do you want me to take that on, too?

Yes

SmithSamuelM:
>Not sure how much of this you'd like me to include in this PR. You said the current offer-commits-to-the-real-SAID approach is fine for the example, and I've kept it so far. But if the negotiation model above is roughly where you want to go, I'd be glad to improve this example, or to build the next layered example around it — Pather-based path lists, a normative disclosureSet(), and the metadata-ACDC terms artifact.


Basically whether is this same pull request or another for the refactored example is ok either way. 

What we are doing is sussing out what the tradeoffs are so we can make reccommendations for the ecosystem governance framework for SEDI around these graduated disclosure presentations.  The vLEI ecosystem governance framework does not have graduated disclosure so its a NOP for the EGF but no so for SEDI's EGF.

SmithSamuelM:
> Is the whole apply→admit run one exchange transaction, or two? The diagrams show all five under a single Boxer, which would make apply the xip inception and the rest exn continuations. But negotiation (apply/offer/agree) and disclosure (grant/admit) feel like they could be separate transactions — the agree closes the terms, the grant opens the disclosure. I assume normal case is one transaction, but did you intend for two txns to also be conceivable? It decides whether the automated example builds one anchored transaction or two, and where the xip boundary falls.



I think I misunderstood your question. I assumed you meant should a failed transaction that starts over be one or two transaction but you were asking if a single linear transaction should be one or two where the boundary is where the anchors occur.

For linear transactions I see no reason to split.  What we are anchoring is the tail of the transaction not the head. Because the events in an exhange transaction are hash chained together, anchoring the tail also anchors the whole transation including the head.  

Multiple anchors are not meant to anchor multiple transactions but as continencies for the transaction to proceed.  
In a high-stakes transaction, the Discloser wants to see a perpetually verifiable commitment to the terms via the Agreement of the Disclosee before issuing a Grant to the agreeor.  The Disclosee (agreeor, admitor) then wants to see that the Discloser (grantor) anchors the grant before producing an admit.  Often there is mutual accountability in terms and conditions especially in high stakes transactions and so the Disclosee (recipient) wants to have perpetual verifiability of the Discloser's (grantor's) mutual committment before issuing its admit.  

Why the admit, because in contract law, a potential point of legal enforcement friction is that the Disclosee claims they never received the grant. The admit message removes any plausible deniability by the Disclosee. The anchor of the Grant message is incentivized by the following admit. The grantor wants the admit but cant get it until they anchor their grant.   The terms of the  contract could include an escape clause for the grantor if there is no admit message which incentivizes the admit.  And since it must be anchored, the admitor has a way to prove it later, to nullify the escape clause.


Typically high stakes transactions happen under the auspices of a trusted third party who ensures the transaction completes (settlment). But we want to have true peer-to-peer, legally enforceable transactions without trusted third parties, so we need a sequence of commitments that are are successively incentivized without a trusted third party to ensure the completion of the transaction (settlement).  

So the Offer/Agree/Grant/Admit removes plausible deniability which is the bane of peer-to-peer interactions. This is related to the byzantine generals problem which requires a multi-phase commit to solve.

dhh1128:
Reworked per Sam's 07-16 first-principles reconsideration: the core **SEDI identity is now attributive** and the **age credential is aggregative** (an array of boolean threshold flags for 13/16/18/21/55/65). This lands the aggregate form on the use case it's actually for — hiding which age thresholds are asserted — and models the fixed, labeled identity field set attributively (clean labels, clean partial disclosure). As a bonus the one example now demonstrates **both** graduated-disclosure mechanisms: attributive partial disclosure (reveal the photo block, withhold DOB) via `Compactor`, and aggregative selective disclosure (reveal over-21, withhold the other thresholds) via `Aggor`. `.iseaid` is still exercised on both paths, now swapped. All pinned SAIDs regenerated; 8 tests green across JSON/CESR/CBOR/MGPK. The negotiation-syntax points have moved to discussion #1512.

dhh1128:
> Why the admit... Typically high stakes transactions happen... So the Offer/Agree/Grant/Admit removes plausible deniability

Makes total sense. However, what are the expected consequences of a spurn at each of these stages? A spurn after offer is obvious. A spurn after agree just means the discloser backed out, right? But what about a spurn after grant? Is that just a simple "I didn't admit this after all, so I'm not bound by T&Cs" ?

SmithSamuelM:
>Makes total sense. However, what are the expected consequences of a spurn at each of these stages? A spurn after offer is obvious. A spurn after agree just means the discloser backed out, right? But what about a spurn after grant? Is that just a simple "I didn't admit this after all, so I'm not bound by T&Cs" ?

No we want to avoid such assumption. An agree is legally binding not sending an admit does not convey backing out. The purpose of the admit is not to amend the agreement. That agree defines the agreement. The admit is a corner case that enables easy enforcement by removing plausible deniablity.  Confidentiality is about how one came by the information. So without an admit, the agreeer can claim they could not have leaked because they did not actually get the grant. The grant transmission failed.  So the admit removes that excuse.  The incentive for the admit would come from mutual commitments on the part of the grantor. The grantor gets an escape clause unless the admitter verifiably admits. So when and if enforcement happens, the grantor can't back out if the admitter produces proof of admission, which also means the admitter has no plausible deniability.  It balances the enforceability by both parties. 

 I see this kind of language in legal contracts and NDAs where the lawyers attempt to close off arguable defenses, not changing the terms but finding ways to make the contract more easily enforceable.

This is why its related to the byzantine generals problem, if I send an order to attack I need an acknowledgment the other party got the order or I don't attack, but then the other party needs to know I got their acknowledgment or they don't attack which means I need to know they got my acknowledgement of their acknowledgment. It never resolves unless there is a ratchet mechanism and that looks like a 3 of 4 threshold.

In legal contracts the problem is not so dire as the byantine generals problem. The problem is enforceability and eliminating easy defences to enforcement like I did not get the grant so even though I agreed to treat the data as confidential, I never got it so you can't punish me.  The admit removes the easy out. But to incentivize the admit the grantor needs an escape clause. Neither becomes important unless and until there is an enforcement action and basically they nullify each other. Its a clever legal mechanism for confidentiality provisions.

there are other legal mechanisms like a blanket clause that says no matter how you get this information you treat it as confidential going forward, but how can I enforce if they don't know what the information is in the first place. Usually this means I have to hand deliver the information so I can simultaneously prove they got it as they get it.
When this happens remotely over the internet then that provability goes away.

 So when we are explicitly not disclosing until the terms are agreed to, we can't use these blanket clauses over the internet. The plausible deniability that the grant was never recieved offers and easy out.  The admit message defeats that easy out. The challenge is then incentivizes the admit.  Even without an admit, we can enforce the agreed-upon terms, but now we have a burden of proof problem. We have to show evidence that the receiver did indeed get the information in some way.   

This is why the duty of data loyalty is so useful. Because with the regulation of the duty of data loyalty, it doesn't matter how the receiver got the information, if they use information about us, no matter the source, in a self-dealing (not in our best interests) way, we have recourse.

So CLC and DDL are mutually reinforcing. Once data processors understand that they have unlimited liability with a duty of data loyalty, then they will earnestly want a regulatory safe harbor, and in order to get that regulatory safe harbor, we can make it so that they have to provide the admission.  So now we have an incentive that forces the receiver to remove their own plausible deniability. It really ties the bow on CLC and the duty of data loyalty in my opinion.

SmithSamuelM:
@dhh1128 

As is typical for me, I am a lucid dreamer and when I sleep after spending a day thinking about something I usually remember my dreams the next morning where I solve problems in my sleep.  Often this means recalling or being reminded of things I missed or overlooked or got wrong.  So once again I made a mistake that needs to be corrected.

The mistake is that for the worked example, the `I2I` unary edge operator is misapplied.  The reason for the oversight is that so far every one of  the prior chained ACDC examples used chaining via edges to define delegation relationships. But this worked example is not a delegation chain. We want to specify a relatioship between the Issuee of the core SED identity credential and the Issuee of the age credential. The Issuer of both is the same and the Issuee of both is the same but the Issuer is not the Issuee of both.  So the edge defines an an identity relation not a delegation relation.

We need new edge operators.  Sorry but the example is wrong and needs to be corrected. 
There are two ways to fix.

1) no edge operator, the edge relation is defined by the EGF
2) create new edge operators for the edge relation.

What follows is a detailed proposal for new edge operators.


## New Unary Edge operators

### Unary Delegative Operators

With the exception of the NOT operator, the current set of Unary operators are delegative. They modulate delegation relations between the near and far ACDCs one an edge.


| Unary Operator | Description | Default|
| :---: | :--- | :--- | 
|`I2I`| Issuer-To-Issuee, The Issuer AID of this ACDC MUST be the Issuee AID of the node this Edge points to.  | Yes |
|`NI2I`| Not-Issuer-To-Issuee, The Issuer AID of this ACDC MAY or MAY not be the Issuee AID of the node that this Edge points to. |  No |
|`DI2I`| Delegated-Issuer-To-Issuee, The Issuer AID of this ACDC MUST be either the Issuee AID or a delegated AID of the Issuee AID of the node this Edge points to. | No |
|`NOT`| Logical NOT. The validity of the node this Edge points to is inverted. If valid, then not valid. If invalid, then valid.  | No |


When the Operator, `o`, field is missing or empty or is present but does not include any of the `I2I`, `NI2I` or `DI2I` Operators then:

- If the node pointed to by the Edge is a targeted ACDC, i.e., has an Issuee, then the `I2I` Operator MUST be appended to the Operator, `o`, field's effective list value.

- If the node pointed to by the Edge block is an Untargeted ACDC i.e., does not have an Issuee, then the `NI2I` Operator MUST be appended to the Operator, `o`, field's effective list value.

Many ACDC chains use Targeted ACDCs (i.e., have Issuees). A chain of Issuer-To-Issuee-To-Issuer Targeted ACDCs in which each Issuee becomes the Issuer of the next ACDC in the chain can be used to provide a chain-of-authority. A common use case of a chain-of-authority is a delegation chain for authorization.

The `I2I` unary operator, when present, means that the Issuer AID of the current ACDC in which the Edge resides MUST be the Issuee AID of the node to which the Edge points. Therefore, to be valid, the ACDC node pointed to by this Edge MUST be a Targeted ACDC. This is the default value when the Operator 'o' field value is missing or empty.

The `NI2I` unary Operator, when present, removes or nullifies any requirement expressed by the `I2I` Operator described above. In other words, any REQUIREMENT that the Issuer AID of the current ACDC in which the Edge resides MUST be the Issuee AID, if any, of the node the Edge points to is relaxed (not applicable). To clarify, when operative (present), the `NI2I` Operator means that both an Untargeted ACDC or Targeted ACDC, as the node pointed to by the Edge, MAY be valid even when Untargeted or if Targeted even when the Issuer of the ACDC in which the Edge appears is not the Issuee AID, of that node to which the Edge points.

The `DI2I` unary Operator, when present, expands the class of allowed Issuer AIDs of the node the Edge resides in to include not only the Issuee AID but also any delegated AIDs of the Issuee of the node to which the Edge points.  Therefore, to be valid, the ACDC node pointed to by this Edge MUST be a Targeted ACDC.

The `NOT` unary Operator, when present, inverts the validation truthiness of the node pointed to by this Edge. If this Edge's far node ACDC is invalid, then the presence of the `NOT` operator makes this Edge valid. Conversely, if this Edge's far node ACDC is valid, then the presence of the `NOT` Operator makes this Edge invalid.

### Unary Identity and Equivalence operators

The unary operators needed by the vLEI ecosystem were delegative, i.e. the associated edge provides a delegation relationship. The SEDI ecosystem, however, benefits from additional classes of unary operators, (the worked example is such a case). In SEDI, a common case is that an edge provides an identity relationship or an equivalence relationship.  

The syntax convention for edge operators in general is to use Base64 characters for the operator symbols to ensure the CESR serialization of the operator is compact. So far only uppercase and number characters have been used. This follows a convention in automated reasoning systems to make the operator symbols stand out from their operand symbols which are lowercase.

An edge points backward from the near side ACDC holding to the edge to the far side acdc whose SAID is the node `n` field in the edge block. 

In the case of `I2I` the number `2` is used to symbolize delegation that flows down from far side  to (`2`) near side on an edge that points from near side to (`2`) far side. The delegation authority flows in the opposite direction of the edge. To clarify, the edge points from near `2` far  and delegation authority points from far `2` near where far is the delegator and near is the delegatee.

To clarify, `I2I`  means the Issuer AID of the near side is the Issuee AID of the far side. This means the far side ACDC Issuer delegated issuance authority to its Issuee AID, which AID is equal to the near side Issuer AID. To elaborate, in order for the edge to be valid, the near side Issuer AID MUST equal the far side's Issuee AID.
In the shorthand expression Issuer-to-Issuee, the "to" is pointing along the edge direction not the delegation authority direction.  

This convention can include a modifier on the identifier relationship defined by the edge. 

For example: `DI2I` means the near side Issuer AID MUST either be the far side Issuee AID or be a delegated AID of the far side Issuee AID.  So from a delegation authority point-of-view the "D" modifies who the delegatee can be. From an edge point of view it modifies what the near side Issuer AID can be for the edge to be valid, and that modification is that the near side Issuer AID can be equal to the far side Issuee AID or a delegated AID of the far side Issuee AID.

For SEDI, in addition to delegation relationships, we need identity and equivalence relationships. Specifically relationships between the near side issuee AID and the far side Issuee AID and also the near side Issuer AID and the far side Issuer AID. To clarify, these are not delegation relationships but either an identity or equivalence relationship. 

For example, one identity relationship (reflexive) is that the Issuee AID on the near side must equal the Issuee AID on the far side of the edge. Another would be that the Issuer AID on the near side must equal the Issuer AID on the far side of the edge. This poses a problem for the syntax because Issuer and Issuee both start with the letter "I". This is not a problem for a delegation relationship; there is only one pairing between Issuer and Issuee that is meaningful, so I2I is not ambiguous. For identity and equivalence relationships, however, we need to disambiguate Issuer from Issuee. 

Let's use E to represent IssueE, and continue to use I to represent Issuer. 
We could then alias the existing `I2I` operator to be `I2E` for Issuer to Issuee.  In hindsight using `I2I` was nearsighted. But we can keep it and add the alias `I2E` or replace it with `I2E` or leave as is. Likewise for the pair `DI2I` and `DI2E`, and the pair `NI2I` and `NI2E`.

Let's also use the number `1` to represent an identity relationship, which is evocative of the number `1` being the multiplicative identity. This is not the same as a delegation relationship. Delegation implies some form of transitivity.

So we have the new proposed identity relationship operators

`I1I` which means near side Issuer AID must equal far side Issuer AID.
`E1E` which means near side IssueE AID must equal far side IssueE AID.

And we could then have in addition.
`I1E` which means near side Issuer AID must equal far side IssueE AID.  This would not imply a delegation relationship, just an equality. So its not the same semantics as `I2E` or
`E1I` which means near side IssueE AID must equal far side Issuer AID.  

In mathematics, an equivalence relationship adds symmetry and transitivity to the reflexivity of an identity relationship. While not the same mathematically, in ACDC land an analogous type of equivalence relationship is delegation. And there are two senses of the use of the term delegation in KERI/ACDC. We have ACDC edges that delegate and we have delegated identifiers this are not mutually exclusive so we can have delegated edges between delegated identifiers as is the case for the `DI2I` operator.

So we can apply the same modifier to our identity relationships.
`DI1I` would mean that the near side Issuer MUST must be either the far side Issuer AID or a delegated AID of the far side Issuer AID. This would allow an agency relationship where say a driver's license was issued by a different Issuer AID from the core SEDI credential but must be a delegated AID. Or we don't have an identity operator for the Issuer on the edge but allow the EGF to specify the relationship.

So to describe more complex equivalent operators where what is meant by "equivalence" may be ecosystems specific. If we try to use unique operators for each ecosystem specific definition of equivalence we create for us a difficult configuration management task of tracking what might be an unlimited number of unary operators.  If instead we default to saying that "equivalence" is defined by the EGF for a given set of ACDC types then we can define a smaller set of Unary operators whose operation is assigned on an ecosystem (EGF) basis.

The tradeoff is that we want to automate the process of validating or invalidating edges, so the more precisely we can define operators the more we can performance optimize tooling that is less prone to implementation errors.
So its a tradespace judgement call.

Lets use the number `3` to represent generic equivalence relationships where the actual meaning of equivalence is EGF defined. The `3` is evocative of the mathematical equivalence relation that has 3 properties, (reflexivity, symmetry, and transitivity). This provides a placeholder for explicitly requiring such a relationship on an edge, rather than merely implying it. But the relationship is looked up or relative to the EGF.

This gives us:
`I3I` and `E3E` for EGF defined edge equivalence relations between near side and far side Issuers and near side and far side IssueEs.

dhh1128:
> We need new edge operators. Sorry but the example is wrong and needs to be corrected. There are two ways to fix.

Okay, I will raise a follow-up that fixes this. Look for a new PR today.

SmithSamuelM:
@dhh1128 

I created a companion discussion on the new edge operators

https://github.com/WebOfTrust/keripy/discussions/1515

PR #1507, opened 2026-07-15
fix: set IPEX grant recipient field

lenkan:
## Summary
- populate the IPEX GRANT `rp` field with its recipient AID
- cover recipient routing for standard and OOBI grants

## Test plan
- [x] `pytest tests/vc/test_protocoling.py`


lenkan:
Wrong base

PR #1509, opened 2026-07-15
Disable KRAM for V2 OOBI service endpoint replies

KeaxD:
- Add built-in KRAM denials for V2 rpy /end/role and rpy /loc/scheme
- Keep OOBI service-endpoint discovery on the BADA reply path instead of routing through kramit
- Preserve config precedence by only adding the built-in denials when existing config does not already cover them
- Add tests for V2 reply messages with KRAM disabled

### DIFF
--- a/src/keri/core/kraming.py
+++ b/src/keri/core/kraming.py
@@ -245,6 +245,7 @@ class Kramer:
     def fullDenials(self):
         """Returns raw denials list from config.
         Each denial is [[major, minor], ilk, route]
+        OOBIs routes /end/role and /loc/scheme are automatically added to the list when KRAM is enabled.
         """
         return self._fullDenials
 
--- a/src/keri/core/kraming.py
+++ b/src/keri/core/kraming.py
@@ -95,6 +95,11 @@ class Kramer:
         denials (list): Compacted denial strings for exempted messages
         fullDenials (list): Raw denial configurations
     """
+    OobiDenials = (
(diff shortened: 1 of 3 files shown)


tests/app/test_oobiing.py
```diff
@@ -245,6 +246,169 @@ def test_loaded_v1_endpoint_replies_use_stored_reply_framing():
         assert locer.url == "http://127.0.0.1:5555"
 
 
+def test_loaded_v2_oobi_endpoint_replies_bypass_kram(mockHelpingNowUTC):
+    cuekwa = dict(version=Vrsn_2_0, kind=Kinds.cesr, gvrsn=Vrsn_2_0)
```
ryan-hansen: Unused.

src/keri/core/kraming.py
ryan-hansen, line 249: We should probably note here that this will also contain OobiDenials, not just denials from config.

tests/core/test_kraming.py
```diff
@@ -426,6 +426,46 @@ def mock_kramit(msg, kwa=None):
...
+                return msg
+
+            kramer.kramit = mock_kramit
+
+            serder = reply(
+                pre=TEST_PRE,
+                route="/end/role/add",
+                data=dict(a=1),
+                pvrsn=Vrsn_2_0,
+            )
```
ryan-hansen, line 461: Is it worth doing the same for /loc/scheme just to be thorough?

PR #1510, opened 2026-07-15
test: make Ogler global-state tests xdist-safe

jaelliot:
## Summary

Makes the `keri.help.ogler` tests independent of execution order and
pytest-xdist worker scheduling.

## Problem

Keripy initializes its package-global `Ogler` with `syslogged=False`, but
parts of `tests/help/test_ogling.py` depended on another test replacing that
global with HIO's default `syslogged=True` configuration.

Under xdist, `test_reset_levels` may run in a separate worker and see the
production default. It then incorrectly expects a syslog handler and fails
with `AssertionError: assert 1 == 2`.

## Changes

- give global-Ogler tests explicit production-equivalent state
- restore `syslogged=False` during cleanup
- assert configured handler behavior instead of relying on test order
- preserve generic HIO Ogler coverage

## Validation

- isolated `test_reset_levels`: passed
- five repeated isolated runs: passed
- complete Ogler test file: passed
- complete file under xdist: passed
- separated-worker order checks: passed
- `git diff --check`: passed

The full pre-push pytest suite was bypassed on this push because
current `main` has an independent failure in
`tests/app/test_apping.py::test_consoler` (bytes-vs-str TypeError).
All other hooks ran normally. That unrelated defect is being addressed
in a separate PR.

## Scope

Test-only change. No Keripy or HIO runtime behavior is modified.

## Relationship to #1490

The failure was discovered by PR #1490's macOS CI run but is independent of
that workflow-only change.

### DIFF
--- a/tests/help/test_ogling.py
+++ b/tests/help/test_ogling.py
@@ -242,7 +242,7 @@ def test_ogler():
     assert not os.path.exists(path)
     assert ogler.opened == False
 
-    help.ogler = ogling.initOgler(prefix='keri')  # reset help.ogler to defaults
+    help.ogler = ogling.initOgler(prefix='keri', syslogged=False)  # restore production config
     """End Test"""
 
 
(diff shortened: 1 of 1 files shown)


dhh1128 (commented):
I spent time tracing this, and I think the diagnosis is exactly right. keripy sets the package global to `syslogged=False` in `help/__init__.py`, but the old `test_init_ogler` and `test_reset_levels` asserted 2/3 handlers on non-Windows, which only holds when `syslogged=True`. As far as I can tell they were quietly depending on an earlier test's cleanup line (`initOgler(prefix='keri')`, i.e. HIO's `syslogged=True` default) leaking into `help.ogler` — so under xdist a worker that gets the pristine global hits `assert 1 == 2`. Reinitializing to `syslogged=False` at the start of each of these tests, and asserting handler membership rather than counts, is the right way to make them self-contained. Doing the reset at the start also means a failure in a preceding test can't corrupt the next one, and I like that you left the locally-constructed-ogler tests (`test_openogler`, `test_ogler`) alone, since their counts are still valid.

I should flag a conflict of interest: I have an open PR (#1522) going after the same #1521 flakiness from a slightly different angle, so there's real overlap. I don't think we want both to merge, and I find your approach here — resetting `syslogged=False` up front and asserting membership instead of magic counts — cleaner than mine. I'd be glad to close or rework mine so we consolidate on one.

One thing I wanted to ask about: in `test_reset_levels` the fix removes the old checks that logged through a stale `logger` and asserted the new file stayed empty (the "old path on file handler" cases). I follow the reasoning in your comment about `getLogger()` returning the same named logger and re-adding handlers each call, so I understand why those assertions were awkward. I'm a little hesitant to lose that behavior coverage entirely — would it make sense to keep a smaller version of it, or do you think it was testing an artifact of the shared-logger setup rather than real behavior? I might be missing something, but wanted to raise it before it goes in.

The syslog branches deserve a note, though nothing needs to change: since these tests now force `syslogged=False`, the `if ... .syslogged and platform.system() != "Windows"` syslog branches never execute in our config, so they're documentation of intent rather than exercised assertions. That's fine by me — I'm just noting it so it's a conscious choice.

dhh1128:
For the record, this fixes #1521 — please add `Fixes #1521` to the PR description so the issue auto-closes on merge. I've closed my own #1522 in favor of this approach.

One thing I'd still like to see before this goes in: in `test_reset_levels` the change removes the old checks that logged through a stale `logger` and asserted the newly-opened file stayed empty (the "old path on file handler" cases). I follow the reasoning about `getLogger()` returning the same named logger, but I'm hesitant to lose that behavior coverage entirely. Could you keep a smaller version of that staleness assertion rather than dropping it outright? If you're confident it was only testing an artifact of the shared-logger setup rather than real behavior, say so and I'll defer — I just don't want it to disappear silently.

ryan-hansen:
This got picked up in PR 1548.  Closing this one.

Sorry...saw the other one first.

PR #1511, opened 2026-07-15
fix: write Consoler output as bytes

jaelliot:
## Summary

Fixes `Consoler` output paths that passed `str` values to HIO's byte-oriented
`Console.put()` API.

## Problem

`hio.core.serial.serialing.Console.put()` delegates directly to `os.write()` and
requires bytes. Several empty/invalid-command responses in
`keri.app.apping.Consoler.recur()` passed strings, resulting in:

```
TypeError: a bytes-like object is required, not 'str'
```

The existing test used the real interactive console and was documented as
manual-only, so this behavior was not covered deterministically.

## Changes

- send bytes on every `Consoler` output path
- add a `FakeConsole` class for deterministic test coverage
- test valid, empty, whitespace, and invalid commands
- verify all emitted output is bytes
- remove dependence on a real terminal for automated coverage

## Validation

- `tests/app/test_apping.py`: passed
- repeated `test_consoler`: 5/5 passed
- `tests/app/test_apping.py` + `tests/app/test_directing.py`: passed
- `git diff --check`: passed
- ruff: all checks passed

The full pre-push pytest suite was bypassed on this push using the same
documented baseline-failure exception as the Ogler fix (PR #1510).  All
other hooks ran normally.

## Scope

Focused Keripy runtime and test correction. No HIO changes.

### DIFF
--- a/src/keri/app/apping.py
+++ b/src/keri/app/apping.py
@@ -78,7 +78,7 @@ class Consoler(doing.Doer):
         chunks = line.lower().split()
 
         if not chunks:
-            self.console.put("Try one of: l[eft] r[ight] w[alk] s[top]\n")
+            self.console.put(b"Try one of: l[eft] r[ight] w[alk] s[top]\n")
             return False
         verb = chunks[0]
 
@@ -95,11 +95,13 @@ class Consoler(doing.Doer):
             command = ('stop', '')
 
         else:
-            self.console.put("Invalid command: {0}\n".format(verb))
-            self.console.put("Try one of: t[urn] s[top] w[alk]\n")
+            self.console.put(b"Invalid command: " + verb + b"\n")
+            self.console.put(b"Try one of: l[eft] r[ight] w[alk] s[top]\n")
             return False
 
-        self.console.put("Did: {} {}\n".format(command[0], command[1]).encode("utf-8"))
+        self.console.put(b"Did: " +
+                         command[0].encode("utf-8") + b" " +
+                         str(command[1]).encode("utf-8") + b"\n")
 
         return False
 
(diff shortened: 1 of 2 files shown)


dhh1128 (commented):
This looks right to me, and I appreciate that it replaces the manual-only test with something deterministic. I confirmed the failing paths: `serialing.Console.put()` ultimately calls `os.write()`, so the `str` arguments at the old `apping.py:68`, `:85`, and `:86` would have raised `TypeError` at runtime, while the already-encoded `:89` was the only path that worked. The new byte outputs all line up with the assertions, including the trailing space in `b"Did: stop \n"` and `str(command[1])` handling the integer walk case, so that all seems solid.

One small thing I noticed, more to confirm intent than to object: the invalid-command message on line 86 changed from `"Try one of: t[urn] s[top] w[alk]\n"` to `"Try one of: l[eft] r[ight] w[alk] s[top]\n"`. I actually think that's an improvement, since the parser branches on `r`/`l`/`w`/`s` and the old "t[urn]" hint didn't match any accepted verb — but since it's a user-facing string I wanted to make sure it was a deliberate cleanup rather than an accidental copy from the line 68 message. Assuming it was intentional, this all reads well to me.

dhh1128 (approved):
Approving — having traced it, I'm satisfied this is correct: sending bytes on every path fixes a real runtime `TypeError` (`serialing.Console.put()` ends up in `os.write()`, which won't take `str`), and the new `FakeConsole` gives it genuine deterministic coverage in place of the old manual-only test. My question about the reworded invalid-command help string was just a confirm-intent note rather than a blocker, so I don't want it to stand in the way. Thank you.

PR #1514, opened 2026-07-16
Docstring Clean up

KeaxD:
The changes are mostly:
- Fix Identations and spacing
- Replace `Args `with `Parameters` inside docstrings. Currently, `Parameters` should now be used accross the codebase instead of `Args` inside docstrings
- Add `Returns` section that was missing in some functions
- Fix some malformed sections headers like `Returns::`

Test:

Ran `sphinx-build -b dirhtml docs/ docs/_build/dirhtml` with 611 warnings against main
Ran `sphinx-build -b dirhtml docs/ docs/_build/dirhtml` with 473 warnings after


src/keri/app/habbing.py
```diff
@@ -3004,7 +2909,7 @@ def incept(self, *, secrecies=None, iridx=0, code=MtrDex.Blake3_256,
             algo (str or None): Key-creation algorithm code used by the key
                 manager when generating new keys.
             salt (str or None): qb64 salt for randomisation when the salty
-                algorithm is used.
+            algorithm: used.
```
ryan-hansen: Surely this is not right.

src/keri/app/habbing.py
```diff
@@ -1905,14 +1850,13 @@ def replay(self, pre=None, fn=0, gvrsn=Vrsn_1_0, *, version=None):
 
         Parameters:
             pre (str or None): qb64 str or bytes of identifier prefix.
-                Default is own ``.pre``.
+            Default: own ``.pre``.
```
ryan-hansen: Also not right.

src/keri/app/habbing.py
```diff
@@ -2330,13 +2257,13 @@ def loadEndRole(self, cid, eid, role=Roles.controller, framed=False,
...
 
-        Parameters::
+        Parameters:
             url (str): url of endpoint. May have scheme missing or not. An
                 empty url nullifies the location.
             eid (str or None): qb64 of endpoint provider to be authorized.
                 None means use own ``.pre``.
             scheme (str): url scheme; must match scheme in url if present.
-                Default is ``"http"``.
+            Default: ``"http"``.
```
ryan-hansen: Also not right.

ryan-hansen (changes requested):

KeaxD:
Docstrings still need more work for sure but I think doing a PR per file would be easier for review. 
- Some docstrings are inaccurate to their functions
- There is an inconsistency in parameters sections where params aren't using colons like `serder: Instance of Serder` and instead are verbosely described like `serder is a Serder Instance`

ryan-hansen (approved):

PR #1516, opened 2026-07-20
Fix Poller keeping caller-supplied msgs deck

### LINKED ISSUE #1500: Poller.__init__ inverted None-check sets self.msgs = None when a msgs deck is passed
### Version

`main` @ `39d2df02` (v2.0.0-dev6) (re-verified at this commit; code permalinks below pin `d31be826`, where the line is identical). Also verified present in release tag `1.3.5` (`indirecting.py:762`), maintenance branch `v1.2.14` (`indirecting.py:753`), and the 1.1 line (`indirecting.py:745` at branch `1.1.45`) — long-standing, not a recent regression.

### Affected code

[`src/keri/app/indirecting.py#L780`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L780) (verified at commit `d31be826`):

```python
def __init__(self, hab, witness, topics, msgs=None, retry=1000, version=None, gvrsn=None, kind=None, **kwa):
    ...
    self.msgs = None if msgs is not None else decking.Deck()
```

The conditional is inverted relative to the idiom used everywhere else in the codebase (e.g. `WitnessPublisher.__init__`, [`src/keri/app/agenting.py#L632`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/agenting.py#L632)):

- `msgs=None` (caller omits it) → `self.msgs = decking.Deck()` — works, by accident of the inversion.
- `msgs=<Deck>` (caller supplies a deck) → `self.msgs = None` — the supplied deck is silently discarded.

### Reproduction

```python
from hio.help import decking
from keri.app import habbing, indirecting

with habbing.openHab(name="test", transferable=True, temp=True) as (hby, hab):
    msgs = decking.Deck()
    poller = indirecting.Poller(hab=hab,
                                witness="BD8-gMSJ6K1PQ7_gG5ZJn2NfHYXehzYuieKD8-ofrY2k",
                                topics=["/receipt"],
                                msgs=msgs)
    assert poller.msgs is msgs   # AssertionError: poller.msgs is None
```

If such a poller is actually scheduled, it crashes as soon as a mailbox event arrives, in `Poller.eventDo` at [`src/keri/app/indirecting.py#L862`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L862):

```python
self.msgs.append(msg.encode("utf=8"))   # AttributeError: 'NoneType' object has no attribute 'append'
```

and `MailboxDirector.processPollIter` would likewise fail at [`src/keri/app/indirecting.py#L679`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L679) (`poller.msgs.popleft()`).

### Expected behavior

Passing `msgs=<Deck>` should wire the poller to deliver received mailbox messages into the caller's deck, exactly as the docstring parameter and the constructor pattern used by every other doer in the codebase (`msgs if msgs is not None else decking.Deck()`) imply.

### Actual behavior

Passing `msgs` yields `self.msgs is None`; the first received event raises `AttributeError`. The bug is masked in-tree because all internal call sites (`indirecting.py` lines [647](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L647), [655](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L655), [663](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L663)) omit `msgs` and therefore take the accidental-default branch.

### Suggested fix

```diff
--- a/src/keri/app/indirecting.py
+++ b/src/keri/app/indirecting.py
@@ class Poller(doing.DoDoer):
         self.version = version
         self.gvrsn = version if gvrsn is None else gvrsn
         self.kind = kind
-        self.msgs = None if msgs is not None else decking.Deck()
+        self.msgs = msgs if msgs is not None else decking.Deck()
         self.times = dict()
```

### Severity

Low impact today (no in-tree caller passes `msgs`), but the parameter is part of the public constructor signature and fails unconditionally whenever used. External code that wires a shared deck into `Poller` gets a guaranteed `AttributeError`. One-line fix.

Note: the dormant `feat/wit-mbx-repackage` refactor branch copies this inverted conditional verbatim into its new `src/keri/mailbox/poller.py` (line 45), so the fix should be mirrored there if that refactor is revived.

This was found during an internal code audit of a downstream project that embeds keripy.

SAY-5:
`Poller.__init__` had an inverted None-check: `self.msgs = None if msgs is not None else decking.Deck()`. When a caller passes their own deck, it was discarded and `self.msgs` became `None`, so the later `self.msgs.append(...)` in `eventDo` raises `AttributeError`. This flips it to `msgs if msgs is not None else decking.Deck()`, matching the idiom every other class in `agenting.py`/`watching.py` already uses.

Added a regression test alongside the existing `QueryEnd` injection test (I couldn't get the full suite building locally because of the pysodium/libsodium native dep, but the test mirrors that neighbor and CI will run it).

Closes #1500

### DIFF
--- a/src/keri/app/indirecting.py
+++ b/src/keri/app/indirecting.py
@@ -777,7 +777,7 @@ class Poller(doing.DoDoer):
         self.version = version
         self.gvrsn = version if gvrsn is None else gvrsn
         self.kind = kind
-        self.msgs = None if msgs is not None else decking.Deck()
+        self.msgs = msgs if msgs is not None else decking.Deck()
         self.times = dict()
 
         doers = [doing.doify(self.eventDo)]
--- a/src/keri/app/indirecting.py
+++ b/src/keri/app/indirecting.py
@@ -859,7 +859,7 @@ class Poller(doing.DoDoer):
                         logger.error(f"bad mailbox event: {evt}")
                         continue
 
-                    self.msgs.append(msg.encode("utf=8"))
+                    self.msgs.append(msg.encode("utf-8"))
                     yield self.tock
 
                     witrec.topics[tpc] = int(idx)
--- a/tests/app/test_indirecting.py
+++ b/tests/app/test_indirecting.py
@@ -486,6 +486,23 @@ def test_query_end_reuses_injected_reger():
     assert qe.reger is mock_reger
 
 
+def test_poller_keeps_supplied_msgs_deck():
+    """Poller must keep a caller-supplied msgs deck instead of discarding it.
+    See #1500.
+    """
+    from keri.app.indirecting import Poller
+    from unittest.mock import MagicMock
+
+    mock_hab = MagicMock()
+    msgs = decking.Deck()
+
+    poller = Poller(hab=mock_hab, witness="wit", topics=["/receipt"], msgs=msgs)
+    assert poller.msgs is msgs
+
+    poller = Poller(hab=mock_hab, witness="wit", topics=["/receipt"])
+    assert isinstance(poller.msgs, decking.Deck)
+
+
 if __name__ == "__main__":
     test_mailbox_iter()
     test_qrymailbox_iter()


dhh1128 (commented):
This looks right to me, and thanks for adding the test that pins down both branches. As I read it, the old `None if msgs is not None else decking.Deck()` meant that whenever a caller actually passed in a deck we threw it away and stored `None`, and then `self.msgs.append(...)` in `eventDo` would have blown up with an `AttributeError` the first time a mailbox event arrived, so this also closes a latent crash and not just a discarded-argument bug. The fix reads cleanly.

One tiny, totally-optional aside since it's on the adjacent line and not part of your change: `msg.encode("utf=8")` in `eventDo` has an `=` where a `-` was probably intended. It happens to work because Python normalizes the codec name to `utf_8`, so it's harmless, but if you felt like fixing it while you're in here I wouldn't object. Either way I'm happy with this as-is.

dhh1128 (approved):
Approving this — it's a clean, correct one-line fix, and the added test pins down both the supplied-deck and default-deck branches, which is exactly what I'd want to see here. My earlier note about the adjacent `utf=8` codec string is entirely optional and shouldn't hold this up. Thanks for catching and fixing this.

PR #1517, opened 2026-07-20
Add V2 IPEX builders and linear dispatcher

KeaxD:
- Summary:
  - implement V2-native IPEX builders: `apply`, `offer`, `agree`, `grant`, `admit`, `spurn`
  - add a basic linear `IpexHandler` dispatcher and `loadHandlers` route registration
  - replace V1-style embedding with V2 `exchange` messages plus nested substream framing

- Changes:
  - build all outer IPEX messages as signed V2 `exn` streams
  - carry `acdc`, `iss`, and `anc` as nested substreams instead of `specialExchange`/embedded payloads
  - add nested-stream normalization so bare artifacts and already-nested artifacts both parse correctly
  - support both transferable and non-transferable signing paths for the outer exchange
  - export the new V2 IPEX API from `keri.acdc`

- Dispatcher behavior:
  - enforce the current linear flow rules for `apply -> offer -> agree -> grant -> admit`
  - allow `spurn` for supported prior verbs
  - reject out-of-order replies and duplicate responses to the same prior

- Tests:
  - add V2-focused `tests/acdc/test_ipexing.py`
  - cover builder parseability, nested artifact handling, linear dispatch, spurn flow, non-transferable signing, and wrong-prior rejection

### DIFF
--- a/src/keri/vc/protocoling.py
+++ b/src/keri/vc/protocoling.py
@@ -456,7 +456,7 @@ def ipexSpurnExn(hab, message, spurned, version=None, pvrsn=None,
     Parameters:
         hab(Hab): identifier environment for issuer of credential
         message(str): Human readable message regarding the admission
-        spurned (Serder): apply, offer, agree or grant received
+        spurned (Serder): apply, offer, or agree received
         version (Versionage | None): explicit KERI protocol version override.
                         When omitted, defaults to the habitat's established version.
         pvrsn (Versionage): explicit KERI protocol version
--- a/src/keri/vc/protocoling.py
+++ b/src/keri/vc/protocoling.py
@@ -456,7 +456,7 @@ def ipexSpurnExn(hab, message, spurned, version=None, pvrsn=None,
     Parameters:
         hab(Hab): identifier environment for issuer of credential
         message(str): Human readable message regarding the admission
-        spurned (Serder): apply, offer, or agree received
+        spurned (Serder): apply, offer, agree or grant received
         version (Versionage | None): explicit KERI protocol version override.
                         When omitted, defaults to the habitat's established version.
         pvrsn (Versionage): explicit KERI protocol version
--- a/src/keri/acdc/__init__.py
+++ b/src/keri/acdc/__init__.py
@@ -9,3 +9,5 @@ from .messaging import (regcept, blindate, update, acdcmap, acdcatt, acdcagg,
                        sectionate,
                        actSchemaDefault, acgSchemaDefault, acmSchemaDefault)
 
+from .ipexing import (Ipex, IpexHandler, loadHandlers, apply, offer, agree,
+                      grant, admit, spurn)
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -28,7 +28,7 @@ PreviousRoutes = {
     Ipex.agree: (Ipex.offer,),
     Ipex.grant: (Ipex.agree,),
     Ipex.admit: (Ipex.grant,),
-    Ipex.spurn: (Ipex.apply, Ipex.offer, Ipex.agree, Ipex.grant),
+    Ipex.spurn: (Ipex.apply, Ipex.offer, Ipex.agree),
 }
 
 
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -6,3 +6,539 @@ IPEx protocol service support (Issuance and Presentation Exchange)
 
 """
 
+import os
+from collections import namedtuple
+
+from hio.help import ogler
+
+from .. import Kinds
+from ..kering import Colds, Vrsn_2_0, sniff
+from ..core import (Counter, Codens, Diger, GenDex, Number, Serdery, Texter,
+                    exchange, messagize)
+from ..peer import cloneMessage
+
+logger = ogler.getLogger()
+
+Ipexage = namedtuple("Ipexage", "apply offer agree grant admit spurn")
+Ipex = Ipexage(apply="apply", offer="offer", agree="agree",
+               grant="grant", admit="admit", spurn="spurn")
+
+PreviousRoutes = {
+    Ipex.offer: (Ipex.apply,),
+    Ipex.agree: (Ipex.offer,),
+    Ipex.grant: (Ipex.agree,),
+    Ipex.admit: (Ipex.grant,),
+    Ipex.spurn: (Ipex.apply, Ipex.offer, Ipex.agree, Ipex.grant),
+}
+
+
+def _streamSerder(stream):
+    """Extract the message serder from a bare or nested artifact stream.
+
+    Parameters:
+        stream (Serder | bytes | bytearray): Artifact body, artifact stream, or
+            already-parsed serder-like object.
+
+    Returns:
+        Serder: Deserialized message serder for the provided artifact.
+    """
+
+    # if the input already looks like a serder with .said and .raw, just return it
+    if hasattr(stream, "said") and hasattr(stream, "raw"):
+        return stream
+
+    # If the input has .raw, use that; otherwise treat the input itself as bytes
+    ims = bytearray(stream.raw) if hasattr(stream, "raw") else bytearray(stream)
+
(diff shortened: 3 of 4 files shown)


dhh1128 (changes requested):
Thanks for doing this. Going through the diff, I came away convinced by the shape of it: carrying acdc, iss, and anc as nested CESR substreams inside one signed v2 exn is cleaner than the V1 approach, and the dispatcher is easy to follow. A few things I'd want to settle before it merges. I've probably misread some of it, so push back where I have.

The one I care most about is the spurn prior-set. In V2, PreviousRoutes[spurn] is (apply, offer, agree) (ipexing.py:26-32) — no grant. V1 still allows grant (protocoling.py:17-23), and the only V1 change here is the docstring at protocoling.py:456, which now describes behavior the V1 code doesn't have. So V1 disagrees with itself, and V1 and V2 disagree about whether a grant can be spurned. What worries me is downstream: when a V2 verify() returns False, processEvent drops the message without storing it or telling the sender (exchanging.py:169-172), so a recipient can no longer refuse an unwanted granted credential nonrepudiably, and a V1 or signify-ts peer that spurns a grant is accepted on one side and silently rejected on the other. test_ipexing.py:409-434 pins the current behavior. Narrowing it might make sense — "a grant is answered by admit, never spurn" is a reasonable rule. If that's the intent, could we say so and gate it on version? And either way, pin the canonical set in one place, so V1's dict and its docstring agree and V2 either matches or diverges on purpose. Most of my other questions fall out of this one.

Next, the nested-substream framing. The whole carriage format — outer exn plus nested substreams, the _isNestedStream / _normalizeNestedStream / _streamSerder helpers (ipexing.py:78-151), consumed through messagize(nests=) (eventing.py:1559-1563) — lives only in code, with nothing normative I could find. _normalizeNestedStream takes either a bare artifact or an already-nested one and tells them apart by matching a leading Counter.name against four body-group codes; anything it doesn't recognize it treats as bare and re-wraps rather than refuses. My worry is interop: another implementation that frames a little differently gets silently re-framed instead of cleanly rejected, and silent mis-framing is painful to debug across implementations. Could we write the framing down as a normative shape, and make an unrecognized leading frame a hard, version-gated reject instead of a re-wrap? keripy is the reference implementation, so the code may be the spec for now — I'd still want reject-on-unknown even if the prose comes later.

A few smaller things:

- verify() parses the route two ways. Incoming routes go through route.split('/') with a length-3 check (ipexing.py:229-234); the prior's verb comes from os.path.basename(os.path.normpath(proute)) (:258-260), which tolerates trailing or doubled slashes. So a stored prior /ipex/offer/ matches, while the same string as a fresh message is rejected. One parser for both?

- The non-transferable signing path looks untested. The cigars branch in _sign is there (ipexing.py:171-188), but both tests build a transferable hab (test_ipexing.py:89,285), so that branch — and messagize framing under a non-transferable signer — isn't exercised. Add a test with a non-transferable hab carrying nested artifacts, or soften the claim if it's meant for later?

- Two IPEX handlers now bind the same routes. The new loadHandlers in ipexing.py and the existing one in protocoling.py both register /ipex/*, with the PreviousRoutes disagreement above and no migration note. Which handler owns which pvrsn, and is registering both on one Exchanger allowed or a mistake?

Two I'm unsure about:

- How is a nested artifact bound to the signed body? _sign signs serder.raw (ipexing.py:153-188), and the nested acdc/iss/anc ride along as attachment bytes; verify and handle take attachments=None and, as far as I can see, never recompute nest.serder.said against the signed a.acdc / a.iss / a.anc. It round-trips fine, but the binding then rests on the SAID in the signed body, not the bytes — so a relay could swap a nested acdc for another well-formed one and the outer signature would still check out. I'd guess that's fine if the credential-ingestion pipeline recomputes and enforces each nested SAID on store. Does that recheck exist somewhere I missed? If so, this is just a doc note.

- Is counterparty role enforced anywhere? verify() binds the prior by SAID and checks ordering, but I didn't see it confirm that an admit comes from the party the grant was addressed to. So a well-signed third party holding a valid prior digest could advance an exchange it isn't part of — unless the Exchanger or a caller enforces role identity upstream. Do you know where that lives? If it's handled, ignore this.

A couple of things I chased and decided not to raise. I looked at whether the structural-only prior-binding could let a forged offer confer authority, and don't think it can — IPEX here is bookkeeping, not the authorization gate; issuance authority still binds through processEvent's signature checks and the credential edges, and a wedged exchange recovers with a fresh apply. I wondered whether the verification half was left unwired, and don't think so — V1's handler is the same notifier-only plumbing, and real credential verification lives in the Verifier/Tever subsystem at admit time either way. And the except AttributeError around the flow guard fails closed, not open: the dereferences raise KeyError/TypeError, which propagate to a reject.

I'm marking this request-changes mainly so the spurn decision gets made on purpose rather than by omission. This is going to be a valuable contribution, and I'd be glad to be wrong where I'm guessing. Thanks again.

KeaxD:
@dhh1128  Thank you for your comment. It was really helpful, I'll answer the spurn point right now: I was also wondering about that because in the presentation Sam showed us during our meeting last Tuesday, one of the slides, showed that spurn was only used on apply/offer/agree however when looking at v1 docstrings it does indeed include grant. 
I asked Sam about it, so I'll just paste his reply here:

The grant should not be spurnable. It think that docstrings is an error.
The agreement is binding. Otherwise you could have race condition where the receipient (disclosee) tries to change their mind but the damage is done one the grant is sent. There is no way to pull it back and trust the receipient. 

So this is why I removed grant

dhh1128:
@KeaxD : that makes sense on the spurn. Can we update the comments or docstring to make it clear that this is a v2 change? Possibly we should have a test that makes the behavior change explicit?

dhh1128:
Building the real-IPEX companion on these builders, I ran into a gap: `offer`, `agree`, `admit`, and `spurn` never pass `receiver=`, so their `rp`/`ri` come out empty, while `apply` and `grant` thread `receiver=recp`. That leaves the responder half of an exchange unaddressed — the messages don't name who they're for. It came up in #1542, and Sam's reply there confirms `ri` shouldn't be empty: an authenticatable negotiation needs a named sender/receiver pair on every leg, not just the two verbs that already take a `recp`.

It can't be patched caller-side, since `exchange()` SAIDifies the envelope before I get the serder back — the receiver has to be set inside the builder. The good news is each of these four already receives the prior it answers, and the party to address is just that prior's sender:

```python
def offer(hab, message, acdc, apply=None, recp=None, ...):
    receiver = recp if recp is not None else (apply.ked["i"] if apply is not None else "")
    serder = exchange(sender=hab.pre, receiver=receiver, prior=..., route="/ipex/offer", ...)

# agree  -> offer.ked["i"]
# admit  -> grant.ked["i"]
# spurn  -> spurned.ked["i"]
```

The optional `recp` covers the one case with no prior — an offer-first bootstrap, where the discloser opens the exchange. Everything else derives, so the receiver falls out symmetrically and there's nothing for a caller to get wrong. I'm raising a small PR against your branch with this change, in the hope it's helpful here — no attachment to my shape if you'd rather handle it differently.

src/keri/vc/protocoling.py
```diff
@@ -456,7 +456,7 @@ def ipexSpurnExn(hab, message, spurned, version=None, pvrsn=None,
     Parameters:
         hab(Hab): identifier environment for issuer of credential
         message(str): Human readable message regarding the admission
-        spurned (Serder): apply, offer, agree or grant received
+        spurned (Serder): apply, offer, or agree received
```
ryan-hansen: PreviousRoutes still allows spurn of grant for v1, so dropping “grant” from this v1 docstring makes the docs inaccurate. I would leave the v1 behavior alone and make the docs explicit: v1 still accepts grant; v2 does not.  If we want to talk about making v1 behave the same way, that's probably a separate discussion and separate PR.

ryan-hansen (changes requested):

ryan-hansen (approved):
I'm going to approve and merge this as is, with a follow on task to persist and validate nested streams.

https://github.com/orgs/keri-foundation/projects/2/views/7?pane=issue&itemId=219366551&issue=keri-foundation%7CDevelopment%7C62

PR #1519, opened 2026-07-21
V1 -> V2 Cleanup

ryan-hansen:
- Address some remaining hard-coded v1 fallbacks with proper gvrsn passing
- Remove "convenience" code for handling version-related params in tests
- Update test param handling for consistency across the test suite


dhh1128 (commented):
Not asking for changes, just notes from a close read for whenever you come back to this. Deriving the attachment genus from each event's own `serder.pvrsn` instead of a hard `Vrsn_1_0` fallback is the right direction, and the v1 happy path stays byte-stable under the existing tests.

The one worth a closer look is the `if gvrsn is Version` gate in `replay` (habbing.py:1921-1924). Omitting `gvrsn` passes the module singleton, which caps down to the target's `pvrsn`; but an explicit `Vrsn_2_0`, or a freshly built `Versionage(2,0)`, is a distinct object, so it skips the cap, and since `messagize` floors up and never down (eventing.py:1597-1599), a v1 event then goes out with v2 counter codes. Two callers that both think they asked for v2 get different wire genus depending on whether they passed the singleton or a value-equal object. There's a real case for the identity check being deliberate — the sentinel's value coincides with `Vrsn_2_0` on purpose, and `is` is the only way to tell "auto-detect" from "force v2", which makes an explicit `Vrsn_2_0` against a v1 target arguably caller error rather than a bug. If that's the intent, a docstring line saying so, and maybe rejecting a v2 `gvrsn` on a v1 target instead of emitting mixed framing, would keep the next person from tripping over it. Nothing in-tree hits this today.

A couple I'm less sure about:

- `replay` over-framing a delegation leg: a v2 delegate replaying a still-v1 delegator's KEL would stamp v2 genus on the delegator's v1 events, since the `pvrsn` comes from the delegate `pre` (habbing.py:1923-1924) and `cloneDelegation`'s per-kever `if gvrsn is None` path (basing.py:1915-1916) is skipped. I couldn't convince myself it's actually reachable — a mixed v1-delegator/v2-delegate chain may not be a supported config, and `cloneEvtMsg` uses `genusify=False`, so I may be chasing something latent. Worth a look if mixed-version delegation is in scope.
- The refactor swapped `gvrsn if gvrsn is not None else Vrsn_1_0` for `kwa.get('gvrsn', Version)` (habbing.py:2523, 3590, 3604, 3608). A present-but-`None` `gvrsn` now returns `None` and rides down into `messagize`, which dereferences `gvrsn.major` and raises `AttributeError` instead of the old graceful v1 fallback. Nothing in-tree passes an explicit `None`, but KERIA/agent kwargs could; `loadLocScheme` still tolerates it.
- In `replyToOobi`/`replyEndRole`, a leftover `gvrsn = kwa.get('gvrsn')` (habbing.py:2521-2523) still feeds `loadLocScheme` while the `replay` calls in the same body use `kwa.get('gvrsn', Version)`. Same value today, but two resolutions in one method read as a trap.

Last one is about the diff rather than the code. The behavioral change is about twenty lines in `replay`, but inlining the deleted `tests/common.py` helpers spreads roughly nine hundred lines across thirty-odd test files, which buries the wire-affecting default flip inside a large cleanup and turns the next v1-to-v2 test flip into a 43-file edit rather than a one-line change to a shared constant. Splitting the behavioral fix into its own small PR would make it much easier to review, if you're open to it — I can see the greppability argument for explicit per-test versions cutting the other way.

I'm inferring intent in a few places, so tell me where I've misread the flow.

SmithSamuelM:
@dhh1128 @ryan-hansen 

> Deriving the attachment genus from each event's own serder.pvrsn instead of a hard Vrsn_1_0 fallback is the right direction, and the v1 happy path stays byte-stable under the existing tests.

I don't think that should be the default rule. Here is why.

The embedded primitives in a message are setup before any attachments.  The message constructors have both a pvrsn and gvrsn so that the message constructor can control which version of primitives are used. This has nothing to do with the attachments. It has everything to do with is one talking to a v1 only Parser on replay or a V2 parser.

The attachments to a message can be a completely different gvrsn. The only reason to use old version attachments is because you are sending messages to a node that is using a v1 parser.  If we default to using v1 attachments for every  v1 message we run into a an upgrade path problem for key event messages.

Key event messages, because they are signed and must be verifiable indefinitely, can't be upgraded. The parser has to respect that internal pvrsn and gvrsn of the message forever.  We store the raw serialization of the v1 Key event message so the primitive serializations embedded in it never change. The pvrsn and gvrsn is for the parser. 


But for attachments, there should be no reason that attachments can't be version 2 even when the message is V1, except when talking to a V1-only parser.  And we want to deprecate v1 only parsers as fast as possible. 

So if someone is running v2 and talking to someone else running v2, there is no valid use case for v1 attachments,
Attached substreams can switch versions.

Its only if someone needs to replay to a v1 only parser that the attachments should be v1. And if its a v1 only parser then one can't talk to it with messages that were generated v2. There is no interoperability path for v2 messages with v1 only parsers.. 
Which is why we want to migrate to v2 parsers ASAP.

The branches that are stuck on v1 parsers need to accelerate their migration, because there is no backward-compatible path for new v2 key event messages. The signatures won't verify on a v1 conversion of the message.  So if the parser recognizes v2 messages it also recognizes V2 attachments.  

The other messages  (non key event messages) are a mixed bag, if doing a multi-party exchange then all parties have to use the same version for the message. In a two party interaction existing v2 allows one to force v1 for both messages and attachments.  

But understand that for key event messages, it should not be the case that v1 key event messages require v1 attachments unless the parser the message is going to is v1 only.  

We have a problem that production v1 implementations have not yet adopted the v2 parser (which has been out for over a year). The goal was that they would backport the v2 parser on main to whatever production branch they were on.  But that didn't happen. 

The v2 parser can handle both v1 and v2 messages and attachments.  So a gradual update migration should start with using the v2 parser.  Even if one only generates new v1 messages, one could then at least interoperate with someone sending one v1 or v2  messages and v2 attachments. 

And one can stage the migration to accepting v2 messages of different types. 

The change between v1 and v2 for Key Event messages is very small and is already encoded in the Kever Class on main. So accepting v2 key events is the first step.  

Then one can implement v2 support for the query and reply messages which also have minimal changes from v1 to v2.  

Exchange messages last since they have the biggest changes and we are still bulding the v2 IPEX on main. 

So the main branch is in transition. 

Handling of v1 or v2 key event messages just works.  
Handling of v1 reply, query, and exchange message messages is still V1 logic. 

Handling of v2 replay, query, and exchange messages is a work in progress.


There were two primitive use changes, moving from seqner instances to number instances and moving from saider instances to diger instances.  The serialization doesnt change in the latter but could change in the former as Number will by default use the most compact primitive code for a number unless its told to do differently such as with the huge method.

ryan-hansen:
The common.py roll-back/inlining can and probably should be a separate PR.  I just got a little lazy with it.  I'll break it out.

dhh1128:
@SmithSamuelM You're right, and I stated that too strongly. The version a message was signed at and the version its attachments are framed at are different questions. The first is fixed forever because the signature has to verify; the second depends only on what parser is on the other end. Coupling attachment genus to the event's `pvrsn` is the wrong default for exactly the reason you give — it strands a v1 key event on v1 attachments even when both ends run a v2 parser that would take v2 framing happily.

Where that coupling actually lives in this PR, in case it helps locate the fix: on this branch `Version` is `(2, 0)`, and `messagize` only ever floors `gvrsn` up toward the serder's version, never down —

```python
if (gvrsn.major < svrsn.major or
        (gvrsn.major == svrsn.major and gvrsn.minor < svrsn.minor)):
    gvrsn = svrsn
```

So the default, with nobody overriding `gvrsn=Version`, already emits v2 attachments — the behavior you're describing. What pulls it back down to v1 is the new line in `replay` —

```python
if gvrsn is Version and pre in self.kevers:
    gvrsn = self.kevers[pre].serder.pvrsn
```

plus the `gvrsn=exn.pvrsn` / `gvrsn=creder.pvrsn` calls in exchanging and credentialing. Those are the override. Take them out and replay goes back to defaulting v2, with the explicit `gvrsn` argument still there for the one case that needs it: replaying to a v1-only parser.

The catch is byte stability. That override is almost certainly what keeps the v1 golden vectors passing — today `replay` over a v1 KEL emits v1 attachments, and a stack of tests assert those exact bytes. Delete it and every one of those vectors has to be regenerated to expect v2 framing. That is the right move if v2-attachments-by-default is the intended wire behavior for v1 replay, but it is a wire-behavior change, not a cleanup. So the one question that decides this PR: do you want `replay` over a v1 KEL to start emitting v2 attachments by default? If yes, the override comes out and the v1 vectors move. If the byte stability is load-bearing for something I'm not seeing, the override stays and earns a comment saying why.

Either way, I'd lift the ~20 lines of behavioral change in `replay` out of the ~900-line test cleanup around it. A flip in default framing is the kind of decision that shouldn't ride inside a 43-file rename where nobody finds it later.

Two smaller things that stand regardless of which way that goes:

- `kwa.get('gvrsn', Version)` only defaults when the key is absent. A present-but-`None` `gvrsn` — which the old `gvrsn if gvrsn is not None else Vrsn_1_0` absorbed — now rides down into `messagize` and raises `AttributeError` on `gvrsn.major`. Nothing in-tree passes an explicit `None`, but KERIA agent kwargs could.
- `replyToOobi` / `replyEndRole` still read `gvrsn = kwa.get('gvrsn')` for `loadLocScheme` while the `replay` calls in the same body use `kwa.get('gvrsn', Version)`. Same value today, two resolutions in one method.

And the delegation case I flagged inverts under your rule: a v2 delegate replaying a still-v1 delegator would stamp v2 genus on the delegator's v1 events, which I called over-framing — but if attachments default to v2, that's the correct output, not a bug. Scratch that one.

dhh1128:
Following up on my own note above, because it interacts with #1535 and because I framed one thing loosely enough that I'd rather correct it before it carries weight.

This PR and #1535 both rewrite v1 golden-master assertions, and they overlap in eight test files (`test_reply`, `test_grouping`, `test_oobiing`, `test_delegating` ×2, `test_bare`, `test_eventing_v1`, `test_proving`). #1535 shows that some of those v1 vectors were themselves wrong — they assert a v2 `Number` where genuine v1 CESR requires a `Seqner`, which is a malformed v1 stream a real keri-1.x node can't parse. So the "byte stability" I leaned on earlier is, in part, stability against an incorrect baseline. Freezing those bytes was freezing a bug.

That makes merge order matter. #1535 is the smaller, self-contained correctness fix — test-first, full suite green — and it establishes the *correct* v1 wire form. Landing it first gives this PR a right baseline to regenerate its vectors against. The other order costs more: if this PR lands first, its regenerated v1 vectors bake the `Number` form back in, and #1535 has to unpick them a second time.

None of that changes the default-genus question you and I are still working out — it just means the v1 side of the ledger should be corrected before either of us freezes it.

ryan-hansen:
Closing this one to make way for two new separate PRs:
- v2-attachments-default - just the attachment wiring
- v2-clean-up-tests - just clean up stuff

PR #1522, opened 2026-07-21
Fix xdist flake in ogling tests from leaked global logger state (#1521)

### LINKED ISSUE #1521: Ogling tests leak global logger state: test_init_ogler / test_reset_levels flake under pytest-xdist (handler count 1 != 2)
## Summary

`tests/help/test_ogling.py::test_init_ogler` and `::test_reset_levels` fail with `AssertionError: assert 1 == 2` (logger **handler count**) when run under `pytest-xdist`. The ogler/logger is a process-global singleton — the test file itself notes *"Because loggers are singletons it still affects loggers"* — and these two tests assert an **absolute** handler count without isolating that global state from sibling tests that xdist may co-schedule into the same worker.

## Reproduction (deterministic here)

```
pytest tests/help -n auto
```
→ `2 failed, 11 passed`:
```
tests/help/test_ogling.py:264: AssertionError: assert 1 == 2   # test_init_ogler
tests/help/test_ogling.py:361: AssertionError: assert 1 == 2   # test_reset_levels
```
(both are `assert len(logger.handlers) == 2`, observed value 1.)

Serial / isolated runs pass:
```
pytest tests/help                    # 13 passed
pytest tests/help/test_ogling.py     # 4 passed (5x, stable)
```

## Why it can bite CI

CI runs `pytest tests/ --ignore tests/demo/ --ignore test/scripts -n auto` (`.github/workflows/python-app-ci.yml`). Whether these tests fail depends on **how xdist distributes tests across workers**: if the ogling tests land in a worker alongside a sibling `tests/help` test that has already mutated the singleton logger's handlers, the handler-count assertions see `1` instead of `2`. Today the full-suite distribution happens to be benign, so CI is green *by luck of scheduling*; restricting to `tests/help -n auto` forces the interfering sibling into the same small pool and makes the failure deterministic. This is a **latent flake**, not a guaranteed failure — exactly the kind that surfaces intermittently as the suite grows or worker counts change.

## Root cause

- `help.ogler` is a module-global singleton, and Python `logging` loggers are global-by-name. Handlers added/removed by one test persist into the next test that shares the worker process.
- `test_init_ogler` (~line 264) and `test_reset_levels` (~line 361) assert `len(logger.handlers) == 2`, which only holds if no earlier test in the worker has changed that singleton's handler set first. Under xdist co-scheduling that precondition isn't guaranteed; under serial execution the load order happens to be benign.

## Fix direction (for discussion)

Make the ogling tests independent of pre-existing global logger state — e.g. an autouse fixture that clears/re-inits the singleton logger's handlers at test start, or assert on the *delta* rather than an absolute handler count. This is analogous in spirit to the per-worker store isolation added in #1496 (which fixed the store-race flake in the same de-flaking pass and explicitly left this logging flake out of scope).

_Found while de-flaking the ACDC worked-examples PR (#1483 / #1496): the store-race flake was fixed in #1496; this global-logging-state flake was deferred as out-of-scope and is filed here with a deterministic reproduction._

dhh1128:
## Summary

Fixes #1521 — a latent `pytest-xdist` flake in `tests/help/test_ogling.py`. `test_init_ogler` and `test_reset_levels` fail with `AssertionError: assert 1 == 2` (logger **handler count**) when xdist schedules them into a worker without a preceding sibling from the same module.

## Root cause

Both tests read the **process-global** `help.ogler` and assert an **absolute** handler count. `getLogger()` produces `consoled(1) + syslogged(1) + (filed && opened)(1)` handlers, so the count is entirely determined by the global ogler's flags.

keripy's real default global is created in `keri/help/__init__.py` with `syslogged=False` (console only → **1** handler). But these two tests assert `2` (unopened) / `3` (opened), i.e. they assume `syslogged=True`. That precondition is only met because an **earlier sibling in the same module** (`test_ogler`, `test_init_ogler`) ends with `help.ogler = ogling.initOgler(prefix='keri')`, which defaults `syslogged=True` and leaks a 2-handler global into the next test.

Under xdist load-scheduling, tests from one module are **not guaranteed** to run in the same worker in definition order, so the leaked precondition can be absent and the assertions see `1 != 2`. Serial runs pass only by luck of load order.

Confirmed deterministically by running just the two tests in isolation (no sibling): `2 failed` with the exact `assert 1 == 2`.

## Fix

Add a function-scoped `keri_ogler` fixture (test file only, **no product code change**) that snapshots `help.ogler`, pins it to a known `syslogged=True` default, and restores the real global on teardown; apply it to the two affected tests. They now establish their own precondition instead of inheriting it, making them independent of sibling scheduling order. Teardown also stops the pre-existing leak of a `syslogged=True` global to later tests.

This mirrors the per-worker store isolation added in #1496 during the same de-flaking pass, which explicitly left this logging flake out of scope.

## Verification

- Two tests in isolation: `2 failed` → **`2 passed`**
- `pytest tests/help -n auto` (issue's deterministic reproduction): was `2 failed, 11 passed` → **`13 passed`**
- `pytest tests/help/test_ogling.py` serial: **`4 passed`** (no regression)

---

_Root-caused and implemented with assistance from Claude (Claude Code)._

### DIFF
--- a/tests/help/test_ogling.py
+++ b/tests/help/test_ogling.py
@@ -9,9 +9,32 @@ import os
 import platform
 import logging
 
+import pytest
+
 from hio.help import ogling
 
 from keri import help
+
+
+@pytest.fixture
+def keri_ogler():
+    """Pin help.ogler to a known syslogged=True default so handler-count
+    assertions don't depend on sibling scheduling order (xdist-safe, #1521).
+
+    The process-global help.ogler is created in keri.help with syslogged=False
+    (console only = 1 handler), but these tests assert the syslogged=True
+    default (console + syslog = 2 handlers, +file = 3 when opened). Serially
+    that precondition happens to be met by a preceding sibling test that resets
+    help.ogler via initOgler(prefix='keri'); under pytest-xdist load-scheduling
+    it is not guaranteed. Establish it here instead, and restore the real
+    global on teardown so nothing leaks to later tests.
+    """
+    saved = help.ogler
+    help.ogler = ogling.initOgler(prefix='keri')  # syslogged=True -> 2 handlers unopened
+    try:
+        yield help.ogler
+    finally:
+        help.ogler = saved
 
 
 def test_openogler():
@@ -246,7 +269,7 @@ def test_ogler():
     """End Test"""
 
 
-def test_init_ogler():
+def test_init_ogler(keri_ogler):
     """
     Test initOgler function for ogler global
     """
@@ -344,7 +367,7 @@ def test_init_ogler():
     """End Test"""
 
 
-def test_reset_levels():
+def test_reset_levels(keri_ogler):
     """
     Test resetLevel on preexisting loggers
     """


dhh1128:
Closing this as redundant. After tracing both, I realized #1510 goes after the same #1521 flake from a cleaner angle — resetting to the production default (`syslogged=False`) up front and asserting handler membership instead of magic counts — and I'd rather consolidate on that one. Nothing depends on this branch, so I'm closing in favor of #1510.

PR #1538, opened 2026-07-23
Fix small klass -> klas typo

kentbull:
Looks like there was an accidental 's' character included in the ESSR database when it was added.

### DIFF
--- a/src/keri/db/basing.py
+++ b/src/keri/db/basing.py
@@ -1163,7 +1163,7 @@ class Baser(dbing.LMDBer):
         # TODO: clean
         self.epath = subing.IoSetSuber(db=self, subkey=".epath")
 
-        self.essrs = subing.CesrIoSetSuber(db=self, subkey=".essrs", klass=coring.Texter)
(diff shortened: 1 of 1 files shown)


dhh1128 (approved):
Confirmed correct. The keyword-only parameter is spelled `klas` (defined in `CesrSuberBase.__init__`, `subing.py:831`); the old `klass=` silently fell into `**kwa` and was swallowed by `SuberBase.__init__`, so `klas` defaulted to `coring.Matter`. That meant the `.essrs` sub-db deserialized ESSR values as `Matter` instead of the `Texter` instances callers write (`exchanging.py:189,284`) and read back (`exchanging.py:237`, `parsing.py:966`).

`klas=coring.Texter` (a single class) is the right shape here — `CesrIoSetSuber` extends `CesrSuberBase`, whose `_des` calls `self.klas(qb64b=val)` on a single class. No migration risk: writes already serialized `Texter.qb64b`, and `Texter` subclasses `Matter` sharing the code tables, so this only fixes the returned type on read.

LGTM.

PR #1541, opened 2026-07-23
regbaser and webregbaser db scaffold

evanja57:
In this PR:
- Add RegBaser and WebRegBaser with the same six sub-databases for LMDB and IndexedDB
- Replace the V1 tvts, states, oots, and taes responsibilities with evts, heads, ooes, and maes
- Store complete registry events in evts, accepted sequence entries in tels, KEL anchor references in ancs, and rebuildable current heads in heads
- Keep tels as the accepted sequence index and ancs as the KEL anchor reference store
- Drop V1 backer, partial witness, timestamp, and unused signature stores that do not apply to blindable registries
- Keep credential and wallet data outside this database; TEL processing goes in registering.py, registrar workflow in registraring.py, and IPEX in ipexing.py
- Remove the duplicate webregbaser.py stub

### DIFF
--- a/src/keri/acdc/webregbasing.py
+++ b/src/keri/acdc/webregbasing.py
@@ -122,6 +122,7 @@ class WebRegBaser(WebDBer):
             clear=clear or self.temp,
             storageOpener=opener,
         )
+        self._version = None
         self.env = self.db.env
 
         _before = set(self.__dict__)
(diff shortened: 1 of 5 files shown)


src/keri/acdc/webregbasing.py
```diff
@@ -1,8 +1,225 @@
...
+            try:
+                delattr(self, name)
+            except AttributeError:
+                pass
+
+        try:
+            loop = asyncio.get_running_loop()
+            loop.create_task(db.flush())
+        except RuntimeError:
+            pass
```
ryan-hansen: Is there any reasonable scenario where writes have been made but there is no running loop?  In that case, flush would never happen and writes would be dropped silently.  

Maybe in the exception case, do something like this?

```
asyncio.run(db.flush())
```

evanja57: decided to raise an error if the sync close is called when there are dirty stores. Keeps the lifecycle more explicit that way I think

ryan-hansen (approved):
Looks ok to me.  Definitely want @SmithSamuelM's opinion on it though.

PR #1546, opened 2026-07-24
V2 attachments default

ryan-hansen:
- Default omitted/None/Version attachment genus to v2, even on v1 event bodies; use Vrsn_1_0 only when talking to v1-only peers
- Remove Hab Version → pvrsn caps (replay, reply, msgOwnEvent, msgOtherEvent, group query)
- Unify OOBI / end-role gvrsn resolution (kwa.get("gvrsn") or Version); drop v1 fallbacks in replyToOobi / replyEndRole
- Default cloneDelegation to Version (no derive-from-kever)
- Replace hand-rolled CESR attachment builders with messagize (exchange, signing, grouping, credentialing, Reger TEL clone)
- Thread gvrsn through Reger clonePreIter / cloneTvt / cloneCreds (default v2; parse counters via Codens)
- Tests: new policy regressions (v1 body + v2 attachments); pin remaining v1-peer paths with gvrsn=Vrsn_1_0

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -55,7 +55,7 @@ class Receiptor(doing.DoDoer):
     def receipt(self, pre, sn=None, auths=None):
         """Returns a generator performing witness receipting of KEL events.
 
-        The returns a generator that will submit the designated event to witnesses for receipts using
+        This returns a generator that will submit the designated event to witnesses for receipts using
         the synchronous witness API, then propagate the receipts to each of the other witnesses.
         Delegates to .catchup to catch up any new witnesses to the current state of the KEL.
 
@@ -546,7 +546,7 @@ class WitnessInquisitor(doing.DoDoer):
 
             msg = hab.query(target, src=witer.wit, route=r, query=q, **kwa)  # Query for remote pre Event
 
-            kel = introduce(hab, witer.wit)
+            kel = introduce(hab, witer.wit, gvrsn=self.hby.version)
             if kel:
                 witer.msgs.append(bytearray(kel))
 
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -74,7 +74,9 @@ class Receiptor(doing.DoDoer):
         hab = self.hby.habs[pre]
         sn = sn if sn is not None else hab.kever.sner.num
 
-        msg = hab.msgOwnEvent(sn=sn, framed=True)
+        # Match attachment genus to the event body so v1-only witnesses can parse.
+        serder, _, _ = hab.getOwnEvent(sn=sn)
+        msg = hab.msgOwnEvent(sn=sn, framed=True, gvrsn=serder.pvrsn)
         ser = serdering.SerderKERI(raw=msg)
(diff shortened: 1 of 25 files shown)


src/keri/app/agenting.py
```diff
@@ -74,7 +74,9 @@ def receipt(self, pre, sn=None, auths=None):
         hab = self.hby.habs[pre]
         sn = sn if sn is not None else hab.kever.sner.num
 
-        msg = hab.msgOwnEvent(sn=sn, framed=True)
+        # Match attachment genus to the event body so v1-only witnesses can parse.
+        serder, _, _ = hab.getOwnEvent(sn=sn)
+        msg = hab.msgOwnEvent(sn=sn, framed=True, gvrsn=serder.pvrsn)
```
evanja57, line 79: just leaving this here as a reminder in my review

src/keri/app/forwarding.py
```diff
@@ -604,10 +604,13 @@ def introduce(hab, wit):
                 break
 
     if not found:  # no receipt from remote so pre-send own inception
-        # no vrcs or rct of own icp from remote so send own inception
-        for msg in hab.db.clonePreIter(pre=hab.pre, version=hab.kever.serder.pvrsn):
+        # no vrcs or rct of own icp from remote so send own inception.
+        # Match attachment genus to this hab's event version so v1-only
+        # witnesses can parse the introduction stream.
+        gvrsn = hab.kever.serder.pvrsn
```
evanja57: I think we can remove the explicit gvrsn here and just let the functions use the version default

src/keri/app/habbing.py
```diff
@@ -1664,15 +1664,14 @@ def exchange(self, *,
             end = bytearray()
 
         if gvrsn is None:
-            gvrsn = serder.pvrsn
+            gvrsn = Version
```
evanja57, line 1675: we should set gvrsn from version in the specialExchange branch so that even when its v1 it is correct

src/keri/peer/exchanging.py
```diff
@@ -636,56 +636,37 @@ def serializeMessage(hby, said, framed=False):
...
-            atc.extend(Counter(Codens.TransIdxSigGroups, count=1,
-                                    version=Vrsn_1_0).qb64b)
-            atc.extend(prefixer.qb64b)
-            atc.extend(seqner.qb64b)
-            atc.extend(saider.qb64b)
+    aims = bytearray()
+    if tsgs or cigars:
+        # Authenticator attachments via messagize; framed=True so we can append
+        # pathed embeds after (pathed material is outside messagize support).
+        full = messagize(exn, tsgs=tsgs or None, cigars=cigars or None,
```
evanja57, line 649: since this is on a v1 only function we should pass Vrsn_1_0 to messagize here

src/keri/app/forwarding.py
ryan-hansen: Good call. Had to add and optional gvrsn on introduce so callers can pass their Habery/Poster version through when needed; otherwise default.

PR #1578, opened 2026-08-05
Add tests for tests examples with IPEX

KeaxD:
- Add IPEX-backed companion tests in test_examples.py for:
  - `test_registry_issuance_lifecycle_IPEX_JSON`
  - `test_selective_disclosure_aggregate_IPEX_JSON`
  - `test_partial_disclosure_compaction_IPEX_JSON`
  - `test_blindable_registry_correlation_minimizing_IPEX_JSON`

- Use `keri.acdc.ipexing` flow in each test:
  - `ipexGrant`
  - `ipexAdmit`
  - `loadHandlers`
  - `Exchanger`
  - `Parser`

- Verify that Daniel’s worked-example artifacts can be transported as IPEX `exn` messages while preserving the same ACDC/TEL commitments and SAIDs.

- Assert the privacy properties still hold after transport:
  - selectively withheld fields stay hidden
  - blinded registry states stay unreadable without the correct per-event nonce
  - salts/nonces are not carried on the IPEX wire

### DIFF
--- a/tests/acdc/test_examples.py
+++ b/tests/acdc/test_examples.py
@@ -24,8 +24,12 @@ from jsonschema import Draft202012Validator
 from jsonschema.exceptions import SchemaError, ValidationError
 
 from keri import Vrsn_2_0, Kinds, Protocols, Ilks
-from keri.core import Noncer, Blinder, GenDex, Aggor, Compactor, Diger, DigDex
-from keri.acdc import regcept, blindate, acdcmap, acdcagg
+from keri.core import (Noncer, Blinder, GenDex, Aggor, Compactor, Diger,
+                       DigDex, Parser)
+from keri.acdc import (regcept, blindate, acdcmap, acdcagg, loadHandlers,
+                       grant as ipexGrant, admit as ipexAdmit)
+from keri.app import openHby
+from keri.peer import Exchanger
 
 
 # Spec-aligned example fixtures (see test_acdc_examples_setup in the spec tests).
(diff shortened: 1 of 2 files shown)


tests/acdc/test_examples.py
```diff
@@ -316,6 +516,158 @@ def test_selective_disclosure_aggregate_JSON():
...
+    disclosure against the committed AGID while the withheld values stay off the
+    wire.
+    """
+
+    class Recorder:
+        def __init__(self):
+            self.items = []
+
+        def add(self, attrs):
+            self.items.append(attrs)
```
ryan-hansen: Any reason we're repeating these class definitions in each test?  Shouldn't it be just one Recorder for the module?

tests/acdc/test_examples.py
```diff
@@ -229,6 +233,202 @@ def test_registry_issuance_lifecycle_JSON():
...
+        assert issuedGrantResult.serder.ked['a']['anc'] == ripper.said
+        assert [nest.serder.said for nest in issuedGrantResult.nests] == [
+            acdc.said, issued.said, ripper.said
+        ]
+        issuedGrantWire = bytes(issuedGrant.raw) + bytes(issuedGrantAtc)
+        # The disclosed credential is present on the wire, but the registry state
+        # word itself remains hidden because the nested `bup` is blindable
+        assert b'issued' not in issuedGrantWire
+        assert b'revoked' not in issuedGrantWire
+
```
ryan-hansen: Maybe a good idea as a safety precaution to assert that the salt never gets passed on the wire as well.  Something like this:

```
assert salt.encode() not in issuedGrantWire
```

Same for anywhere we're asserting things are not on the wire.  The code currently doesn't have any way for salt to get in there (on the wire), but just as a protection against some future salt leak regression, can't hurt, right?

tests/acdc/test_examples.py
ryan-hansen, line 1283: Should we add the new tests here as well?

ryan-hansen:
The four IPEX companions all do basically the same stuff: 
- build a grant
- parse it so we can look at the nests
- build wire bytes for the privacy checks
- parse it again through the exchanger so it actually gets stored
- then admit and check that prior link. 
The partial and blindable tests already tuck most of that into a local ipexExn, but the other two still write it out by hand, and the two helpers aren’t quite the same shape (one takes iss/anc, one doesn’t).

Might be worth pulling all that into one module-level helper, something that returns the grant result, wire bytes, and the grant/admit serders, so the structural IPEX plumbing is only in one place and each test just does its artifact setup and applicable privacy/commitment asserts. 

Not a blocker, just less copy-paste if/when the Parser or ipexing pattern shifts.

PR #1580, opened 2026-08-06
feat(parsing): optional fault sink so embedders can tell stream failures apart

dhh1128:
I'm tagging @SmithSamuelM and @pfeairheller , because I think you're the two most likely to have an opinion about this. But I'd love to hear from anyone else.

There's a comment at src/keri/core/parsing.py:894 that reads `except ExtractionError as ex:  # maybe this needs to be more granular`. This PR adds that feature. If the PR were a line or two, I think it would be a no-brainer, but since it requires some real code additions, that make me ask if it's worth the cost. I need the feature but could leave it in my clone of keripy, permanently diverging in this one way from what keripy provides. Or maybe you'll convince me that the whole feature is wrong-headed... This is why I left the PR in draft state for the moment.

Parser knows exactly why it declined a message and then discards that knowledge. parse(), parseOne() and allParsator() never raise to a caller, and the 4 except pairs they share log the exception and drop it. kering.py defines something like 81 error classes, and the distinctions are real: ShortageError means more bytes may help and ColdStartError means they won't; MissingWitnessSignatureError and LikelyDuplicitousError describe very different states of the world. All of it is available at the moment of failure and gone one frame later.

The parser has already made that coarser judgement. Each of those pairs splits into `except (ColdStartError, ExtractionError)`, which flushes the stream, and `except (ValidationError, Exception)`, which resumes — recoverable against not. That decision doesn't leave the frame either.

I'm building a library that consumes keripy goodness (lots of background that I'll skip), and I need the distinction for two reasons. When a stream fails to establish what it should, I have to decide how much of that failure to impute to which party's reputation, and I have to be able to say what the flaw in the evidence actually was. Right now a truncated stream, a desynced stream, an under-witnessed log, a duplicitous log, and a stream about the wrong identifier all look the same from outside: each establishes less than it should, and parse returns normally in every case. Working out which one happened means reading database residue and guessing at control flow that has already run.

So this doesn't change what parse does. It gives it somewhere to put what it currently only logs. There's an optional faults= argument on the parse entry points, threaded down through the generators, defaulting to a new Parser.faults which itself defaults to None. When nothing is supplied, nothing is collected and no code path changes. Parser is tolerant by design and should stay that way: it's a streaming parser, a live stream legitimately runs short and has to wait for more bytes rather than fail, and anything that made parse raise by default would break every existing consumer.

Each recovered error appends a frozen Fault. It carries a kind, which is the class name of the most specific KeriError in the chain; a disp, which is flush, resume, or group, the disposition the call site had already chosen; an offset, which is a lower bound on how many bytes the call consumed before the fault, and exact for a fixed stream; pre, sn, and said when the message had already been extracted; and the exception itself, with its chain intact and its traceback cleared.

kind walks the chain rather than naming the class the handler caught, because the parser rewraps as it unwinds. A truncated stream arrives at allParsator as an ExtractionError() with no message at all, and only the chain still says ShortageError. The walk follows `__cause__` where the raise was explicit and `__context__` otherwise, since all 4 SizedGroupError raise sites are bare raises inside an except block.

I got this next part wrong in my own first draft, so it's stated in the Fault docstring as plainly as I could manage. A Fault records what the parse loop did, not what became of the message. Kevery escrows an event and then raises, so OutOfOrderError, MissingWitnessSignatureError, MissingDelegationError, and LikelyDuplicitousError all produce a Fault for an event that a later processEscrows() may accept in full. Reading a Fault as "this message was rejected" turns ordinary out-of-order KEL traffic into a hard failure. A Fault isn't duplicity evidence either, since the conflicting event is already preserved in the duplicitous escrow, and "Likely" in LikelyDuplicitousError is there for a reason. This seems to me like the part most worth pushing on if it looks wrong.

disp got a namedtuple codex because it has three values. kind didn't, because pinning a normative name for each of 81 exception classes seems like your call rather than mine. As it stands, kind is documented as keripy's internal taxonomy: not stable across releases, not shared with signify-ts or KERIA, and an unrecognized value to be treated as a fault rather than as no fault. If you'd rather it were normative and cross-implementation, I can do that instead.

*Two things I ran into while reading these paths look like bugs rather than gaps*, and I left both alone because fixing either changes behavior. At parsing.py:1082, self.mucodes.FixBodyGroup raises AttributeError under CESR v1, because MessageUniversalCodex_1_0 defines only NonNativeBodyGroup — native bodies being v2-only. A v1 stream that starts on a stray attachment counter therefore dies with an AttributeError rather than with the ColdStartError two branches below, which seems to be what was intended. And at parsing.py:864, `except (ExtractionError, Exception) as ex: raise ExtractionError from ex` swallows SizedGroupError, whose docstring says the group has already been stripped and so the stream must not be flushed. Rewrapped as a plain ExtractionError it reaches allParsator and flushes the rest of the stream anyway. I can open issues for both if that's useful.

I added a bunch of new tests in tests/core/test_parsing.py, over both CESR v1 and v2 streams, native CESR bodies included. No existing test changed, and all of them pass cleanly.

### DIFF
--- a/src/keri/core/parsing.py
+++ b/src/keri/core/parsing.py
@@ -16,7 +16,7 @@ from ..kering import (Colds, sniff, Vrsn_2_0, Ilks,
                       UnexpectedCountCodeError, MaterialError, ValidationError,
                       QueryNotFoundError, ExtractionError, ShortageError,
                       ColdStartError, InvalidVersionError,
-                      SizedGroupError, TopLevelStreamError)
+                      SizedGroupError, TopLevelStreamError, KeriError)
 
 from .coring import (Seqner, Cigar, Diger, Noncer, Labeler, Number, Verser,
                      Dater, Verfer, Prefixer, Saider, Texter)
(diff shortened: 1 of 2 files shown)


SmithSamuelM:
@dhh1128 

The core parsing methods in Parser are .groupParsator .msgParsator and .msgProcess.

For parsing not processing, .groupParsator calls .msgParsator. 
.msgParsator does not swallow its exceptions so any caller of .msgParsator will get the exact exception type raised

.groupParsator shifts the exception  when it calls .msgParsator. It shifts the exceptions to all be ExtractionError but
raise with the `from ex` which means the originating exception is still provided. 

See the python docs on "from" in a re-raise

This allows the other Parser methods that call group parsator to easily log and swallow exceptions

There is a practical reason for swallowing exceptions in a long running service that faces the open internet. The consumers of .groupParsator were designed to support that function. In a long running service that faces the open internet, best practice for "security" related protocols is to not inform the caller why a given message sent to the service failed, this minimizes the ability for an oracle to probe the service for vulnerabilities.

One of the probes is to cause a crash the forces a restart of the server and the time to restart is a strong signal that an uncaught exception was triggered. So the approach is to swallow and log all exceptions, not bubble them up.


It sounds like you want to use the parsing to pre-qualify messages you plan on sending. In that case, I suggest just creating your own wrapper method for .groupParsator that does not swallow the exceptions but exposes them back to the caller. 


Pushing that new wrapper as a new Parser method would be appreciated. Like parseOneBare or parseAllBare, which does not swallow exceptions.

SmithSamuelM:
```python

try:
    # Code that triggers chained exceptions
    pass
except Exception as e:
    current = e
    chain = []
    
    while current:
        chain.append(current)
        # Prioritize explicit cause, fallback to implicit context
        current = current.__cause__ or current.__context__
        
    print("Extracted chain:", chain)

```

PR #1583, opened 2026-08-06
fix: preserve group hab rmids across Habery reinitialization

m00sey:
## Summary

`Habery.loadHabs` constructs `GroupHab` (and `SignifyGroupHab`) passing only `smids=habord.smids` — the stored `rmids` are never passed. `GroupHab.__init__` defaults `rmids` to `smids` when not provided:

```python
self.rmids = rmids or smids
```

so **every Habery reinitialization silently collapses a group's rotation-member roster to its current signing members**, even though the `HabitatRecord` on disk is correct.

## Impact

For any group whose rotation authority is spread wider than its current signing set (e.g. 3 current signers with a 7-member next-key commitment, `nt` weighted across 7), every flow that reads `ghab.rmids` after a restart operates on the wrong member set:

- multisig exn recipient lists gossip proposals only to the collapsed set;
- rotation events built from `ghab.smids`/`ghab.rmids` re-commit a shrunken next-key membership (`n` drops from 7 digests to 3), permanently ejecting the other rotation members if the event completes;
- delegation-approval anchoring rotations (EstOnly delegators) inherit the same collapsed set.

Observed in production: a Root multisig delegator with `k`=3/`n`=7 rendered and proposed with only 3 rotation members after every application start, on every member's machine, while the stored record held all 7 rmids the whole time.

## Fix

Pass `rmids=habord.rmids` through in both `loadHabs` branches (`GroupHab` and `SignifyGroupHab`). A stored `None` behaves exactly as before (defaults to smids).

## Testing

New regression test `test_habery_reinitialization_group_rmids`: builds a group with `rmids ⊃ smids` in a non-temp Habery, reopens it, and asserts the roster survives (fails on the base branch, passes with the fix). `tests/app/test_habbing.py` and `tests/app/test_grouping.py` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/keri/app/habbing.py
+++ b/src/keri/app/habbing.py
@@ -336,14 +336,16 @@ class Habery:
             if habord.mid and not habord.sid:
                 hab = GroupHab(ks=self.ks, db=self.db, cf=self.cf, mgr=self.mgr,
                                rtr=self.rtr, rvy=self.rvy, kvy=self.kvy, psr=self.psr,
-                               name=habord.name, pre=pre, temp=self.temp, smids=habord.smids)
+                               name=habord.name, pre=pre, temp=self.temp,
+                               smids=habord.smids, rmids=habord.rmids)
                 groups.append(habord)
             elif habord.sid and not habord.mid:
                 hab = SignifyHab(ks=self.ks, db=self.db, cf=self.cf, mgr=self.mgr,
                                  rtr=self.rtr, rvy=self.rvy, kvy=self.kvy, psr=self.psr,
                                  name=habord.name, pre=habord.sid)
             elif habord.sid and habord.mid:
-                hab = SignifyGroupHab(smids=habord.smids, ks=self.ks, db=self.db, cf=self.cf, mgr=self.mgr,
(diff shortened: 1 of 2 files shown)


kentbull:
This bug still exists for both GroupHab and SignifyGroupHab.

Reproducing this with one signer and three rotation-only members, afte rreopen, shows `ghab.rmids` to contain only the one signer even though the stored rotation member roster and KEL next-key commitments remain correct.

There are two related issues:
- Both constructors use `rmids or smids` causing an explicit `rmids=[]` to become the signing roster and is persisted incorrectly. Forwarding `rmids` to `loadHabs` does not fix this since the default must distinguish `None` from `[]`
- KLI multisig rotation defaults omitted `--rmids` to the signing roster independent of restart. This PR does not address that behavior.

## Impact

The reload bug can exclude rotation-only members from notifications. And after restarting, the KLI or KERIA might incorrectly remember 3 rotation members instead of 7. But the existing signed event still contains the next-key commitments for all 7. This impacts the KLI more than KERIA since the KERIA external API endpoints report the correct member lists which Signify then makes edge events based on.

Issue #1569, opened 2026-08-01
Kevery.escrowLDEvent calls removed Baser.addLde — likely-duplicitous escrow (ldes) unreachable, LikelyDuplicitousError never raised


Nicholas-Keystate:
## Summary

Both write paths to the likely-duplicitous escrow call `self.db.addLde(...)`, a method that was removed from `Baser` in PR #1174. Any event that should be classified likely-duplicitous now raises `AttributeError` instead of `LikelyDuplicitousError`: nothing is ever written to `.ldes`, `processEscrowDuplicitous` never has anything to process, and superseding recovery loses the escrow-retry path that would legitimize a held event once a recovery window opens. When the event arrives through `Parser`, the catch-all converts the `AttributeError` into a misleading `"No kevery to process so dropped msg"` diagnostic, so the failure is invisible in normal operation.

No test currently exercises the KEL-layer likely-duplicitous path (the only `LikelyDuplicitousError` assertions are in `tests/vdr` against Tevery), which is how this has survived on `main` since February.

## Reproduction (current `main`, also `2.0.0-dev6`)

```python
from keri.core import eventing, signing
from keri.db import basing
from keri.kering import LikelyDuplicitousError, Kinds

signer = signing.Signer(transferable=True)  # Ed25519, basic 'D' prefix

icp0 = eventing.incept(keys=[signer.verfer.qb64], kind=Kinds.json)
icp1 = eventing.incept(keys=[signer.verfer.qb64], kind=Kinds.json,
                       data=[{"i": "conflict"}])
assert icp0.pre == icp1.pre          # same AID (basic derivation from k[0])
assert icp0.said != icp1.said        # different SAID -> duplicitous icp

with basing.openDB(name="repro") as db:
    kvy = eventing.Kevery(db=db, lax=True, local=False)

    kvy.processEvent(serder=icp0, sigers=[signer.sign(ser=icp0.raw, index=0)])
    print("first icp accepted:", icp0.pre in kvy.kevers)

    try:
        kvy.processEvent(serder=icp1, sigers=[signer.sign(ser=icp1.raw, index=0)])
    except LikelyDuplicitousError as ex:
        print("EXPECTED: LikelyDuplicitousError:", ex)
    except AttributeError as ex:
        print("ACTUAL: AttributeError:", ex)
        print("ldes entries:", [x for x in db.ldes.getItemIter(keys=(icp0.preb,))])
```

Output:

```
first icp accepted: True
[...]

PR #1587, opened 2026-08-06
fix(eventing): escrow duplicitous events in .ldes, not through a removed Baser method

### LINKED ISSUE #1586: Kevery.escrowLDEvent calls a removed Baser method, so duplicitous events are silently dropped
`Kevery.escrowLDEvent` calls `self.db.addLde(...)`, a Baser method that no longer exists, so every likely-duplicitous event is discarded instead of escrowed.

## Where

`src/keri/core/eventing.py`, in `escrowLDEvent`:

```python
self.db.dtss.put(keys=dgkey, val=Dater())
self.db.sigs.put(keys=dgkey, vals=sigers)
self.db.evts.put(keys=(serder.preb, serder.saidb), val=serder)
self.db.addLde(snKey(serder.preb, serder.sn), serder.saidb)   # <-- AttributeError
```

`Baser`'s escrows moved from the `addXxx(key, val)` API onto `subing` instances, and the sibling escrows moved with them — `escrowOOEvent` writes `self.db.ooes.add(keys=serder.preb, on=serder.sn, val=serder.saidb)` and `escrowPWEvent` writes `self.db.pwes.add(...)`. This call site was left behind. `Baser` defines no `addLde`.

## What happens

`escrowLDEvent` raises `AttributeError` from inside `Kevery.processEvent`. `Parser.msgParsator` catches `AttributeError` around the `kvy.processEvent(**exts)` call and re-raises it as

```
ValidationError: No kevery to process so dropped msg=<said>
```

which is a plausible message for an unrelated condition, so the symptom reads as a routing problem rather than a missing method. The event is dropped. `.ldes` is never written, `.evts` and `.sigs` never receive the conflicting event's body or signatures, and `processEscrowDuplicitous` has nothing to iterate.

## Why it matters

The material that duplicity is proven with is silently lost. Feed a `Kevery` two validly signed conflicting events at one sequence number — two `ixn`s chaining from the same prior, say — and:

- the KEL keeps whichever arrived first;
- the other leaves no trace anywhere in the database;
- which version wins is decided by arrival order;
- a verifier that reads `.ldes` to detect duplicity sees an empty escrow for a duplicitous identifier.

That last point is the one that bites. The `Fault` type added for parser diagnostics is explicit that it "is NOT evidence… the conflicting event is already preserved in the duplicitous escrow", so `.ldes` is the documented place to look, and it is empty. An embedder doing local replay to decide whether an identifier is duplicitous gets a clean answer for a forked log.

## Reproducing

```python
psr.parse(ims=signed(icp), kvy=kvy)
this = interact(pre=pre, dig=icpdig, sn=1, data=[dict(fork="this")])
that = interact(pre=pre, dig=icpdig, sn=1, data=[dict(fork="that")])
psr.parse(ims=signed(this), kvy=kvy)
psr.parse(ims=signed(that), kvy=kvy)

assert kvy.kevers[pre].serder.said == this.said   # passes
assert db.ldes.get(keys=pre, on=1) == [that.said] # fails: []
```

## Fix

One line, mirroring the sibling escrows:

```python
self.db.ldes.add(keys=serder.preb, on=serder.sn, val=serder.saidb)
```

There was no test over the KEL duplicity path, which is how a call to a removed method survived — `.ldes` is referenced in `tests/db/test_basing.py` and in the TEL tests, never in `tests/core/test_escrow.py`. PR attached adds `test_likely_duplicitous_escrow` alongside the fix, covering the escrow entry, the preserved body and signatures, the unchanged KEL, escrow survival across a `processEscrowDuplicitous` pass, and a plain duplicate adding no entry.

Full suite passes with the fix (667 passed, 3 skipped).

dhh1128:
Fixes #1586.

I found this issue while working in some code that consumes keripy. I think it's just caused by a partial edit.

`Kevery.escrowLDEvent` ended with `self.db.addLde(snKey(...), serder.saidb)`, an `IoDupSuber`-era `Baser` method that no longer exists. `Baser`'s escrows moved onto `subing` instances and the sibling escrows moved with them — `escrowOOEvent` writes `self.db.ooes.add(keys=..., on=..., val=...)`, `escrowPWEvent` writes `self.db.pwes.add(...)` — but this call site was left behind.

So every likely-duplicitous event raised `AttributeError` from inside `processEvent`, which `Parser.msgParsator` catches and re-raises as `ValidationError: No kevery to process so dropped msg=…`. The event was discarded: `.ldes` never written, `.evts` and `.sigs` never given the body or its signatures, `processEscrowDuplicitous` left with nothing to iterate.

The consequence is a silent loss of exactly the material duplicity is proven with. Two validly signed conflicting events at one sn leave no trace — the KEL keeps whichever arrived first, arrival order decides which, and a verifier reading `.ldes` sees an empty escrow for a duplicitous identifier. Since `Fault` is explicitly not evidence and points at the duplicitous escrow instead, `.ldes` is the documented place to look.

The fix is one line, mirroring the sibling escrows:

```python
self.db.ldes.add(keys=serder.preb, on=serder.sn, val=serder.saidb)
```

There was no test over the KEL duplicity path, which is how a call to a removed method survived. `test_likely_duplicitous_escrow` covers it: two `ixn` events at one sn, neither superseding the other, asserting the escrow entry, the preserved body and signatures, the unchanged KEL, escrow survival across a `processEscrowDuplicitous` pass, and a plain duplicate adding no entry.

### DIFF
--- a/src/keri/core/eventing.py
+++ b/src/keri/core/eventing.py
@@ -5811,7 +5811,7 @@ class Kevery:
         self.db.dtss.put(keys=dgkey, val=Dater())
         self.db.sigs.put(keys=dgkey, vals=sigers)
         self.db.evts.put(keys=(serder.preb, serder.saidb), val=serder)
-        self.db.addLde(snKey(serder.preb, serder.sn), serder.saidb)
+        self.db.ldes.add(keys=serder.preb, on=serder.sn, val=serder.saidb)
         # log duplicitous
         logger.debug("Kevery process: escrowed likely duplicitous event=\n%s\n", serder.pretty())
 
--- a/tests/core/test_escrow.py
+++ b/tests/core/test_escrow.py
@@ -2050,6 +2050,91 @@ def test_unverified_trans_receipt_escrow():
     """End Test"""
 
 
+def test_likely_duplicitous_escrow():
+    """
+    Test likely duplicitous escrow
+
+    Two validly signed events at one sn that neither supersedes the other must
+    be preserved rather than dropped: duplicity is proven by placing the
+    conflicting events side by side, so the second one is escrowed in .ldes
+    with its body in .evts and its signatures in .sigs, while the KEL keeps
+    the version it accepted first.
+    """
+    salt = Salter(raw=b'0123456789abcdef').qb64  # init wes Salter
+    psr = parsing.Parser(version=Vrsn_1_0)
+
+    # init event DB and keep DB
+    with openDB(name="edy", temp=True) as db, keeping.openKS(name="edy") as ks:
+        # Init key pair manager
+        mgr = keeping.Manager(ks=ks, salt=salt)
+
+        # Init Kevery with event DB
+        kvy = Kevery(db=db)
+
+        verfers, digers = mgr.incept(icount=1, ncount=1, stem='wes', temp=True)
+
+        srdr = incept(keys=[verfer.qb64 for verfer in verfers],
+                      ndigs=[diger.qb64 for diger in digers],
+                      code=MtrDex.Blake3_256, version=Vrsn_1_0, kind=Kinds.json)
+
+        pre = srdr.ked["i"]
+        icpdig = srdr.said
+
+        mgr.move(old=verfers[0].qb64, new=pre)  # move key pair label to prefix
+
+        def signed(srdr):
+            """Serialize srdr with its controller signatures attached"""
+            sigers = mgr.sign(ser=srdr.raw, verfers=verfers)
+            msg = bytearray(srdr.raw)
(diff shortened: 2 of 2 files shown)


Nicholas-Keystate:
Same defect as #1569 (filed Aug 1, with repro against `main` and `2.0.0-dev6`) — this PR would close both. Confirming independently: the fix form here is the one we verified at both the current head and 79e31cc8, and a conformance battery that exercised the likely-duplicitous path went from AttributeError to fully green under exactly this change. One note for review context: this wasn't a simple partial edit — #1174 did convert this call site, but #1181 (branched before #1174 merged) restored the stale line at merge. Worth a check that no other Baser-era call sites rode back in the same way (`git grep 'db.add'` comes back clean, for what it's worth).

PR #1612, opened 2026-08-11
ACDC Blindable State Registry

KeaxD:
- Added a new V2 registry service in `keri.acdc`:
  - `RegistryStore`
  - `Regery`
  - `Registry`
  - `Registrar`
- Added support for V2 registry events:
  - `rip` for registry inception
  - `bup` for blindable state updates
- Implemented registry event processing behavior for:
  - KEL anchor verification
  - accepted TEL/head tracking
  - missing-anchor escrow
  - out-of-order escrow
  - replay/processing of queued escrows
  - persisted registry alias reload on reopen
- Exposed the new registry helpers from `keri.acdc.__init__`
- Added focused tests for the new registry service, including:
  - happy-path `rip -> bup`
  - malformed and conflicting event rejection
  - explicit vs discovered anchor handling
  - escrow replay behavior
  - persisted registry reload behavior
- Expanded V2 IPEX coverage for registry-backed disclosures:
  - builder/parse happy paths
  - dispatch and notification flow
  - clear registry update roundtrip
  - blind registry update roundtrip
  - nested artifact mismatch / omission rejection cases


SmithSamuelM:
@KeaxD 
We are deprecating the 'upd' event so no reason to implement it. Only do the 'bup'

so the happy path should be Added focused tests for the new registry service, including:
happy-path rip  -> bup -> bup etc
So revise  We are going to drop support for `upd` in v1.1

dhh1128:
I may be missing an intended re-anchoring behavior here, so check me.

Since putAnchor() (in registring.py) uses pin, calling anchorMsg() again for an already anchored queued event replaces its original anchor, I think. I get this from:

>put(...)  # Does not overwrite
>pin(...)  # Overwrites

Therefore, these two calls do different things:

>ancs.put(keys=event_said, val=first_anchor)
>ancs.put(keys=event_said, val=later_anchor)

The second call fails or leaves the first value intact. But with pin():

>ancs.pin(keys=event_said, val=first_anchor)
>ancs.pin(keys=event_said, val=later_anchor)

...the stored value becomes later_anchor. Since putAnchor() calls pin(), it has the second behavior.

Given this, and adding the fact that _processQueued() sorts siblings using the replacement anchor, an event that was actually anchored first can get lost after being sealed again.

Would it make sense to preserve the first verified anchor—perhaps using insert-only semantics—and add a regression test covering A anchored, B anchored, A re-anchored, then predecessor committed?

To be clear, this is not strictly a concern about the PR. It's just noticing that the PR touches some code that doesn't protect against this kind of misbehavior. As a result, the code is fragile and might need an update.

SmithSamuelM:
At a high level, anchoring TEL events should follow the idempotent first seen rule used for KEL events.  So put not pin. And if a  new anchor is different, it does not clobber a previously verified anchor

PR #1619, opened 2026-08-15
Add some more TEL verification (residuals after 1612 merged)

dhh1128:
This PR is partly redundant with #1612, which just merged. I am in the process of stripping out that redundancy, leaving behind just the parts that still have residual value. That's why it's still in draft. I'll post a description when I'm done.


dhh1128:
Closed in favor of 1631

PR #1620, opened 2026-08-15
feat: upgrade to Hio 0.7.19 and Python 3.14.0

kentbull:
It looks like upgrading to Hio is a small lift for KERIpy 1.2.14, from what I can tell.


kentbull:
Closing as we should not force KERIpy consumers to upgrade to Python 3.14 just yet.

Issue #1628, opened 2026-08-19
new question about spurned grants


dhh1128:
In #1517, grant was removed from the list of routes a spurn can answer, and the reasoning quoted there from @SmithSamuelM is that the agreement is binding: once the grant is sent the disclosure cannot be pulled back, so letting the disclosee change their mind afterward would sanction a race the discloser can't win. I agreed at the time, and I still think that's right for the exchange it describes. Where an apply, an offer, and an agree came first, the disclosee has already committed to the terms, and a spurn arriving after the data would be repudiation after performance.

The trouble is that a grant doesn't always follow an agree. `src/keri/acdc/ipexing.py:279` lets a grant open a flow with an empty prior, and that path is the unsolicited disclosure — somebody pushes a credential at me that I never applied for. There was no agreement in that case, so there is nothing binding to repudiate, and the reasoning doesn't apply. The test that @KeaxD built to pin the current rule uses exactly this case: the grant it builds at `tests/acdc/test_ipexing.py:413` is labeled "Bare grant without agreement", and the spurn against it is asserted to be rejected. So the party with the fewest obligations in the whole exchange, someone who asked for nothing, is the one left with no way to say no. Silence is all that remains, and silence is also what a disclosee produces when they never received the message at all. A spurn is evidence of receipt besides, since you can't spurn something you didn't get; refusing to accept one doesn't keep the discloser from having disclosed, it just removes the record that the data arrived.

So I'd wondering if we need a narrower version of the rule. A grant is spurnable when its own `p` is empty, and not spurnable when it closes an apply/offer/agree chain. That's a conditional in `verify()` rather than an entry in `PreviousRoutes`, and I think it's about three lines. Whichever way this goes, the docstrings and probably the spec should articulate the rule clearly, since as @ryan-hansen pointed out, v1 still accepts a spurn of a grant, which was raised on the PR at the time and is still true at HEAD; an implementer reading only the v1 side would get this wrong. I may be missing a reason the bare-grant case should be treated like the agreed one, so I'd welcome Sam's thinking here.

SmithSamuelM:
I agree. an unsolicited grant i.e using a xip message should be spurnable.  that is nuance i did not anticipate. If KRAM supports a bare exn transaction (no xip but with empty p field then that as well)

KeaxD:
I'll have a PR for that

PR #1629, opened 2026-08-20
Refactor V2 IPEX message shapes and data validation

KeaxD:
This PR fixes the v2 IPEX message shape, tightens message verifier, and aligns KRAM handling with flow-starting IPEX messages.

- Refactored `src/keri/acdc/ipexing.py` to generate `xid` internally for flow starters (apply, grant, offer) for KRAM
- Tightened verifier checks so replies:
  - must target the prior sender,
  - must come from the prior receiver,
  - must inherit the prior `xid`.

 Those changes are from the v2 ipex fields discussion: https://github.com/WebOfTrust/keripy/discussions/1595
- Changed `apply()` so:
  - the dead `schema` parameter is removed,
  - `attrs` is treated only as body payload,
  - disclosure paths must come from `modifiers["dp"]`.
- Changed `offer()` so:
  - it no longer carries a nested ACDC artifact,
  - it can open a flow directly when `recp` is provided,
  - it defaults `q.dp` to an empty list.
- Changed `grant()` so:
  - it now uses `origin` plus `artifacts`,
  - `a.o` identifies the first required nested artifact,
  - additional carried artifacts are transported only as ordered nested substreams.

- Updated `spurn(grant)` semantics so only a flow-starting grant may be spurned. see (https://github.com/WebOfTrust/keripy/issues/1628)

Notable changes:
- `apply()` no longer accepts `schema`
- `offer()` no longer accepts `acdc`
- `grant()` now takes `origin` / `artifacts` instead of `acdc` / `iss` / `anc`

It also adds some tests for KRAM, IPEX and registries functionality


SmithSamuelM (commented):
The offer message also needs to have an origin `o` field to replace the ACDC field. This is so the offer can provide a DAG of metadata ACDCs to better define the shape of what is to be presented in the grant.

src/keri/acdc/ipexing.py
```diff
@@ -385,48 +400,74 @@ def apply(hab, recp, message, schema, attrs, dt=None, kind=None, gvrsn=None):
...
+            raise ValueError("recp does not match prior exchange sender")
+        receiver = apply.ked["i"]
+        pxid = apply.ked.get("x", "")
+        if pxid:
+            xid = pxid
+        else:
+            xid = ""
+    data = dict(attrs) if attrs is not None else {}
+    data["m"] = message
+    data["o"] = _streamSerder(origin).said
```
ryan-hansen, line 468: Sam mentioned that offer o replaces the old ACDC field so the offer can carry a DAG of metadata ACDCs. Grant o is the real presentation origin (different object).

Right now this only writes a.o as a SAID and the verifier rejects offer nests, so the origin body never goes on the wire. The tests also pass the same issued ACDC used later in the grant, which makes offer o and grant o the same SAID.

Either attach that metadata DAG (nest 0 bound to a.o, like grant) or confirm with Sam that SAID-only is enough for this slice. As written, o is required but does not do the protocol job he described.

SmithSamuelM: The offer `o` is just the SAID of the Metadata ACDC DAG origin.  The actual metadata ACDCs are attached in a substream. So structurally the offer looks a lot like the eventual  grant when the offer is using a metadata acdc to convey least disclosure. For example, the metatdata ACDC in the offer disloses the rule section but not the attribute section.  This enables the agreent  to agree to the rules. The schema in the 'dp' field doesn't have the rule detail, so the only way for the agreent to know what it is agreeing to is for it to see the metadata ACDC DAG attached to the offer.

PR #1631, opened 2026-08-21
Add the party-side verification core for v2 blindable state registries

dhh1128:
#1612 gave keripy the issuer side of v2 blindable state registries. This is the other half: the verification a party performs on someone else's registry. Specifically, this fills the `keri/acdc/registring.py` stub (now renamed to regeventing.py, BTW) with another party's registry evidence and decides what it establishes. `vet()` takes a registry event chain, a `Baser` the caller has already populated with the issuer's KEL, and optionally a disclosed blinded-state block, and returns a `RegStateRecord` or raises a named refusal. It fetches nothing, opens nothing, and needs no local `Hab`.

This is what it checks, beyond the field and SAID validation `SerderACDC` already does:

- Chain rules: `n` starts at 0 and increments by exactly one, `p` chains by SAID, `rd` names the `rip`. I believe a gap is malformed in this case, so it is a refusal rather than an escrow candidate.
- KEL anchoring per event, matched by seal digest. A bare digest seal, a seal mapping with an `s` hint, and one without all anchor; several seals per key event are tolerated; the `s` is a lookup hint but not more. The anchoring KEL's controller must be the `rip`'s `i`.
- The binding equalities, as first-class named checks: the head state's `td` equals the presented ACDC's SAID, and the ACDC's `rd`, when it has one, equals the `rip`'s SAID. v2 ACDCs are not directly signed, so these two are the issuer's only commitment to the credential. A verifier that skips them discharges every other obligation and still accepts a substituted one, so there's a test in the PR that proves we reject a revoked ACDC riding a sibling registry's genuine issued chain, same issuer, where nothing else is wrong.
- Latest state at the highest `n` that chain- and anchor-verifies, with two distinct anchored events at one `n` refused as duplicity in either evidence order.
- BLID reproduction for a disclosed `bup` state. There is no salt parameter anywhere in the module — only the per-event disclosure is accepted, which is the independent verifier's position rather than the holder's. `Blinder.unblind` already serves the holder who has the uuid; nothing wired that to a `bup`'s `b` field.

Every presented event must anchor-verify before any conclusion is drawn. A missing anchor makes the whole determination retryable (`MissingAnchorError`) rather than partially conclusive.

### DIFF
--- a/src/keri/acdc/regeventing.py
+++ b/src/keri/acdc/regeventing.py
@@ -1,6 +1,6 @@
 # -*- encoding: utf-8 -*-
 """
-keri.acdc.registring module
+keri.acdc.regeventing module
 
 Registry (TEL) event support
 
--- a/src/keri/acdc/registring.py
+++ b/src/keri/acdc/registring.py
@@ -4,5 +4,447 @@ keri.acdc.registring module
 
 Registry (TEL) event support
 
+Party-side verification core for ACDC v2 blindable state registries (TELs)
+made of 'rip' (registry inception) and 'bup' (blindable update) events.  Given
+a supplied registry event chain, the issuer's KEL evidence, and optionally a
+disclosed blinded-state block, vet() and the vet* helpers determine the
+registry's verified state.  The core fetches nothing and concludes only from
+the evidence handed to it; KEL verification remains the caller's job.
 
+This complements keri.acdc.registraring, which manages the registries a local
+habitat controls.  Where that module issues, this one judges: its evidence is
+another party's, its anchor search runs against that party's KEL, and its
+refusals are named so a caller can tell a permanent one from a retryable one.
 """
+
+from collections.abc import Mapping
+from collections import namedtuple
+
+from .. import help
+from ..kering import (Ilks, ValidationError, InvalidValueError,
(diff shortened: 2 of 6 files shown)


dhh1128:
I am toggling this to draft status because I need to go over the code one more time. PR #1619 was where some of this code used to live, but I cut a bunch of it away (it was redundant with #1612) and reorganized a bit, and I did the slice and dice with AI, so I need to go back through and make sure I'm fully behind its new shape.

dhh1128:
@KeaxD I've added the commit to this PR that renames registring.py to regeventing.py

dhh1128:
I think this PR is now ready for review. Changing from Draft status to Ready.

KeaxD:
After looking through, It looks like we have similar functions between this PR and #1630 but instead of redundancy, I think they complement each other in a way where this PR focuses on verifier-side while #1630 is the issuer-side. We could have data validation helpers that could be shared from regeventing.py to registraring.py for issuer-side verification. 

So we would have this split:
regeventing = TEL validation core + verifier validation 
registraring = registry orchestration + issuer validation (using validation helpers from regeventing)

I'm not requesting any changes just putting out how we could merge those two PRs.

KeaxD (approved):

PR #1632, opened 2026-08-24
More v1 to v2 tests

ryan-hansen:


tests/vdr/test_verifying.py
```diff
@@ -1156,7 +1139,7 @@ def nearWithOp(op, claim):
         assert verfer.reger.saved.get(keys=near.saidb) is not None
 
         # A tuple is accepted as a sequence too, not just a list.
-        near = nearWithOp(("NI2I",), "tuple form")
+        near = nearWithOp(("NI2I"), "tuple form")
```
evanja57: ("NI2I") without the trailing comma is just a string; we should restore the comma so that this tuple stays a tuple

ryan-hansen: Good catch.  Fixed.

evanja57 (approved):

PR #1637, opened 2026-08-27
Split local TEL eventing into regeventing.py

KeaxD:
This PR moves the local TEL event engine out of `registraring.py` and into `regeventing.py`.

The intended architecture:
- `registraring.py` coordinates registries
- `regeventing.py` owns TEL event logic
- shared TEL rules now live in one place instead of being split across issuer-side and verifier-side implementations

`regeventing.py`
- Added a private issuer-side engine: `_RegEventer`
- Moved the local TEL workflow into `_RegEventer`, including:
  - `processEvent`
  - `processEscrows`
  - `anchorMsg`
  - local missing-anchor replay
  - out-of-order replay
  - staged-frontier walking for `blind()`
  - local anchor-couple verification
- Added a normalized event record helper to avoid repeatedly re-parsing TEL event fields across local and verifier paths
- Consolidated shared TEL validation rules into shared helpers for:
  - `rip` validation
  - `bup` validation
  - prior-link checking
  - explicit local anchor-couple verification

`registraring.py`
- Reduced `Registry` to a thin public facade over `_RegEventer`
- Kept `Regery` responsible for creating/loading registries and processing escrows across managed registries
- Preserved the caller-facing `Registry` and `Registrar` API while moving TEL state-machine behavior out of the registrar layer

### DIFF
--- a/src/keri/acdc/regeventing.py
+++ b/src/keri/acdc/regeventing.py
@@ -4,29 +4,37 @@ keri.acdc.regeventing module
 
 Registry (TEL) event support
 
-Party-side verification core for ACDC v2 blindable state registries (TELs)
-made of 'rip' (registry inception) and 'bup' (blindable update) events.  Given
-a supplied registry event chain, the issuer's KEL evidence, and optionally a
-disclosed blinded-state block, vet() and the vet* helpers determine the
-registry's verified state.  The core fetches nothing and concludes only from
-the evidence handed to it; KEL verification remains the caller's job.
+Shared TEL support for ACDC v2 blindable state registries made of 'rip'
+(registry inception) and 'bup' (blindable update) events.
 
-This complements keri.acdc.registraring, which manages the registries a local
-habitat controls.  Where that module issues, this one judges: its evidence is
-another party's, its anchor search runs against that party's KEL, and its
-refusals are named so a caller can tell a permanent one from a retryable one.
+This module now holds three closely related concerns:
+
+1. Verifier-side evidence vetting through vet() and the vet* helpers. Given a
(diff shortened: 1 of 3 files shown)


src/keri/acdc/regeventing.py
```diff
@@ -448,3 +634,571 @@ def vet(rip, updates=None, *, db, acdc=None, blinder=None, sources=None):
...
+        # Ask the local habitat KEL for the latest key event whose seal payload
+        # points at this TEL event.
+        aserder = self.hab.db.fetchLastSealingEventByEventSeal(
+            pre=self.hab.pre,
+            seal=dict(i=event.regk, s=event.serder.sad['n'], d=event.said),
+        )
+        if aserder is None:
+            return False
+
+        # Retrieve the sn and and dig from the sealing event and verify the anchor couple
```
ryan-hansen: double "and" in comment.   Not the end of the world, but might as well clean it up.

src/keri/acdc/registraring.py
```diff
@@ -169,8 +168,29 @@ def __init__(self, hab, store, name="test", regk=None):
         self.hab = hab
         self.store = store
         self.name = name
-        self.regk = regk
-        self.regd = None
+        # Registry remains the public facade, but the TEL event state machine
+        # now lives in the private regeventing engine.
+        self._eventer = regeventing._RegEventer(hab=hab,
```
ryan-hansen, line 173: If we want to be strict, we probably shouldn't use a "private" method here.  Personally, I'm not gonna lose any sleep over it, but figured I'd call it out.

PR #1638, opened 2026-08-27
Close direct connections without Reactants immediately

kentbull:
## Summary

- close a receive-cutoff Remoter immediately when it has no Reactant and no input
- log any residual transmit bytes as unexpected unowned output before cleanup
- reserve the absolute drain deadline for Reactant-backed parser, response, and transmit work

## Rationale

The only KERIpy producer for output on a Directant-owned Remoter is Reactant.sendMessage. Without a Reactant, waiting up to 30 seconds cannot advance application-owned output and instead assigns a speculative lifecycle to a state KERIpy does not produce.

This change closes that ownerless state immediately. It does not change the existing bounded drain for accepted CESR input, active response production, or legitimate Reactant-owned transmit bytes.

### DIFF
--- a/src/keri/app/directing.py
+++ b/src/keri/app/directing.py
@@ -387,8 +387,9 @@ class Directant(doing.DoDoer):
         terminal. The whole-drain deadline bounds parser, producer, and egress
         work and logs their remaining state before forced removal.
 
-        A receive-closed Remoter with no parser work still waits for any queued
-        ``txbs`` to drain, become terminal, or reach the same deadline.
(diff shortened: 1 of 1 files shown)


kentbull:
Cleans up a prior overly cautious no-Reactant logic branch that really is never hit in production code nor should it be. Whenever there is no Reactant then we are safe to tear down a Remoter a Directant has access to when a peer TCP EOF is received on that Remoter prior to it being assigned to a Reactant.

kentbull:
Reverting so we can get a maintainer review before we merge.

PR #1639, opened 2026-08-27
Reconcile Reactants after transport removal

kentbull:
## Summary

- reconcile Directant Reactants against the exact Remoters currently registered by HIO
- remove stale Reactant doers and drain deadlines before scheduling children
- preserve and admit a replacement Remoter that reuses the same connection address
- document that HIO owns transport connections while Directant owns their application lifecycle

## Rationale

HIO Server service may remove a Remoter from `server.ixes` after a receive error. That path bypasses `Directant.closeConnection`, leaving the corresponding Reactant scheduled and its drain deadline retained even though the socket is already gone.

Directant now performs a small identity-based sweep at the start of each `recur`, before `DoDoer` schedules children. Address membership alone is insufficient because a newly accepted connection may reuse the same address; the registered Remoter must be the exact object owned by the Reactant.

The sweep removes only stale application children. It does not close the current HIO connection, so a replacement Remoter remains active and receives a new Reactant during normal service.

## Tests

- `tests/app/test_directing.py`: 12 passed

The focused regression enters a Reactant-backed response drain, models the HIO ownership boundary through `Server.removeIx`, installs a replacement under the same address, and verifies stale-child cleanup plus replacement ownership. It does not duplicate HIO errno classification tests.

### DIFF
--- a/src/keri/app/directing.py
+++ b/src/keri/app/directing.py
@@ -307,11 +307,15 @@ class Reactor(doing.DoDoer):
 class Directant(doing.DoDoer):
     """Supervise inbound direct-mode TCP connections with one Reactant each.
 
-    Directant owns every accepted raw-TCP ``Remoter`` and its corresponding
-    ``Reactant``. HIO ``cutoff`` means only that receive is closed, so it is a
-    transition into draining rather than permission to discard accepted input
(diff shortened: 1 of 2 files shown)


kentbull:
Closing until upstream HIO changes land.

PR #1640, opened 2026-08-28
Preserve non-Ed25519 key algorithm across rotations

dhh1128:
An identifier incepted with a non-Ed25519 signing key changes algorithm at its first rotation, and nothing reports it:

```python
hab = hby.makeHab(name="p256", icode=MtrDex.ECDSA_256r1_Seed, wits=[], toad=0)
hab.kever.verfers[0].code   # '1AAJ'  -- ECDSA_256r1, as asked for
hab.rotate()
hab.kever.verfers[0].code   # 'D'     -- Ed25519
```

`Hab.incept` takes `icode` and passes it to `Manager.incept`, but doesn't pass `ncode`, so the pre-rotated next keys get generated with the `MtrDex.Ed25519_Seed` default no matter what the current keys were made with (habbing.py:2953). `Hab.rotate` has the same gap at its `Manager.rotate` call (habbing.py:3065). The manager itself has had `ncode` all along; it just never receives one.

I think this is worth fixing rather than documenting, because the failure is silent and impacts a relying party more than on the caller. Anything holding a certificate or a cached key for the old algorithm stops being able to verify, at a rotation nobody thought was a change of algorithm.

The fix defaults `ncode` to `icode` at inception, on the theory that an identifier signs with one algorithm and a next-key commitment on a different one is rarely what someone means. `Hab.rotate` takes `ncode` too, defaulting to the algorithm the identifier is already using. Recovering that turned out to need a small inverse map: the key store keeps public keys and seeds but not the code they were made with, so the current public key is the only record of the algorithm. `SeedCodeByVerferCode` in `core.signing` is that inverse, sitting next to the forward derivation in `Signer.__init__` so the two stay together. Both parameters stay explicit, so a caller who does want to change algorithm can pre-commit a different one, and the change then takes effect at the rotation the commitment already promised.

There's a second commit for something adjacent that I hit immediately afterward. `Encrypter.encrypt` picks its cipher code from a two-branch check naming `Salt_128` and `Ed25519_Seed`, with a `# future other seed codes` comment on the second (signing.py:806-808). So a P-256 identifier could be incepted into a plain key store but not an encrypted one, which surfaced as `Unsupported primitive with code = Q when cipher code is missing` from a passcode-protected store after the same inception had worked fine against a temp one. One cipher code covers every seed, since the sealed box is over the plain text's fully qualified qb64 and the seed's own derivation code travels inside the cipher. `X25519_Cipher_Seed` says how the wrapping was done rather than which algorithm was wrapped.

There are 10 tests, and 6 of the 9 in `test_habbing_v2.py` fail without the first commit. One in `test_signing.py` drives every seed code through encrypt and decrypt and checks the code survives the round trip. The mapping is kept complete by a test rather than by memory — a newly supported algorithm with no entry would fall back to Ed25519 silently, which is the failure the whole change is about.

Where this came from: Bakobo needs P-256 identifiers for EU digital identity work, where EdDSA isn't available to us. The ARF binds wallet providers, attestation providers, and relying parties to the ECCG Agreed Cryptographic Mechanisms v2.0, and that list has no EdDSA in it. I don't think any of that is keripy's problem, and the bug seems like a defect on its own terms — an `icode` parameter that quietly stops applying after one rotation is surprising regardless of which algorithm you asked for.

Branched off current main; `tests/app` and `tests/core` pass, 421 tests. If you'd rather review the two commits apart, I can split them into separate PRs.

### DIFF
--- a/src/keri/core/signing.py
+++ b/src/keri/core/signing.py
@@ -29,6 +29,25 @@ Tierage = namedtuple("Tierage", 'low med high')
 
 Tiers = Tierage(low='low', med='med', high='high')
 
+# Inverse of the seed-code -> verfer-code derivation Signer.__init__ performs below.
+#
+# Rotation has to carry an identifier's signing algorithm forward, and on an identifier that
+# already exists the only record of that algorithm is the code of its current public key: the
+# key store keeps public keys and seeds, not the code they were made with. Mapping back is
+# therefore how a rotation stays on the curve the inception chose, rather than silently
+# reverting to the Ed25519 default.
+#
+# Both the transferable and non-transferable public codes map to the same seed code, because
+# transferability is a property of how the public key is encoded and not of the algorithm.
+SeedCodeByVerferCode = {
+    MtrDex.Ed25519: MtrDex.Ed25519_Seed,
+    MtrDex.Ed25519N: MtrDex.Ed25519_Seed,
+    MtrDex.ECDSA_256r1: MtrDex.ECDSA_256r1_Seed,
+    MtrDex.ECDSA_256r1N: MtrDex.ECDSA_256r1_Seed,
+    MtrDex.ECDSA_256k1: MtrDex.ECDSA_256k1_Seed,
+    MtrDex.ECDSA_256k1N: MtrDex.ECDSA_256k1_Seed,
+}
+
 
 class Signer(Matter):
     """
--- a/src/keri/core/signing.py
+++ b/src/keri/core/signing.py
@@ -48,6 +48,10 @@ SeedCodeByVerferCode = {
     MtrDex.ECDSA_256k1N: MtrDex.ECDSA_256k1_Seed,
 }
 
+# Every private key seed Signer can be built from. Derived from the mapping above rather than
+# written out again, so the two cannot disagree about which algorithms exist.
+SeedCodes = frozenset(SeedCodeByVerferCode.values())
+
 
 class Signer(Matter):
     """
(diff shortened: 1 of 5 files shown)


src/keri/core/signing.py
```diff
@@ -29,6 +29,29 @@
...
+#
+# Rotation has to carry an identifier's signing algorithm forward, and on an identifier that
+# already exists the only record of that algorithm is the code of its current public key: the
+# key store keeps public keys and seeds, not the code they were made with. Mapping back is
+# therefore how a rotation stays on the curve the inception chose, rather than silently
+# reverting to the Ed25519 default.
+#
+# Both the transferable and non-transferable public codes map to the same seed code, because
+# transferability is a property of how the public key is encoded and not of the algorithm.
+SeedCodeByVerferCode = {
```
daidoji, line 42: Maybe Ed448 belongs here as well.

https://github.com/WebOfTrust/keripy/blob/f2faa2866f4dd8bf3cd9ba56ff4868375424670b/src/keri/core/coring.py#L245C6-L245C19

daidoji (changes requested):

src/keri/core/signing.py
dhh1128: This makes total sense. However, I poked around a bit, and Ed448 support isn't fully present yet. So if I add it here, I'm kind of pulling on a carpet thread and the PR gets a lot bigger. I am make a note to implement Ed448 more fully, including putting it here.

daidoji: ahh, got ya 👍

PR #1643, opened 2026-08-31
Add single-DAG IPEX nesting, origin walk, and node verification

KeaxD:
This PR updates V2 IPEX message handling to use the single-DAG wire shape discussed for v1.1, so disclosed ACDC nodes travel as per-node nested substreams and `grant` verification can walk the disclosed origin graph

It also wires in verifier-side registry issuer-auth checks for registry-backed disclosed nodes by validating the node-local proof group against locally available TEL/KEL evidence.

- `q.dp` round-trips as a list of disclosure-path entries, e.g. `[[schema, "/", []]]`
- `a.o` round-trips as a one-element origin SAID list
- `a.ax` is accepted and round-tripped as an optional list field on `apply`, `offer`, and `grant`
- single-DAG only is enforced for `o`; multi-DAG is for a later task

- `grant()` now treat `origin` + `artifacts` as disclosed ACDC DAG nodes
- each disclosed node is emitted as its own nested `BodyWithAttachmentGroup`
- the first nest is always the origin node named by `a.o[0]`
- issuer-auth proof material now lives on the node’s attachment section

- `IpexHandler.verify()` was refactored into staged helpers for readability
- `grant` verification now:
  - validates origin nest shape
  - builds a `said -> nest` map
  - BFS walks `e` edges from `a.o[0]`
  - rejects dangling edge references
  - rejects malformed or compact non-walkable edge shapes
  - rejects extra unreachable disclosed nests

- added `_verifyIssuerAuthNode()` to enforce node-local verifier checks for disclosed registry-backed ACDCs
- when a disclosed node has top-level `rd`, the handler now expects a node-local blind proof group and vets it with `regeventing.vet(...)`
- `loadHandlers(...)` now accepts optional `rgy` so IPEX handlers can reuse an injected registry manager

- updated `serializeParsedSubstream()` so parsed blind proof groups reserialize correctly when accepted IPEX messages are stored and replayed

### DIFF
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -393,15 +393,14 @@ class IpexHandler:
         elif not self._verifyReplyChain(verb=verb, serder=serder, dig=dig):
             return False
 
-        # Stage 4: when an offer carries a metadata DAG, or when a grant
-        # carries the final disclosure DAG, the nests must describe one exact
-        # reachable graph rooted at the message's `a.o[0]`.
+        # Stage 4: offer may disclose only a reachable metadata subgraph,
+        # while grant must disclose one fully closed reachable DAG rooted at
+        # the message's `a.o[0]`.
         if verb == Ipex.offer and nests:
-            if self._walkGraph(origin=attrs["o"][0], nests=nests) is None:
+            if self._walkGraph(origin=attrs["o"][0], nests=nests, closed=False) is None:
                 return False
         elif verb == Ipex.grant:
-            walked = self._walkGraph(origin=attrs["o"][0], nests=nests)
-            if walked is None:
+            if self._walkGraph(origin=attrs["o"][0], nests=nests, closed=True) is None:
                 return False
 
             # Stage 5: after the disclosed graph shape is accepted, each walked
@@ -486,17 +485,21 @@ class IpexHandler:
 
         return True
 
-    def _walkGraph(self, origin, nests):
+    def _walkGraph(self, origin, nests, *, closed):
         """Walk the disclosed origin DAG and return the visited node order.
 
         Parameters:
             origin (str): SAID of the origin node named by ``a.o[0]``.
             nests (list): Parsed nested ACDC node substreams carried by the
-                grant message.
+                offer or grant message.
+            closed (bool): True requires every referenced edge target to be
+                carried in ``nests``. False allows undisclosed far nodes, but
+                every carried nest must still be reachable from ``origin``.
 
         Returns:
             tuple | None: ``(nodes, order)`` when the disclosed nests form one
-                exact DAG rooted at ``origin``; otherwise ``None``.
+                reachable disclosed graph rooted at ``origin`` under the
+                selected closure rule; otherwise ``None``.
         """
         # Reuse the disclosed-node validation so graph walking starts from a
         # well-formed set of unique ACDC nests.
@@ -514,7 +517,9 @@ class IpexHandler:
         order = []
         queue = deque([origin])
 
-        # Walk the graph BFS, reject if any edge fails to resolve to a carried nest
+        # Walk the graph BFS. Grant fails closed on dangling references, while
+        # offer may omit farther undisclosed nodes as long as carried nodes
+        # still form one root-reachable subgraph.
         while queue:
             said = queue.popleft()
             if said in seen:
@@ -551,7 +556,9 @@ class IpexHandler:
                             except Exception:
                                 return None
                             if edgeSaid not in nodes:
-                                return None
+                                if closed:
+                                    return None
+                                continue
                             if edgeSaid not in seen:
                                 queue.append(edgeSaid)
                             continue
@@ -563,7 +570,8 @@ class IpexHandler:
                                 return None
                             groups.append(node)
 
-        # If any carried nest was never reached, the payload is not one exact DAG.
+        # Even when offer omits farther nodes, every carried nest must still be
+        # part of the root-reachable disclosed graph.
         if len(seen) != len(nodes):
             return None
 
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -443,8 +443,9 @@ class IpexHandler:
 
         Returns:
             bool: True when the reply points to an allowed prior message, keeps
-                sender/receiver roles consistent, and does not duplicate an
-                existing response; False otherwise.
+                sender/receiver roles consistent, does not duplicate an
+                existing response, and for ``grant`` keeps the disclosed origin
+                committed to the earlier ``offer``; False otherwise.
         """
         pserder, _ = cloneMessage(self.hby, said=dig)
         if pserder is None:
@@ -474,6 +475,23 @@ class IpexHandler:
 
         if self.response(pserder) is not None:
             return False
+
+        # A reply-chain grant must disclose the same origin DAG root the offer
+        # negotiated. Offer stays metadata-only, but its `a.o[0]` is still a
+        # commitment to the exact later grant root, not just a loose topic hint.
+        if verb == Ipex.grant:
+            offerDig = pserder.ked.get("p", "")
+            if not offerDig:
+                return False
+            offer, _ = cloneMessage(self.hby, said=offerDig)
+            if offer is None:
+                return False
+            oattrs = offer.ked.get("a")
+            if not isinstance(oattrs, dict) or not _validSingleDagList(oattrs.get("o"), str):
+                return False
+            if serder.ked["a"]["o"] != oattrs["o"]:
+                return False
+
         return True
 
     def _walkGraph(self, origin, nests):
@@ -782,7 +800,8 @@ def offer(hab, message, origin, artifacts=None, apply=None, recp=None, dt=None,
         message (str): Human-readable offer message.
         origin (Serder | bytes | bytearray): Origin disclosed ACDC node whose
             SAID is copied into ``a.o[0]``. ``offer`` does not disclose nested
-            node bodies; ``grant`` later carries the actual ACDC node nests.
+            node bodies, but this root SAID commits the negotiation to the
+            exact DAG that ``grant`` must later disclose.
         artifacts (list[Serder | bytes | bytearray] | None): Reserved for
             signature compatibility. ``offer`` does not disclose additional
             DAG nodes, so providing artifacts is rejected.
@@ -838,8 +857,8 @@ def offer(hab, message, origin, artifacts=None, apply=None, recp=None, dt=None,
     if "ax" in data and not _validSingleDagList(data["ax"], bool, allow_empty=True):
         raise ValueError("attrs['ax'] must be [] or a one-item list of booleans")
 
-    # Offer stays metadata-only: it names the disclosed DAG
-    # origin, but does not reveal the ACDC node bodies until grant.
+    # Offer stays metadata-only: it names the exact disclosed DAG origin up
+    # front, but does not reveal the ACDC node bodies until grant.
     originSerder = _streamSerder(origin)
     if originSerder.proto != Protocols.acdc or originSerder.ilk not in DisclosedNodeIlks:
         raise ValueError("offer origin must identify a disclosed ACDC node")
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -403,23 +403,43 @@ class IpexHandler:
             except Exception:
                 return False
 
-            # The first disclosed node is always the origin named in a.o[0].
-            # Every carried nest must be an ACDC body plus that node's
-            # attachment section.
-            for idx, nest in enumerate(nests):
-                nserder = nest["serder"] if isinstance(nest, dict) else nest.serder
-                if not nserder.verify():
-                    return False
-                if nserder.proto != Protocols.acdc:
-                    return False
-                if idx == 0 and not nserder.compare(attrs["o"][0]):
-                    return False
+            if self._validNodeNest(origin=attrs["o"][0], nests=nests) is None:
+                return False
 
         # The other verbs never disclose nested ACDC nodes.
         elif nests:
             return False
 
         return True
+
+    def _validNodeNest(self, origin, nests):
+        """Validate disclosed node nests and index them by ACDC SAID.
+
+        Parameters:
+            origin (str): SAID of the origin node named in ``a.o[0]``.
+            nests (list): Parsed nested ACDC node substreams.
+
+        Returns:
+            dict | None: Mapping of disclosed node SAID to its parsed nest when
+                every nest is a unique ACDC node and the first nest matches the
+                origin; otherwise None.
+        """
+        nodes = {}
+        for idx, nest in enumerate(nests):
+            nserder = nest["serder"] if isinstance(nest, dict) else nest.serder
+            if not nserder.verify():
+                return None
+            if nserder.proto != Protocols.acdc:
+                return None
+            if idx == 0 and not nserder.compare(origin):
+                return None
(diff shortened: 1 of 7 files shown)


dhh1128:
This looks really useful. It captures lots of what we've discussed about single-DAG enforcement, the one-element `o` list, etc.                                                                                                                                                                                                                                         I only have one main request for changes: `_walkGraph` assumes a fully expanded, fully disclosed, flat DAG, and treats every departure from that as an error. But that disallows graduated disclosure. Five spec-legal shapes get rejected by the code in this PR, with no diagnostic. One of them is already sitting in a merged test on `main`. Also, the registry check reads from a store that structurally can't hold another party's registry, so the main contribution of the PR isn't ready yet to work across a trust boundary. (The test suite doesn't catch either issue, because it was adjusted around the new verification rather than pointed at what we already had.)

A. More detail about _walkGraph                                                                                                                                                                                                              

`_walkGraph` skips the labels `d` and `o`, then wants every remaining value to be a Mapping carrying a string `n` (`ipexing.py:522-530`). I ran four shapes through the branch with every far node disclosed, and only the flat `{d, n, o}` block gets in:

```
simple-edge-format                     accepted=False
edge-section-with-u                    accepted=False
nested-edge-group-OR                   accepted=False
control-flat-edge-with-n               accepted=True
```

The spec's reserved edge labels are `[d, u, n, s, o, w]` (spec-body.md:1137), an Edge-group is defined by the absence of `n` (`:1069`), a simple compact edge's value is the far-node SAID rather than a block (`:4575`), and the Edge Section "MUST be expressable in the most compact form as merely its SAID" (`:1062`). There is already a concrete example: `upstream/main:tests/acdc/test_clc_disclosure.py:544` builds `ageEdge` with a section-level `u`, so the branch makes an ACDC that's already in the tree impossible to grant. #1560 has `EdgeGroupLabels = ('d', 'u', 'o', 'w')` and a recursive `walkEdgeSection` that would fix both. Consuming it would avoid a fifth edge walker in the codebase... But to be fair: it's the section- or group-level `u`/`w` that breaks, not an edge-level one, and a most-compact far *node* is accepted, so attribute-section graduated disclosure survives. The compact-edge rejections are defensible, since compact and simple-compact edges are indistinguishable without resolving the schema and this PR resolves none. What leaves me a bit uncomfortable is that the walk can't tell a compacted edge it doesn't have to follow from an expanded edge whose target is missing. 

B. More tests for the registry store?                                                                                                                                                                                                        

I think we need a new test, because the two-Habery tests don't seem to exercise things right. A disclosee can't verify a registry-backed grant from anybody else. `_verifyIssuerAuthNode` loads `rip` and the update chain out of `self.rgy.store` and nowhere else (`ipexing.py:647-655`), and nothing can put a foreign TEL in that store: `RegistryStore.accept` has one caller, `Registry._commit` (`registraring.py:305`); a `Registry` only comes from `Regery.makeRegistry(name, prefix)`, which requires `prefix` to be a local Hab (`registraring.py:67-69`); and `loadRegistries` re-checks that on load. So `rip` is `None`, verification returns `False`, and the grant is dropped. The two-Habery tests pass only because the recipient's handlers get the issuer's own `Regery`: `rgy = Regery(hby=issuerHby, ...)` then `loadHandlers(hby=recipientHby, ..., rgy=rgy)` at `test_ipexing.py:2370-2372`, with a comment admitting it. Giving each side its own store there is the test I'd ask for, and I bet it would fail. Underneath the test problem is a design gap: #1595 starts down the path of dropping `iss`/`anc`, but nothing replaces the transport, and `vet()` takes an `updates` argument because a party is expected to be handed the chain. ACDC #204 has the Discloser supply the `rip` and the applicable state update explicitly, and under independent-registry bulk issuance the registry is new on every presentation.
                                                                                                                                                                                                  
 It also looks to me like a node with no `rd` gets no issuer verification at all. `_verifyIssuerAuthNode` wraps its whole body in `if regk:` and otherwise returns `True`. The only other per-node check is `nserder.verify()`, which is SAID self-consistency (`serdering.py:668-684`). V2 ACDCs aren't directly signed — `regeventing.py:307-309` says a verifier that skips these checks "passes every other obligation while accepting a substituted credential" — and the outer exn signature authenticates the discloser, not the issuer. `ipexing.py:660` is the only caller of `regeventing.vet` in all of `src/`, and `exchanging.py:170-190` makes that boolean the sole gate on persistence and the notifier. So a peer can fabricate an ACDC body naming GLEIF as issuer, omit `rd`, nest it as origin, and the grant verifies. Combined with the paragraph above, the only nodes that can currently pass are the unverified ones.                                                                                                                                                                           

C. Some other minor details that might be worth debate                                                                                                                                                                                       

Requiring every edge target to be disclosed is applied to `offer`, which inverts chain-link confidentiality. `if edgeSaid not in nodes: return None` (`ipexing.py:535-536`) runs for `offer` as well as `grant` (`:361-366`), so the only acceptable offer has no edges or discloses the whole DAG. Sam's note on #1629 was that "the metadata ACDC in the offer discloses the rule section but not the attribute section," and #203's §Closures treats a far-side ACDC revealed by an edge as a separate disclosure with its own closure. Refusing unreachable *extra* nests is right and implements #203's one-origin rule; it's the inverse direction that seems like too much to me.                                                                                                                                                                                

`dp` was left flat while `o` and `ax` got wrapped, which doesn't match #1627. I used the same flat shape in #1561, so e might need to change more broadly.                                                                                                                                                                                                              

`except Exception: return False` throws away #1631's refusal taxonomy, which just got merged. `vet`'s docstring separates 8 permanent named refusals from `MissingAnchorError`, which "establishes nothing, retryably" (`regeventing.py:374-379`), and `ipexing.py:659-666` collapses all of it into a silent drop. `if rip is None or head is None: return False` at `:649` does the same to a TEL that merely hasn't replicated. keripy's standing treatment of missing evidence is the opposite: `verifying.py:188-191` escrows, cues `dict(kin="proof", said=nodeSaid)`, and raises `MissingChainError`.                                                                                                                                    

The tests were adjusted around the new verification. All 6 `Blinder.unblind` round-trips in `test_ipexing.py` are gone, each replaced by `assert len(nests[0].bsqs) == 1`, which any four primitives satisfy; they came in with #1612 to prove a blind survives IPEX plus KRAM plus storage with the issuer's `crew` intact. 
 
`test_blindable_registry_correlation_minimizing_IPEX_JSON` lost `laterPeek` and `earlierPeek`, the assertions that gave it its name, added in #1578. And `regid=` came out of about 15 test ACDCs, including two worked examples where the comment says so outright: "intentionally omit rd and avoid the step-4 registry-proof path." That drops exactly the case the new verification exists for, and those examples are the ones we sync into the ACDC spec. Negative coverage for `_verifyIssuerAuthNode` is one test, for a missing proof. It seems like maybe some of this simplification is a loss that we should walk back?

The `dp` prefix in the PR's two-node example seems malformed to me. `test_ipexing.py:2804-2807` has `[schema, "/e/holder", []]`, where #203 requires a prefix to begin and end with `/` and to end in the `_` hop, I think. Concatenation gives `/e/holdera/role`. Relatedly, `_validDisclosurePath` checks types only, so `[["E...", "/", [1, {}]]]` validates and a signify-ts author reading "DAG path" as an RFC-6901 pointer round-trips unchanged.                                                                                                                                                                

An unsolicited `offer` defaults `dp` to `[]`, which ACDC spec PR #203 forbids. `mods.setdefault("dp", [])` at `ipexing.py:820` runs ahead of the apply-versus-opener branch. (The solicited case is correct, I think.)                                                                                                                                                  

The walk evaluates no edge operator and no schema pinning, where `verifying.py:171-215` reads `node['o']` and `node['s']`. So a grant whose every I2I edge is violated is accepted, stored, and replayed.                                                                                                                                                               

Nothing confirms the registry named by `rd` was incepted by the ACDC's own Issuer; `vetBindings` (`regeventing.py:304-338`) checks the two SAID equalities and never compares the ACDC's `i`. ACDC spec PR #204 makes that a MUST. This might deserve its own issue.

Some other nits that my codereview found -- not big, but I'm adding them FWIW: `serializeParsedSubstream` is patched one group at a time (`exchanging.py:792-801`), leaving `sscs`, `ssts`, `tdcs`, and `tmqs`; `BlindState`'s `d` uses `.qb64` where the cast is `Castage(Noncer,'nonce')` and `BoundState`'s `bn` uses `.sn` where the cast is `numh`, with no test on the `bsss` path; a `Regery` is built inside `verify()` and never closed (`ipexing.py:637-645`); `int(head.sad["n"], 16)` sidesteps the radix `Number(numh=...)` enforces; `_proofed`, `_anchor`, and `_serder` are duplicated across both test modules after ryan-hansen asked for the opposite on #1578; and the docstring around `:620` says `vetBlind()` where the code calls `vet()`.                                                                                                                                                                                                                              

## Tangle with #1644?                                                                                                                                                                                                                    

`_normalizeNodeStream` raises on any nest that isn't an ACDC node, and #1644's tests pass TEL and KEL artifacts to the builders in lots of places, so whichever merges second rewrites the other's fixtures. Also, the two encode opposite answers to where a verifier's registry evidence comes from: #1644 persists on-the-wire evidence receiver-side because "after KRAM deletes its temporary records, the receiver cannot find all the evidence it accepted with a grant," while #1643 pulls the TEL from the local store. And #1643's `serializeParsedSubstream` fix is a prerequisite for #1644 rather than an independent change: the parser emits plain tuples, `messagize` does `Structor(data=bond)`, and without the fix a #1643-shaped grant raises inside #1644's `except Exception` in `processEscrowPartialSigned`, which removes the escrow and loses the grant. I think I'd merge #1643 first.

KeaxD:
About your comment for the registry verification. I do agree that having the issuer and verifier in the same Regery was not realistic. However, from what I understand, we assumed that the TEL evidence required for regeventing.vet were provided by Observers, so the recipient would get the TEL evidence from there and run the verifications. I'll explicitly mention that in the tests. 
If there is something I am missing and TEL evidence should be transported through IPEX grant inside the node, then please let me know. But it would make sense to me that registry-based evidence would be provided through observers. 

Also currently this PR focuses on rd verification for issuerAuth, which is the 1st part of a 3 part Task. So the issuerAuth function will eventually check all 3 authentication factors: Registry, KEL anchors, Signatures.

dhh1128:
>from what I understand, we assumed that the TEL evidence required for regeventing.vet were provided by Observers, so the recipient would get the TEL evidence from there and run the verifications. I'll explicitly mention that in the tests.
If there is something I am missing and TEL evidence should be transported through IPEX grant inside the node, then please let me know. But it would make sense to me that registry-based evidence would be provided through observers.

Observers are optional. Watchers are necessary to catch duplicity, but observers are a tradeoff that an issuer makes dependeing on their circumstances. I think we should prove that the vetting works whether Observers are active or not. But I haven't thought this through deeply; maybe the tests already prove this just by being agnostic about where the data they got is coming from.

evanja57:
@dhh1128 on your noted tangle with #1644, I agree that this PR should land first, and once it does I can fix up the conflicts during a rebase onto main. I think keeping this Pr's acdc only node shape & serializer fix is correct

dhh1128:
I re-reviewed. This looks very good now, IMO. I'd merge it.

dhh1128:
A follow-up comment that doesn't change my vote about merging:

`_evaluateLeafEdge` returns None for a DI2I edge (`ipexing.py:684`), and since its docstring says None means the leaf shape is malformed and verification fails closed, a grant carrying any DI2I edge gets refused outright; `test_ipexing.py:992` pins that. DI2I is normative in the ACDC spec's unary Operator table on both main and v1.1 (spec-body.md:1194), so this treats a well-formed edge as malformed input. I'm wondering whether the rejection is deliberate policy or a placeholder until the delegation check exists. If it's the latter, False might say it more accurately than None, since False is "the relation doesn't hold" rather than "this message is broken". Or it might belong in the MissingChainError path this PR just added, because whether an Issuer is a delegated AID of the Issuee is another question the verifier can't answer without evidence it may not have yet.

On the registry side, the separate `issuerRgy`/`recipientRgy` split is what I was hoping for, and the escrow-and-cue path looks right to me. My question is about how the recipient's store gets filled: the test calls `recipientRgy.store.accept(...)` directly, under the comment "Simulate observer retrieval". On v1 that simulation has a real counterpart, since the `/query?typ=tel` endpoint from #887 serves TEL events out of `Reger` and `Tevery` ingests them client-side. The v2 blindable registry doesn't have an equivalent yet: `RegistryStore` is referenced only inside `src/keri/acdc/`, so nothing in `app/` or `vdr/` can put a foreign issuer's `rip`/`bup` chain into the store `_verifyIssuerAuthNode` reads. That seems fine as scope for part one of three, and #887 looks like a good model for the v2 side, but I'd like it stated in the PR body or a follow-up issue so that whoever wires up the first real disclosee finds it written down. Since observers are optional, that ingest path is going to be needed either way.

KeaxD:
Thank you for your follow up comments @dhh1128. You mention that "observers are optional", I might be missing something about v2, I know that v1 had TEL queries for data ingestion but I thought that for v2, observers were the means for data ingestion not an option. What are the other ways for data ingestion in v2?

src/keri/acdc/ipexing.py
```diff
@@ -465,15 +1145,51 @@ def offer(hab, message, origin, artifacts=None, apply=None, recp=None, dt=None,
...
             raise TypeError("artifacts must be a list when provided")
-        for artifact in artifacts:
-            nests.append(_normalizeNestedStream(artifact))
-    mods = dict(modifiers) if modifiers else {}
-    mods.setdefault("dp", [])
+        if origin is None:
+            if artifacts:
+                raise ValueError("offer artifacts require an origin")
+        else:
+            nests = [_normalizeNodeStream(origin)]
```
ryan-hansen, line 1190: The origin is only nested if artifacts is not None. Omitting artifacts sets a.o but sends no body.  Is that intentional?

tests/acdc/test_regeventing.py
```diff
@@ -689,6 +721,7 @@ def test_V23_no_salt_parameter_anywhere():
     test_V13_seal_reference_without_seal_retryable()
     test_V14_substitution_sibling_registry_refused()
     test_V15_acdc_rd_mismatch_refused()
+    test_V15b_acdc_issuer_mismatch_refused()
```
ryan-hansen, line 724: Is this the right test name?

src/keri/acdc/ipexing.py
KeaxD: Yes, when artifacts = None then a.o = originSaid, so we can reference the origin ACDC without disclosing the ACDC body in nest.

when artifacts = [ ] then a.o = acdc.said and nest = acdc with no additional artifacts

when artifacts = child then a.o = acdc.said and nest = acdc + child 

But then you could also have a disclosed metadata acdc in nest where
a.o = metadata.said
nest = metadataAcdc

Does that make sense? It made sense to me for offer since it is used for negotiation/metadata, so we could have a way to reference/bind the origin ACDC if we wanted while keeping the ACDC undisclosed

ryan-hansen: Works for me.

dhh1128:
@KeaxD — picking up a thread from this PR. You mentioned that the registry check was the first of three issuerAuth factors, with KEL anchors and signatures to follow. Is there an issue tracking the other two? I couldn't find one, and I'd rather reference it from some v2 work than duplicate it.

KeaxD:
> @KeaxD — picking up a thread from this PR. You mentioned that the registry check was the first of three issuerAuth factors, with KEL anchors and signatures to follow. Is there an issue tracking the other two? I couldn't find one, and I'd rather reference it from some v2 work than duplicate it.

@dhh1128 It is a task made in the Keri Foundation Team in the Keri Foundation repo, I think it might be private so I don't think you would be able to reference it, I think it might be fine for you to create an issue for it.

PR #1646, opened 2026-09-01
Pin gleif_hio 0.6.20rc3

### LINKED ISSUE #1645: Pin gleif_hio 0.6.20rc3
## Scope

Update the KERIpy v1.2.14 dependency pin from `gleif_hio` 0.6.20rc2 to the published 0.6.20rc3 release candidate.

Accommodate only the two production-relevant regressions exposed by the existing Directant module under rc3:

- `Reactant` must contain the new `TransmitClosedError` instead of letting it escape the shared scheduler and log the rejected response byte count locally because rc3 no longer leaves those bytes in `txbs`.
- The `Directant` timer assertion must observe rc3 restart-from-current-time behavior rather than rc2 timer extension.

## Acceptance

- `setup.py` resolves `gleif_hio` 0.6.20rc3.
- The complete `tests/app/test_directing.py` module passes under rc3.
- No connection-level pending state, demo-only `Director` or `Reactor`, currently unowned `Indirector`, regular Messenger, retry, outcome, package-version, or unrelated lifecycle changes are included.
- The known Bob/Eve demo `Reactor` failure under rc3 remains outside this issue.

kentbull:
## Summary

- pin the published `gleif_hio` 0.6.20rc3 release candidate
- contain the `TransmitClosedError` exposed by the production witness `Reactant` cutoff path and log the rejected response byte count locally after rc3 leaves `txbs` empty
- update the existing `Directant` timer assertion for restart from current time

## Boundaries

This includes only the dependency bump and compatibility changes required by the two production-relevant `Directant` regressions. It does not add connection-level pending state; change the demo-only `Director` or `Reactor`, the currently unowned `Indirector`, or regular Messenger outcomes; add retries; bump the KERIpy package version; or change unrelated lifecycle behavior.

The existing `tests/demo/test_demo.py::test_run_bob_eve_demo` remains red under rc3 because its demo-only `Reactor` sends after the client transmit side has closed; that behavior is deliberately outside this PR.

Fixes #1645

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,7 +83,7 @@ setup(
                         'cbor2>=5.6.2',
                         'multidict>=6.0.5',
                         'ordered-set>=4.1.0',
-                        'gleif_hio==0.6.20rc2',
+                        'gleif_hio==0.6.20rc3',
                         'multicommand==1.0.0',
                         'jsonschema>=4.21.1',
                         'falcon>=3.1.3',
--- a/src/keri/app/directing.py
+++ b/src/keri/app/directing.py
@@ -6,6 +6,7 @@ keri.app.directing module
 simple direct mode demo support classes
 """
 import itertools
+from hio import hioing
 from hio.base import doing
 
 from .. import help, kering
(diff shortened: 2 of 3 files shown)


kentbull:
Closing until we have upstream hio changes merged.

PR #1650, opened 2026-09-02
fix: read witness AIDs from config mapping

kentbull:
## Summary

- read the complete mapping returned by `Configer.get()` before selecting optional `aids`
- forward configured witness restrictions to `setupWitness()`
- exercise happy-path `runWitness()` with actual KERI stores, witness endpoints, servers, controller, and scheduler, then inspect the real `ReceiptEnd`
- allocate distinct unprivileged HTTP and TCP ports so the real-server test is portable to Linux CI

## Tests

- focused: `1 passed in 1.95s`
- complete `tests/app/cli/test_kli_commands.py`: `11 passed in 10.74s`
- `git diff --check`: passed


Essbante (approved):
Thanks for catching this. The fix looks right. Configer.get() returns the whole mapping, so the original call was wrong.

Issue #1504, opened 2026-07-14
Enhancement: idle backoff for cue/msg polling doers, which currently reschedule at tock=0.0 on every Doist pass


Nicholas-Keystate:
> **Triage note:** proposed as a performance/design *enhancement* (feature template), not a correctness bug (see Actual behavior below).

### Version

`main` @ `39d2df02` (v2.0.0-dev6) (polling shape and doify sites re-verified at this commit; code permalinks below pin `d31be826`, where the code is identical).

### Affected code (verified at commit `d31be826`)

The standard polling shape throughout the app layer is:

```python
while True:
    while not self.cues:
        yield self.tock        # self.tock defaults to 0.0
    ...
```

e.g. `Respondant.cueDo` ([`src/keri/app/storing.py#L247-L249`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/storing.py#L247-L249)), `Receiptor.witDo` ([`src/keri/app/agenting.py#L246-L256`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/agenting.py#L246-L256)), `WitnessPublisher.sendDo` ([`src/keri/app/agenting.py#L645-L675`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/agenting.py#L645-L675)), `WitnessStart.cueDo` ([`src/keri/app/indirecting.py#L266-L275`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/indirecting.py#L266-L275)), and the equivalent loops in most other doers.

[...]

Issue #1501, opened 2026-07-14
Receiptor/WitnessPublisher cue decks are scanned but never drained, growing without bound in long-running agents


Nicholas-Keystate:
### Version

`main` @ `39d2df02` (v2.0.0-dev6) (all sites re-verified at this commit; code permalinks below pin `d31be826`, where the code is identical — only `WitnessPublisher.idle` has drifted by one line, to `agenting.py:694`). Also verified present in release tag `1.3.5` (push sites `agenting.py:350/:789`; scans `credentialing.py:709`, `grouping.py:211`, `delegating.py:170`; idle accounting `agenting.py:810`) and maintenance branch `v1.2.14` (push sites `agenting.py:230/:600`; scans `credentialing.py:707`, `grouping.py:213`, `delegating.py:170`; idle accounting `agenting.py:621`).

### Affected code (verified at commit `d31be826`)

Producers push a cue for every completed receipt operation:

- `Receiptor.witDo` — [`src/keri/app/agenting.py#L254`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/agenting.py#L254): `self.cues.push(msg)`
- `WitnessPublisher.sendDo` — [`src/keri/app/agenting.py#L673`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/agenting.py#L673): `self.cues.push(evt)`

Consumers only *iterate* the deck; nothing ever pulls or removes an element:

- `Registrar.processWitnessEscrow` — [`src/keri/vdr/credentialing.py#L772`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/vdr/credentialing.py#L772)
- `Counselor.processPartialWitnessEscrow` — [`src/keri/app/grouping.py#L244`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/grouping.py#L244)
- `Anchorer.processPartialWitnessEscrow` — [`src/keri/app/delegating.py#L179`](https://github.com/WebOfTrust/keripy/blob/d31be826206a9929acd2f578e3f49b5220dccf22/src/keri/app/delegating.py#L179)
[...]

Issue #814, opened 2024-07-05
Communication with witnesses hangs if one witness is not responding


rodolfomiranda:
### Version

all

### Environment

all

### Expected behavior

When `Receiptor` sends events to the list of witnesses, it should try to communicate with all witnesses even if one of them is unresponsive. 

### Actual behavior

Current logic in `Receiptor` 
https://github.com/WebOfTrust/keripy/blob/6b6970be92d851f94e58b7a9f4225a81582d2235/src/keri/app/agenting.py#L90-L110
keeps in an infinite loop in line 96 if one witness is unresponsive, preventing the loop over witnesses (line 90) to iterate to the rest. 
Ideally the client should timeout, instead it tries forever, at least using demo witnesses.
One option is to add a timeout in line 96.

### Steps to reproduce

Create a single sig AID with two witnesses with one of them disconnected.

rodolfomiranda:
We may encounter the infinite loop also in:
```
Receiptor
WitnessReceiptor
WitnessInquisitor
WitnessPublisher
TCPMessanger
TCPStreamMessanger
HTTPMessager
```
One option may be to pass a timeout value to those clases to kill loops after a while.

kentbull:
Yes, we need to pass a timeout in to these loops from the Doer/DoDoer tasks that are causing the HTTP calls.

PR #1651, opened 2026-09-02
fix: resolve WitnessPublisher lifecycle bugs (#1652)

### LINKED ISSUE #1652: WitnessPublisher retains completed messenger doers and misreports idle
### Version

`v1.2.14` at `f6d04d4a9`.

### Expected behavior

`WitnessPublisher.idle` should reflect queued and active work. Completed messenger doers should be removed, and consuming a completion cue should not change idle state.

### Actual behavior

- `idle` compares the monotonic `posted` count with the current length of consumable `cues`. Consuming a cue makes an inactive publisher appear permanently busy. A request skipped after dequeue can also increment `posted` without producing a cue.
- `sendDo` empties its local `witers` list before calling `remove(witers)`, so removal is a no-op and completed messenger doers remain scheduled.

Long-lived publishers can accumulate completed children, while callers relying on `idle` can stall.

### Steps to reproduce

1. Create a `WitnessPublisher` with one witness and a controllable fake messenger.
2. Queue an event and recur once; the messenger becomes active.
3. Mark the messenger idle and recur again.
4. Observe that the messenger remains in `publisher.doers`.
5. Consume the completion cue and observe that `publisher.idle` becomes false despite no remaining work.

### Resolution

PR #1651 tracks active messengers explicitly, removes each completed child, and defines idle as no queued messages and no active messengers.

Related to #1501, but does not resolve its broader unbounded cue-retention problem.

kentbull:
## Summary

- Track queued work and active witness messenger children directly.
- Remove and close each completed messenger before reporting the publisher idle.
- Cover queued, active, completed, and consumed-cue states with a regression test.

Fixes #1652.

## Rationale

The previous `idle` check compared the monotonic `posted` count with the current length of `cues`. That was not reliable lifecycle state: completion cues are consumable, so removing a cue could make an inactive publisher appear permanently busy; `posted` was incremented immediately after dequeue, before habitat validation, so a skipped request could be counted without ever producing a cue; and the local messenger list was emptied before `remove()` received it, leaving completed child doers scheduled even when the counters reported idle.

This change makes the lifecycle explicit. `msgs` represents work waiting to start, `witers` represents messenger children still owned by the publisher, and `cues` remains only a completion-notification channel. The publisher is idle only when both queued and active work are absent, independent of whether completion cues have been consumed.

This is local lifecycle accounting: completion still means each child messenger reached its own idle boundary, not that a remote witness durably stored or processed the message.

## Related Issues

The following issues may be affected, but this PR does not completely resolve or close them:

- [WebOfTrust/keripy#1501](https://github.com/WebOfTrust/keripy/issues/1501): directly related. This PR decouples `WitnessPublisher.idle` from cue retention and removes completed messenger children, but it does not bound or drain `WitnessPublisher.cues` or `Receiptor.cues`.
- [WebOfTrust/keripy#1504](https://github.com/WebOfTrust/keripy/issues/1504): removing leaked children reduces unnecessary scheduled doers, but this PR does not add idle backoff.
- [WebOfTrust/keripy#814](https://github.com/WebOfTrust/keripy/issues/814): active messenger state is now represented accurately, but this PR adds no witness timeout, retry, or failure semantics.
- [WebOfTrust/keria#450](https://github.com/WebOfTrust/keria/issues/450): cleanup may reduce CPU and memory accumulation during repeated issuance and revocation, but no KERIA long-run test establishes how much of that issue it explains.
- [WebOfTrust/keria#232](https://github.com/WebOfTrust/keria/issues/232): cleanup may reduce idle scheduling caused by retained messenger children, while the documented tock, escrow, and logging contributors remain outside this PR.
- [WebOfTrust/keria#433](https://github.com/WebOfTrust/keria/issues/433): this PR improves in-memory lifecycle ownership but does not add durable queues, restart recovery, or replay.
- [WebOfTrust/keria#200](https://github.com/WebOfTrust/keria/issues/200): this PR changes local publisher idleness only; it does not change long-running operation dependency or witness-receipt completion semantics.

## Testing

- `tests/app/test_agenting.py`: 6 passed

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -653,9 +653,9 @@ class WitnessPublisher(doing.DoDoer):
             cues (Deck): completed request cues.
         """
         self.hby = hby
-        self.posted = 0
         self.msgs = msgs if msgs is not None else decking.Deck()
         self.cues = cues if cues is not None else decking.Deck()
+        self.witers = []
         super(WitnessPublisher, self).__init__(
             doers=[doing.doify(self.sendDo,
                                tock=hby.tocks["witnessPublisher"])], **kwa)
@@ -671,7 +671,6 @@ class WitnessPublisher(doing.DoDoer):
         while True:
             while self.msgs:
                 evt = self.msgs.popleft()
-                self.posted += 1
                 pre = evt["pre"]
                 msg = evt["msg"]
 
@@ -681,21 +680,22 @@ class WitnessPublisher(doing.DoDoer):
                 hab = self.hby.habs[pre]
                 wits = hab.kever.wits
 
-                witers = []
                 for wit in wits:
                     witer = messenger(hab, wit)
-                    witers.append(witer)
+                    self.witers.append(witer)
                     witer.msgs.append(bytearray(msg))  # make a copy so everyone munges their own
                     self.extend([witer])
 
                     _ = (yield tock)
 
-                while witers:
-                    witer = witers.pop()
+                while self.witers:
+                    witer = self.witers[0]
                     while not witer.idle:
                         _ = (yield tock)
 
-                self.remove(witers)
+                    self.remove([witer])
+                    self.witers.remove(witer)
+
                 self.cues.push(evt)
 
                 yield tock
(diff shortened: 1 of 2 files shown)


daidoji:
+1

PR #1653, opened 2026-09-02
Fix #1654: Admit TCP stream payloads

### LINKED ISSUE #1654: `streamMessengerFrom` drops payload for TCP-only endpoints
### Expected behavior

When `streamMessengerFrom` selects a TCP endpoint, it should queue and transmit the supplied `msg`.

### Actual behavior

The TCP branch creates an empty `TCPStreamMessenger` without enqueueing `msg`. Because an empty queue is valid, no error is raised; the send loop waits indefinitely and transmits nothing.

This affects `StreamPoster` delivery to TCP-only controller, agent, mailbox, or witness endpoints. HTTP and HTTPS endpoints mask the bug because they take precedence.

### Resolution

Enqueue a copy of `msg` in the TCP messenger and add TCP-only regression coverage.

Fixed by #1653.

kentbull:
Fixes #1654.

## Summary

- enqueue the supplied payload when `streamMessengerFrom` selects TCP
- add a regression test that forces the TCP-only branch and verifies queue admission

## Rationale

`streamMessengerFrom` accepts the message to transmit, and its HTTP branch passes that message directly to `HTTPStreamMessenger`. The TCP branch instead created a `TCPStreamMessenger` with an empty queue and discarded the supplied message. Because an empty queue is valid constructor state, this raised no error: the send loop simply waited for work and transmitted nothing.

Existing coverage did not expose the defect. HTTP and HTTPS take precedence whenever either endpoint is advertised, KERIA agents advertise HTTP endpoints, and witness configurations commonly advertise HTTP alongside TCP. Other TCP tests exercise `messengerFrom`, whose callers enqueue messages separately, rather than this stream factory.

## Testing

- `tests/app/test_agenting.py::test_stream_messenger_from_admits_tcp_payload`
- `git diff --check`

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -1075,6 +1075,7 @@ def streamMessengerFrom(hab, pre, urls, msg, headers=None):
     elif kering.Schemes.tcp in urls:
         url = urls[kering.Schemes.tcp]
         witer = TCPStreamMessenger(hab=hab, wit=pre, url=url)
+        witer.msgs.append(bytearray(msg))
     else:
         raise kering.ConfigurationError(f"unable to find a valid endpoint for witness {pre}")
 
--- a/tests/app/test_agenting.py
+++ b/tests/app/test_agenting.py
@@ -66,29 +66,17 @@ def test_http_messengers_read_state_after_client_service():
             doist.exit()
 
 
-def test_stream_messenger_from_admits_tcp_payload(monkeypatch):
-    class FakeTCPStreamMessenger:
-        def __init__(self, *, hab, wit, url):
-            self.hab = hab
-            self.wit = wit
-            self.url = url
-            self.msgs = []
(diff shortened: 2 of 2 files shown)


src/keri/app/agenting.py
```diff
@@ -1075,6 +1075,7 @@ def streamMessengerFrom(hab, pre, urls, msg, headers=None):
     elif kering.Schemes.tcp in urls:
         url = urls[kering.Schemes.tcp]
         witer = TCPStreamMessenger(hab=hab, wit=pre, url=url)
+        witer.msgs.append(bytearray(msg))
```
daidoji, line 1078: msg is already bytes, why cast to bytearray?

daidoji (approved):
Seems fine.  Good catch.  Test failure seems unrelated??  Maybe I'm wrong though.

kentbull:
yes, test failure is unrelated.

PR #1657, opened 2026-09-02
Fix messenger idle accounting

### LINKED ISSUE #1656: Decouple messenger idle state from sent queues
TCP messengers currently report idle while a message is merely queued, so an owning Doer can remove the messenger before transmission starts. TCP and HTTP messengers also derive completion from `sent`, even though callers consume that notification queue; popping a cue can therefore make completed work appear active again.

Messenger lifecycle state should instead reflect queued, in-progress, and pending-response work. Consuming `sent` must not change whether delivery work is complete.

Acceptance criteria:
- Queued and actively transmitting TCP messages are non-idle.
- HTTP remains non-idle until every admitted request has a response.
- Consuming TCP completion or HTTP response cues does not change completed messengers back to non-idle.
- Cover the behavior with real TCP/HTTP delivery tests.

This is application-level accounting only; EOF/connection lifecycle, TLS, and timeout changes are out of scope.

kentbull:
Fixes #1656.

## Summary
- Track queued and in-progress TCP work directly instead of comparing a cumulative counter with `sent`.
- Track outstanding HTTP responses with `pending`.
- Verify both TCP variants and HTTP behavior using real ephemeral endpoints, independent habitats, and real KERI messages.

## Rationale
`sent` is a consumable notification queue, not lifecycle state. The old accounting could report queued TCP work as idle, allowing premature messenger removal, and could report completed TCP or HTTP work as active after a caller consumed its cue.

## Validation
- `tests/app/test_agenting.py`: 10 passed
- `tests/app/test_forwarding.py`: 4 passed

## Scope
This is application-level accounting for v1.2.14. EOF/connection lifecycle, TLS, and timeout changes remain out of scope.

### DIFF
--- a/src/keri/app/agenting.py
+++ b/src/keri/app/agenting.py
@@ -721,7 +721,11 @@ class WitnessPublisher(doing.DoDoer):
 
 
 class TCPMessenger(doing.DoDoer):
-    """Send outbound CESR messages to a witness via TCP and parse inbound receipts."""
+    """Send outbound CESR messages to a witness via TCP and parse inbound receipts.
+
+    ``sent`` is a consumable notification queue, so lifecycle accounting is
+    kept separately in ``msgs`` and ``messageInProgress``.
+    """
 
     def __init__(self, hab, wit, url, msgs=None, sent=None, doers=None, **kwa):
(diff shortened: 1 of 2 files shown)


daidoji:
+1

PR #1660, opened 2026-09-02
Correct KLI challenge file and mailbox request paths

kentbull:
## Summary

- read file-backed challenge words from the path supplied by the words argument
- remove a trailing slash from the mailbox requester base path before appending the endpoint

## Boundary

This extracts only the two KLI corrections from upstream PRs #1091 and #1344. It does not include tests or their unrelated import, SSH export, Exchanger cue, witness authentication, or Receiptor changes.

## Validation

- `git diff --check`


daidoji:
+1

PR #1668, opened 2026-09-02
Add disclosure path support to Compactor

evanja57:
This PR started by adding a failing section to `test_compactor_compact_expand()`.

The test requests non-adjacent blocks from a sample attribute section based on one of the credentials in `test_sedi.py`. The current Compactor does not create the requested disclosure shape.

This proved the need for the Compactor upgrade.

## Update

The implementation now lets Compactor create one partial for any requested combination of disclosure paths.

This change:

- Resolves disclosure paths and their required ancestors
- Includes all descendants when a trailing separator requests a complete node
- Computes the canonical compact form, then restores the requested branches from cached leaves into one partial
- Preserves numeric path selection and non-strict labels, and creates no partial when no path applies to the section
- Extends the existing tests for sibling combinations, overlapping paths, canonical SAIDs, and mappings without their own SAIDs

Calling compact() without disclosure paths keeps the existing behavior.

### DIFF
--- a/src/keri/core/mapping.py
+++ b/src/keri/core/mapping.py
@@ -22,7 +22,7 @@ from ..kering import (Colds, Kinds, EmptyMaterialError, InvalidValueError,
 from ..help import isNonStringIterable
 
 from .counting import  Codens, Counter
-from .coring import (MtrDex, Matter, Labeler, LabelDex, DecDex, Decimer,
+from .coring import (MtrDex, Matter, Labeler, LabelDex, DecDex, Decimer, Pather,
                      DigDex, Diger)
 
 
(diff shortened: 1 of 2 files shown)


SmithSamuelM:
@evanja57 

Question,  does the change still support the old functionality especially for the default cases where one just needs to efficiently compute the most compact SAID and is not doing graduated disclosure to specific do paths?

evanja57:
@SmithSamuelM Yes, the default call still computes the most compact SAID. While checking that, I realized we can simplify this by compacting normally first, then restoring the requested branches from the cached leaves. When benchmarking it locally, it ended up showing essentially the same default performance with faster targeted disclosure, so I opted for the simpler & faster implementation.

ryan-hansen (approved):
Looks ok to me.  Leaving for @SmithSamuelM to review based on his questions raised in the KF dev meeting of 2026-09-08.

SmithSamuelM:
@evanja57 yes looks good.  Since the full compaction computes all the saids all the way out, Uncompacting to the full but computed said version and then collapsing the undisclosed branches will work in general. What wasnt registering is that once the SAIDs are computed, the uncompacted version can be more simply recompacted by substituting the SAID for the undisclased block.

PR #1674, opened 2026-09-10
Add direct KEL anchoring for truthy ax exchanges

KeaxD:
This PR implements the direct-KEL anchoring layer for the V2 IPEX `ax` field described in the Authentication Factors discussion.

KRAM continues to provide the default, time-bounded signature authentication for IPEX messages. When `ax=[True]` is negotiated, this PR additionally requires the binding `agree`, `grant`, and `admit` messages to be anchored in their sender’s KEL, making those commitments perpetually verifiable.

This PR does not implement presentation-registry anchoring or observer-based TEL retrieval. It will come in a follow up PR.

The single-DAG `ax` value must be exactly one boolean when present:

- Omitted `ax` and `ax=[False]` remain signature-only.
- `ax=[]`, multiple values, and non-boolean values are rejected.
- `ax` is exposed as a dedicated builder parameter and may not be supplied through `attrs`.
- `apply`, `offer`, and bare `grant` can signal anchoring.
- An `offer` may introduce `ax=[True]`, even when its prior `apply` was falsy or omitted.
- A prior-linked `grant` must echo its prior and cannot introduce a truthy `ax`.
- `agree` and `admit` automatically copy a truthy requirement from their prior.
- `spurn` is unchanged and is not part of the anchoring set.

The supported anchored flows:

```text
apply -> offer -> agree -> grant -> admit
apply -> grant -> admit
offer -> agree -> grant -> admit
grant -> admit
```

`apply` and `offer` negotiate the requirement but are not themselves anchored. When anchoring is required:

- `agree` is anchored by the disclosee.
- `grant` is anchored by the grantor.
- `admit` is anchored by the disclosee.

The IPEX builders now create a KEL event sealing the binding EXN SAID and attach a `SealSource` couple alongside the existing signature:

- Normal KELs use an interaction event.
- Establishment-only KELs use a rotation.
- The EXN is signed after creating the anchor so an establishment-only rotation uses the updated key state.
- Non-transferable identifiers cannot create anchored IPEX messages.

`IpexHandler.verify()` now:

- Resolves and returns the accepted prior EXN for negotiation checks.
- Enforces `ax` echo and anti-invention rules.
- Requires sender source-seal evidence for truthy binding messages.
- Looks up the referenced event in the sender’s KEL.
- Verifies the sequence number and event digest.
- Verifies that the referenced event seals the exact EXN SAID.
- Rejects missing source couples, incorrect digests, unrelated seals, and non-sender evidence.

The Exchanger now carries sender source-seal couples through the complete message lifecycle:

```text
parse -> dispatch -> verify -> escrow -> retry -> store -> serialize
```

A new `esrc` database store persists `(Number, Diger)` source couples. `serializeMessage()` restores them so accepted EXNs retain their anchor evidence when replayed or forwarded.

Source couples are passed only to handlers declaring `acceptsSscs`, preserving existing V1 handler signatures.

### DIFF
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -495,6 +495,19 @@ class IpexHandler:
                 return False
 
             number, diger = sscs[-1]
+            kever = self.hby.db.kevers.get(serder.pre)
+            if kever is None:
+                raise MissingSenderKeyStateError(
+                    f"missing current sender key state for {serder.pre}")
+
+            lastEst = kever.lastEst
+            if number.sn < lastEst.s:   # Reject if reference is older thant the current key state
+                return False
+            
+            # Check matching sn and diger
+            if number.sn == lastEst.s and diger.qb64 != lastEst.d:
+                return False
+
             prefix = serder.pre.encode("utf-8")
             eventSaid = self.hby.db.kels.getLast(keys=prefix, on=number.sn)
             if eventSaid is None:
@@ -507,6 +520,10 @@ class IpexHandler:
             if event is None:
                 raise MissingSenderKeyStateError(
                     f"missing sender KEL event body at sn={number.sn} for {serder.pre}")
+            
+            # If reference has a higher sn, check that it is an interaction event, otherwise reject
+            if number.sn > lastEst.s and event.ilk != Ilks.ixn:
+                return False
             if not any(isinstance(seal, Mapping)
                        and seal.get("d") == serder.said
                        for seal in (event.seals or [])):
--- a/src/keri/acdc/ipexing.py
+++ b/src/keri/acdc/ipexing.py
@@ -15,11 +15,11 @@ from hio.help import ogler
 from .. import Kinds, Protocols
 from ..kering import (Colds, DuplicitousRegistryError, Ilks, MisanchorError,
                       MisbindingError, MissingAnchorError, MissingChainError,
-                      MisdigestError, MisregistryError, MissequenceError,
-                      RootSealError, UnverifiedBlindError, ValidationError,
-                      Vrsn_2_0, sniff)
+                      MissingSenderKeyStateError, MisdigestError,
+                      MisregistryError, MissequenceError, RootSealError,
+                      UnverifiedBlindError, ValidationError, Vrsn_2_0, sniff)
 from ..core import (BlindState, Blinder, BoundState, Counter, Codens, Diger, GenDex, Noncer,
-                    Number, Saider, Schemer, Serdery, Texter, exchange, messagize)
+                    Number, Saider, Schemer, SealSource, Serdery, Texter, exchange, messagize)
 from ..peer import cloneMessage
 
 logger = ogler.getLogger()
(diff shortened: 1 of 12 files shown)


src/keri/acdc/ipexing.py
```diff
@@ -437,10 +465,54 @@ def verify(self, serder, attachments=None, nests=None):
...
+        messageRequiresAnchor = messageAx[0] is True
+        priorRequiresAnchor = False
+        if pserder is not None:
+            priorAx = pserder.ked.get("a", {}).get("ax", [False])
+            priorRequiresAnchor = priorAx[0] is True
+
+            if verb in (Ipex.agree, Ipex.grant, Ipex.admit):
+                # Binding replies must exactly preserve the negotiated state.
+                if messageRequiresAnchor != priorRequiresAnchor:
+                    return False
```
ryan-hansen, line 486: ax is negotiated on apply/offer/grant only (according to Sam in the last dev meeting).  agree and admit inherit anchoring from the prior and should not be required to echo a.ax. This check will reject valid agree/admit messages that don't include the ax field.  agree/admit should not be required to have an ax at all.  Am I reading that right?

src/keri/acdc/ipexing.py
```diff
@@ -437,10 +465,54 @@ def verify(self, serder, attachments=None, nests=None):
...
+                raise MissingSenderKeyStateError(
+                    f"missing sender KEL event at sn={number.sn} for {serder.pre}")
+            if eventSaid != diger.qb64:
+                return False
 
-        # Stage 4: offer may disclose only a reachable metadata subgraph,
+            event = self.hby.db.evts.get(keys=(prefix, diger.qb64b))
+            if event is None:
+                raise MissingSenderKeyStateError(
+                    f"missing sender KEL event body at sn={number.sn} for {serder.pre}")
```
ryan-hansen, line 509: Should missing event/eventSaid be escrowed vs. just raising?  Or rather, should we escrow in these cases rather than just raising?

src/keri/acdc/ipexing.py
KeaxD: From Sam's post #1613 : "when an anchored exchange is required and agreed upon then the associated agree, grant, and admit messages when provided must include a truthy ax field."

Only apply, offer and grant have the ax field in their callers so you wouldn't be able to set the ax field in agree/admit, right now they are only echoing it. I don't remember @SmithSamuelM saying that agree and admit should not carry the ax field but we should check.

It is also used as a link between exchange messages for anchoring requirement check. Agree inherits offer's ax field, then Grant inherits agree's, then Admit inherits Grant's. It makes it easy to just take from the prior message.

src/keri/acdc/ipexing.py
KeaxD: This check is really a double check that would not normally go off. Exchanger already performs this checks in exchanging.py. KRAM also checks the event and would emit a keystate cue if it doesn't find it. So technically this check could be considered repetitive, that's just defensive coding which is why I raise it instead of escrow. 

We could just trust the validated data from Exchanger but I think it doesn't hurt to have a check here

ryan-hansen: 👍

PR #1684, opened 2026-09-15
Preserve KRAM caches during database clean

evanja57:
Baser.clean currently drops KRAM state when rebuilding the database.

In this PR:

- Add kramCTYP, kramMSGC, kramTMSC, and kramXDT so clean preserves KRAM settings and the records that prevent repeat requests
- Leave pending messages, signatures, and attachments out because they can contain incomplete or unverified data. Retrying these requests requires a fresh timestamp and new signatures

This first pass adds the four missing KRAM databases. Clean already reverifies the KEL, but many other databases, including these four, are still copied without checking whether their contents were altered. I can add that verification in this PR or handle it in a separate PR.


tests/db/test_basing.py
```diff
@@ -1873,13 +1874,14 @@ def test_baser_clone_all_pre_iter():
         assert hab2.pre in pres
 
 
-def test_clean_baser():
+def test_clean_baser(tmp_path, fakeHelpingClock):
```
ryan-hansen: Importing nowIso8601 from keri.help captures the function before fakeHelpingClock monkeypatches helping.nowIso8601 (see conftest.py), causing the generated messages and Kramer’s receiver clock to disagree.

src/keri/db/basing.py
```diff
@@ -1622,7 +1626,8 @@ def clean(self, gvrsn=Version, *, version=None):
                 unsecured = ["hbys", "schema", "states", "rpys", "eans", "tops", "cgms", "exns", "erpy",
                              "kdts", "ksns", "knas", "oobis", "roobi", "woobi", "moobi", "mfa", "rmfa",
                              "cfld", "cons", "ccigs", "cdel", "migs",
-                             "ifld", "sids", "icigs"]
+                             "ifld", "sids", "icigs",
+                             "kramCTYP", "kramMSGC", "kramTMSC", "kramXDT"]
```
ryan-hansen: kramCTYP appears derived from the active configuration and should not be copied unless startup reconciliation removes entries no longer present in that configuration. Otherwise a stale route-specific cache type survives clean and can override current policy. We'll need to either rebuild this database from configuration or add exact reconciliation and a removed-key test.

ryan-hansen (changes requested):

ryan-hansen (approved):

PR #1685, opened 2026-09-16
let witnesses serve some declared metadata to aid discovery

dhh1128:
I am implementing a public sandbox where people can do experiments with a witness (among other things). I realized that I want a way to tell people that the sandbox witness shouldn't be used in production and has no real-world stakes and no SLA. And then I realized that this kind of statement would really help discoverability in the whole ecosystem. So this PR contributes two ways for a witness to serve some metadata set by its controller: `/decl/tags` and `/decl/attribs`.

Tags are bare names a consumer might act on, and can be unioned cleanly across a witness set. The specific problem I'm trying to solve can be reduced to a tag, "testnet", and I'm going to set that tag on my Bakobo witness; I recommend the convention for other witnesses.

Attribs are name=value pairs. A useful attrib pair might be geo = ca, to tell someone they're talking to a witness in Canada.

All of this metadata is self-asserted.

Reply messages seem like a good home for this. /loc/scheme lets an endpoint say where it is, self-authorized, ordered under BADA, and replayed through OOBI, and it works because it doesn't need the declarer's KEL to mean anything. This adds two routes in that shape, /decl/tags and /decl/attribs, along with the record and store pair they need, handlers registered the usual way, two Hab helpers, and dissemination through replyEndRole so a declaration rides along with the loc scheme a witness is already sending. Two routes rather than one because BADA orders each route on its own: tags change almost never and attribs change often, and sharing a route would mean re-stamping a tag declaration to fix a typo in a contact address.

The /decl segment of the route is debatable. I couldn't find a written route-naming convention in the repo, and something like /meta, or a home under /end, might suit you better, so say if you'd spell it differently. It costs two named sub-databases, which moves the bound in test_basing from 105 to 107 against a MaxNamedDBs of 128. There are 8 new tests in tests/core/test_decls.py, and tests/core, tests/app, tests/db and tests/end pass at 548 passed, 2 skipped.

I'd also be curious if you think it's worth documenting the endpoints and conventions anywhere.


src/keri/core/eventing.py
```diff
@@ -5234,6 +5236,101 @@ def processReplyLocScheme(self, *, serder, diger, route,
...
+                                  "".format(kind, value, Ilks.rpy, serder.ked))
+        return value
+
+
+    def _acceptDecl(self, *, serder, diger, kind, cigars, tsgs, value):
+        """Apply BADA to a decl reply and store it if it is the latest from its declarer."""
+        route = f"/decl/{kind}"
+        eid = serder.ked["a"]["eid"]
+        aid = eid  # a declaration is about, and authorized by, the same identifier
+        keys = (aid, kind)
```
ryan-hansen, line 5303: Does aid/eid need to be stored as qb64?

src/keri/db/basing.py
```diff
@@ -1080,6 +1080,16 @@ def reopen(self, **kwa):
...
+
+        # auth AuthN by declarer at eid of its own declaration of a given kind
+        # maps key=eid.kind to val=diger of decl reply
+        self.dans = subing.CesrSuber(db=self, subkey='dans.', klas=coring.Diger)
+
+        # self-declarations keyed by eid.kind where kind is one of kering.Decls
+        # data extracted from reply /decl/tags or /decl/attribs
+        self.decls = koming.Komer(db=self,
+                                  subkey='decls.',
+                                  klas=DeclRecord, )
```
ryan-hansen, line 1092: We should add these to the docstring

ryan-hansen:
How do decls get populated on a live witness?  Is the idea to add some kind of config/cli thing?

ryan-hansen (changes requested):

dhh1128:
I'm toggling this PR to draft while we take a week to study KAPI.

PR #1688, opened 2026-09-18
improve input validation and harden some code paths affected by input

dhh1128:
These fixes came from a careful code analysis I ran a few days ago, which created a census of input sites and coe paths affected by them. I put each fix in its own commit, so we could kill one without others, if we want.

keripy defines a deliberate exception hierarchy rooted at `keri.kering.KeriError`, and callers across the codebase catch that hierarchy on purpose. In several input-parsing paths, hostile or malformed input instead escapes as a raw Python exception (`KeyError`, `UnicodeDecodeError`, `binascii.Error`, `ZeroDivisionError`, `OverflowError`, `TypeError`, `ValueError`), slipping past every `except KeriError` guard. A couple of paths also did unbounded work on hostile input.

This PR is routine input-validation hardening: each fix is a **pure narrowing** that converts a raw exception (or an unbounded loop) into the appropriate `keri.kering` error, or adds a sane bound. No valid-input behavior changes, no existing `keri` error is weakened, and the CESR wire format is untouched. Inputs accepted before are still accepted with identical output; inputs that already raised a `keri` error still do.

Every fix was developed red/green: a regression test that feeds the malformed input and asserts the desired `keri` error was written first and shown to fail with a raw exception on the base commit, then made to pass by the fix, with a companion valid-input test proving the output is unchanged.

Here are some specific fixes:

**1. `helping.b64ToInt` choke point.** This helper is the funnel for the soft/count/index numeric fields of `Matter`/`Counter`/`Indexer`. On a character outside the Base64 alphabet it raised a raw `KeyError`, and on non-UTF-8 bytes a raw `UnicodeDecodeError`. Both failure modes are now narrowed to `kering.ConversionError` at the choke point (`help/helping.py`). Valid Base64 conversion is unchanged (round-trip proven). Census findings A-53, A-06, A-11, A-32.

**2. Hard-coded `.decode()` on attacker bytes, and the error-handler double-decode.** Two decode families escaped the hierarchy:
- Raw `UnicodeDecodeError` from `.decode("utf-8")` on the non-UTF-8 code selector / soft / count / index fields and from `Texter.text`. A new `helping.decodeUtf8` — a drop-in for the existing `if hasattr(x, "decode"): x = x.decode("utf-8")` idiom that narrows the failure to `ConversionError` — is applied at each site: `coring.py` Matter `_exfil` (selector/hard/soft) and `Texter.text`; `counting.py` Counter `_exfil`; `indexing.py` Indexer `_exfil`.
- The `DeserializeError` handlers re-decoded the binary body they were reporting on, so a malformed CBOR/MsgPack/non-UTF-8-JSON message made the handler itself raise `UnicodeDecodeError`, destroying the intended error. Those messages are now built with `errors="replace"`: `coring.loads` (JSON) and `Serder.loads` (CESR/JSON/MGPK/CBOR branches in `serdering.py`).
- Raw `binascii.Error` from `urlsafe_b64decode` on a value region whose surviving Base64 length is not decodable is narrowed to `ConversionError` in `Matter._exfil` and `Indexer._exfil`.

**3. `Tholder` threshold parsing.** A hostile `sith` produced raw `ZeroDivisionError` (`"1/0"`), `OverflowError` (`"inf"`), `TypeError` (an integer given as a float), and `ValueError` (negative, out-of-range, non-numeric, empty, sum-below-1, or huge-integer fraction) — all escaping the hierarchy — and several pathological thresholds (thousands of clauses/weights) were accepted silently.

`Tholder.weight`, `_processSith`, `_processUnweighted`, and `_processWeighted` now raise the existing `kering.ValidationError` (a `KeriError`) for the failures above, so hostile input no longer escapes `except KeriError`. The ~30 `test_tholder` assertions that caught the raw `ValueError` are updated to `ValidationError`. This keeps the exception hierarchy's public shape — no new class inheriting from a builtin.

`_processSith` also bounds three dimensions before converting any fractions: the clause count and per-clause weight count by `Tholder.Limit` (1000, generous for real multisig), the **total** weight/leaf count across all clauses by the same limit (so `clauses × weights`, e.g. `[["1/1000"]*1000]*1000` at ~10 MB / ~1e6 fractions, cannot blow past the per-dimension caps), and the length of a JSON-string `sith` by `Tholder.MaxSith` before `json.loads` (a deeply-nested string such as `"[" * 100000` otherwise overflows the C stack inside the decoder as a raw `RecursionError`).

**4. `Pather._resolve` SAD-path traversal.** Walking a SAD along a caller-supplied path raised raw `KeyError` (missing key, or traversal into a scalar) and raw `ValueError` (non-digit list index), and recursed with no depth cap. Each failure is narrowed to `kering.InvalidValueError` (the error `Pather` construction already uses), and traversal depth is capped at `Pather.MaxDepth` (256) so a pathological path raises a `keri` error instead of running to a `RecursionError`. Two existing `test_pather` assertions that expected the raw `KeyError` are aligned to `InvalidValueError`, matching Pather's own convention; the sole in-tree caller (`app/forwarding.py`) does not catch `KeyError`, so this is a pure narrowing. Census finding A-17.

The `Pather.path`/`.parts`/`.rparts` accessors reached the raw bytes through a bare `self.raw.decode()` for any code not in `BexDex` (e.g. `Salt_128` `0A`), so `resolve()` raised a raw `UnicodeDecodeError` before `_resolve` ran on a `Pather` built straight from the wire. Those accessors now decode through `helping.decodeUtf8` and narrow to the same `kering.InvalidValueError`.

**5. Truncated / mislabelled primitives and remaining decode neighbours.**
- `Matter._exfil` checked `len(qb64b) < hs` but then sliced the soft size field without checking the buffer covers `cs = hs+ss`, so a truncated variable-size primitive (e.g. `Matter(qb64b=b"4A")`) fed `b64ToInt("")` and raised a raw `ValueError`. It now raises `kering.ShortageError` — matching `Indexer._exfil`, and load-bearing so a stream parser waits for more bytes rather than treating the tail as corrupt. `b64ToInt("")`'s empty-string case is also narrowed to `ConversionError` as defence in depth.
- `Labeler.label` and `Labeler.text` — the neighbours of the already-fixed `Texter.text` and the carriers of CESR-native v2 field labels — did a bare `self.raw.decode()` in their `else` branch and raised raw `UnicodeDecodeError` on non-UTF-8 raw. Both now decode through `helping.decodeUtf8` (narrowing to `ConversionError`), identical to the `Texter` fix.

I wrote new regression suites, all red on the base commit and green after the fix, each with a valid-input companion proving identical output: `tests/core/test_harden_b64toint.py`, `tests/core/test_harden_decode.py`, `tests/core/test_harden_tholder.py`, `tests/core/test_harden_pather.py`.
- The existing `test_coring` / `test_counting` / `test_indexing` / `test_serdering` / `test_helping` / `test_parsing` / `test_annotating` / `test_eventing` / `test_habbing` suites remain green. `test_tholder`'s ~30 threshold-rejection assertions and the two `test_b64_conversions` empty-string assertions are updated from the raw `ValueError`/`ValueError` they caught to the narrowed `kering.ValidationError`/`ConversionError`, and `test_pather` remains green.


src/keri/core/coring.py
```diff
@@ -4494,26 +4554,78 @@ def _processSith(self, sith: int | str | Sequence):
...
+                    # it cannot escape `except KeriError`.
+                    raise ValidationError(f"Threshold JSON nesting too deep.") from ex
+
+            if not isNonStringSequence(sith):  # e.g. float, mapping, or other scalar
+                raise ValidationError(f"Invalid sith = {sith}, expected a sequence "
+                                     f"of weights.")
 
             if not sith:  # empty or None
-                raise ValueError(f"Empty weight list = {sith}.")
+                raise ValidationError(f"Empty weight list = {sith}.")
```
ryan-hansen: Do we need a release note that invalid sith / SAD paths no longer raise ValueError / KeyError. Downstream that uses ValueError/KeyError around Tholder / Pather.resolve will have to switch to kering.ValidationError / kering.InvalidValueError (or KeriError).

ryan-hansen (changes requested):

